PR #116, opened 2024-07-31
Infra/csp headers

### LINKED ISSUE #64: CSP Headers
## Full User Story
The frontend domain(s) should respond with an appropriate set of Content Security Policy headers. These headers instruct the user's browser to not allow unexpected content to load.

### Dependencies
- The domains for each environment need to be known OR we need to come back to this and update when known

### Acceptance Criteria
- A passing score on `https://securityheaders.com/
- IA completes code review
- QA: regression test

jsandoval81:
### Requirements List
- **New CSP Lambda**
    - `/backend/compact-connect/lambdas/cloudfront-csp`
    - Follow README setup
- **Browser frontend**
    - `/webroot`
    - `yarn install --ignore-engines`

### Description List
- Filled out the `cloudfront-csp` lambda function to respond with security headers, including CSP
- Updated the browser frontend app to address any CSP items
    - Updated some SVGs to remove inline styles
    - Updated the way translations are handled & built with `vue-i18n`
- Updated the deployment pipeline with updated domains

### Draft Todo
- [x] Confirm updated IA deployment pipelines (if possible) _**(IA)**_
- [x] Remove any feature branches from pipeline files (if possible) _**(IA)**_

_Note: This is one of the last (or the last?) features cut from the "old" way of managing trunks - so it has a messy commit history. Moving forward, we're using a cleaner process for managing trunks across repos._

Closes #64


jsandoval81:
@jlkravitz This is ready for review.

jlkravitz:
- [ ] ~stand up the site locally~
- [ ] review static code analysis results
- [ ] run a security audit of dependencies (e.g. `npm audit`) to ensure that there are no vulnerabilities that will be deployed to production (as opposed to vulnerabilities that only have an impact on the development environment)
- [ ] review all tests

backend/compact-connect/tests/unit/test_app.py
```diff
@@ -95,6 +95,7 @@ def test_synth_sandbox_no_domain(self):
         self._check_no_annotations(app.sandbox_stage.persistent_stack)
         self._check_no_annotations(app.sandbox_stage.ui_stack)
         self._check_no_annotations(app.sandbox_stage.api_stack)
+        self._check_no_annotations(app.sandbox_stage.ingest_stack)
```
jlkravitz, line 98: This related to this PR or just caught something from a past sprint?

webroot/vue.config.js
```diff
@@ -296,6 +306,9 @@ module.exports = {
         plugins: [
             faviconsPlugin,
             stylelintPlugin,
+            // The runtime-only version of vue-i18n conflicts with test-runner, but is required for built server versions due to CSP.
```
jlkravitz, line 309: What is test-runner?

webroot/src/models/State/State.model.ts
```diff
@@ -38,7 +38,17 @@ export class State implements InterfaceStateCreate {
...
+        let states = this.$tm('common.states') || [];
+
+        if (typeof states[0]?.abbrev === 'function') {
+            const normalize = ([value]) => value;
+
+            states = states.map((st) => ({
+                abbrev: st.abbrev({ normalize }),
+                full: st.full({ normalize }),
+            }));
+        }
```
jlkravitz, line 50: What's going on here? (I understand step-by-step what's happening, but wondering why it needs to be done, and why `abbrev` might be something other than a string.)

Same question on `StateUpload.ts` but I assume the answer is the same/similar.

backend/compact-connect/lambdas/cloudfront-csp/test/config/index.js
```diff
@@ -0,0 +1,59 @@
...
+//  CompactConnect
+//
+//  Created by InspiringApps on 7/22/2024.
+//
+
+const chaiMatchPattern = require('chai-match-pattern');
+const chai = require('chai').use(chaiMatchPattern);
+const chalk = require('chalk');
+const lambdaLocal = require('lambda-local');
+const path = require('path');
```
jlkravitz, line 12: What's with the CommonJS `require` imports? Something to do with lambda requirements or environment?

backend/compact-connect/lambdas/cloudfront-csp/test/index.test.js
```diff
@@ -0,0 +1,388 @@
...
+    ].join(' ');
+    const cspScriptSrcElem = [
+        '\'self\'',
+    ].join(' ');
+    const cspScriptSrcAttr = [
+        '\'self\'',
+    ].join(' ');
+    const cspWorkerSrc = [
+        '\'self\'',
+    ].join(' ');
```
jlkravitz, line 69: I assume your goal here is to let this be extendable going forward? It's definitely less readable to me than just e.g., `cspDefaultSrc = "'none'";` but perhaps worth it. Mind adding a comment at the top just indicating why it's written like this?

backend/compact-connect/lambdas/cloudfront-csp/test/index.test.js
```diff
@@ -0,0 +1,388 @@
...
+    const cspFontSrc = [
+        '\'self\'',
+        'https://fonts.gstatic.com',
+    ].join(' ');
+    const cspImgSrc = [
+        '\'self\'',
+        'data:',
+        dataApiUrl,
```
jlkravitz, line 88: What's this doing?

backend/compact-connect/lambdas/cloudfront-csp/test/index.test.js
```diff
@@ -0,0 +1,388 @@
...
+        '\'none\'',
+    ].join(' ');
+    const cspFormActionSrc = [
+        '\'none\'',
+    ].join(' ');
+    const cspConnectSrc = [
+        '\'self\'',
+        dataApiUrl,
+        s3Url,
+        'https://cognito-idp.us-east-1.amazonaws.com'
```
jlkravitz, line 110: What's this URL? Perhaps hoist up to top of function and label?

backend/compact-connect/lambdas/cloudfront-csp/test/index.test.js
```diff
@@ -0,0 +1,388 @@
...
+        `style-src-attr ${cspStyleSrcAttr};`,
+        `font-src ${cspFontSrc};`,
+        `img-src ${cspImgSrc};`,
+        `media-src ${cspMediaSrc};`,
+        `frame-src ${cspFrameSrc};`,
+        `frame-ancestors ${cspFrameAncestorsSrc};`,
+        `object-src ${cspObjectSrc};`,
+        `form-action ${cspFormActionSrc};`,
+        `connect-src ${cspConnectSrc};`,
+    ].join(' ')}`;
```
jlkravitz, line 131: 🎊

backend/compact-connect/lambdas/cloudfront-csp/test/index.test.js
```diff
@@ -0,0 +1,388 @@
...
+                headers: {},
+            };
+            const config = lambdaConfig({
+                lambdaPath: `index.js`,
+                request,
+                response,
+            });
+            const result = await runLambda(config);
+
+            // console.log(JSON.stringify(result, null, 2)); // @DEBUG
```
jlkravitz, line 158: remove debug code

backend/compact-connect/lambdas/cloudfront-csp/test/index.test.js
```diff
@@ -0,0 +1,388 @@
...
+
+            expect(result.headers['strict-transport-security'][0].key).to.equal('Strict-Transport-Security');
+            expect(result.headers['strict-transport-security'][0].value).to.equal('max-age=31536000; includeSubdomains; preload');
+            expect(result.headers['x-content-type-options'][0].key).to.equal('X-Content-Type-Options');
+            expect(result.headers['x-content-type-options'][0].value).to.equal('nosniff');
+            expect(result.headers['x-frame-options'][0].key).to.equal('X-Frame-Options');
+            expect(result.headers['x-frame-options'][0].value).to.equal('DENY');
+            expect(result.headers['x-xss-protection'][0].key).to.equal('X-Xss-Protection');
+            expect(result.headers['x-xss-protection'][0].value).to.equal('1; mode=block');
```
jlkravitz, line 167: This looks to be nonstandard. If you think we need it, mind adding a comment as to why, and perhaps point out why (or how) the vulnerability it describes is not an issue for us?

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection

backend/compact-connect/lambdas/cloudfront-csp/test/index.test.js
```diff
@@ -0,0 +1,388 @@
...
+            expect(result.headers['referrer-policy'][0].key).to.equal('Referrer-Policy');
+            expect(result.headers['referrer-policy'][0].value).to.equal('strict-origin-when-cross-origin');
+            expect(result.headers['content-security-policy'][0].key).to.equal('Content-Security-Policy');
+            expect(result.headers['content-security-policy'][0].value).to.equal(buildCspHeaders(environment));
+            expect(result.headers.server[0].key).to.equal('Server');
+            expect(result.headers.server[0].value).to.equal('CompactConnect');
+        });
```
jlkravitz, line 318: I know this was a last discussion on last PR... but I think this might cross the line of being explicit on tests. I _expect_ that these tests are all identical, but I really can't tell easily. Could we refactor? It seems the only thing changing is the environment variable.

backend/compact-connect/lambdas/cloudfront-csp/test/index.test.js
```diff
@@ -0,0 +1,388 @@
...
+        it('should successfully return the security headers for csg prod when lambda event is missing the request domain', async () => {
```
jlkravitz, line 355: Thanks for testing these cases!

backend/compact-connect/lambdas/cloudfront-csp/index.js
```diff
@@ -1,5 +1,337 @@
...
+/**
+ * Configuration of supported domains per environment.
+ * @type {object}
+ */
+const environments = {
+    csg: {
+        prod: {
+            webFrontend: `app.compactconnect.org`,
+            dataApi: `api.compactconnect.org`,
+            s3Upload: ``, // @TODO: Waiting for environment configuration.
```
jlkravitz, line 20: WIll this be addressed for this PR, or is it a backlog item? If the latter, can we make sure it's being tracked elsewhere besides the code?

backend/compact-connect/lambdas/cloudfront-csp/index.js
```diff
@@ -1,5 +1,337 @@
...
+            webFrontend: `app.test.compactconnect.org`,
+            dataApi: `api.test.compactconnect.org`,
+            s3Upload: `test-persistentstack-mockbulkuploadsbucket0e8f27eb-ivslg22f2yfz.s3.amazonaws.com`,
+        },
+    },
+    ia: {
+        prod: {
+            webFrontend: `app.jcc.iaapi.io`,
+            dataApi: `api.jcc.iaapi.io`,
+            s3Upload: ``, // @TODO: Waiting for environment configuration.
```
jlkravitz, line 32: Same here.

backend/compact-connect/lambdas/cloudfront-csp/index.js
```diff
@@ -1,5 +1,337 @@
...
+// ============================================================================
+//                                   HELPERS                                  =
+// ============================================================================
+/**
+ * Get the request domain from the lambda event record.
+ * @param  {object} eventRecord The cloudfront record from the lambda event.
+ * @return {string}             The bare domain of the request domain.
+ */
+const getRequestDomain = (eventRecord) => {
+    let requestDomain = environments.csg.prod.webFrontend;
```
jlkravitz, line 56: Why is `prod` the default?

backend/compact-connect/lambdas/cloudfront-csp/index.js
```diff
@@ -1,5 +1,337 @@
...
+
+// ============================================================================
+//                                   HELPERS                                  =
+// ============================================================================
+/**
+ * Get the request domain from the lambda event record.
+ * @param  {object} eventRecord The cloudfront record from the lambda event.
+ * @return {string}             The bare domain of the request domain.
+ */
+const getRequestDomain = (eventRecord) => {
```
jlkravitz, line 55: Is it your style guide's standard to use arrow functions here? (What style standards, if any, are you using for JS?)

backend/compact-connect/lambdas/cloudfront-csp/index.js
```diff
@@ -1,5 +1,337 @@
...
+/**
+ * Helper to escape CSP src keywords.
+ * @param  {string} keyword The standard keyword value.
+ * @return {string}         The escaped keyword value.
+ */
+const srcKeywordEscape = (keyword) => {
+    let escaped = '';
+
+    if (keyword && typeof keyword === 'string') {
+        escaped = `'${keyword}'`.toLowerCase();
```
jlkravitz, line 132: how is lowercasing it "escaping" it?

backend/compact-connect/lambdas/cloudfront-csp/index.js
```diff
@@ -1,5 +1,337 @@
...
+ */
+const srcKeywordsEscape = (srcList, listName = '') => {
+    // https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy#keyword_values
+    const srcKeywordsConfig = [
+        { value: 'self', isAllowed: true },
+        { value: 'none', isAllowed: true },
+        { value: 'strict-dynamic', isAllowed: true },
+        { value: 'report-sample', isAllowed: true },
+        { value: 'inline-speculation-rules', isAllowed: true },
+        { value: 'unsafe-inline', isAllowed: true },
```
jlkravitz, line 151: I see the word `unsafe`, and I get scared 😆 ... do we need to allow this? if yes, add a brief comment above?

backend/compact-connect/lambdas/cloudfront-csp/README.md
```diff
@@ -0,0 +1,46 @@
+# Cloudfront CSP Header Edge Lambda
```
jlkravitz, line 1: I admittedly don't know AWS well enough, but it seems odd to me that setting CSP headers is in its own lambda. Is that just "the way"?

backend/compact-connect/lambdas/cloudfront-csp/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+_[back to top](#cloudfront-csp-header-edge-lambda)_
+
+---
+## Tests
+- `yarn test`
+
+#### Test output verbosity
+- `index.test.js`
+    - Test block `it()`
```
jlkravitz, line 45: What does this mean?

.github/workflows/webroot-deploy-development.yml
```diff
@@ -7,7 +7,7 @@ on:
   # Triggers the workflow on pushes to trunk branches involving changes to web frontend files
```
jlkravitz, line 7: What's the reason for the changes in this file?

.github/workflows/check-webroot.yml
```diff
@@ -9,6 +9,7 @@ on:
     branches:
```
jlkravitz, line 9: What's the reason for the changes in this file?

jlkravitz (commented):
Generally looks good. Several questions to help me better understand!

**Update:** Seems like the `unsafe-eval` is needed for i18n? I get an error on deploy that says:

```text
EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self'".
```

backend/compact-connect/tests/unit/test_app.py
jusdino: This change is not related to this PR, but it is a change that should have already been merged in as part of the domain name PR. We had a bit of a knot tied in our git history as we converted from the per-sprint PRs to per-feature PRs. I'm guessing I somehow backed this change back out of my PR when we were untying that knot. Can we keep this in here so it gets corrected with this merge?

webroot/vue.config.js
jsandoval81: In this case, it's Mocha + Vue Test Utils. Usually this isn't an issue because libraries either don't have a separate runtime-only version, or if they do then it's trivially handled by Webpack. The vue-i18n library is a one-off in that it needs really specific build / plugin handling. They claim to have addressed this in their latest Beta version, but we're not ready to move to that yet.

webroot/src/models/State/State.model.ts
jsandoval81: Good question. It's related to the previous item about the separate runtime-only version of vue-i18n. Their current **non**-runtime-only version has an issue with CSP; and their build bundling is a little chaotic. It was a decision to stick with that library for now, as they claim to have addressed these things in their latest Beta version. Once they've gone stable with that new version, we can evaluate if we want to stick with it, or migrate away.

backend/compact-connect/lambdas/cloudfront-csp/test/config/index.js
jsandoval81: Basically just that CommonJS is still the default in Node; and maybe also that many libraries support CommonJS. Things like ESM tree-shaking aren't ever in play in our backend code. So I guess just simplicity overall.

backend/compact-connect/lambdas/cloudfront-csp/test/index.test.js
jsandoval81: Added a comment in the code.

backend/compact-connect/lambdas/cloudfront-csp/test/index.test.js
jsandoval81: Sometimes SVG images have base64 values / bodies. In those case the `<img src` uses the `data:` scheme. It's oddly common with font & css libraries. Here we're restricting it to image sources.

backend/compact-connect/lambdas/cloudfront-csp/test/index.test.js
jsandoval81: Moved to the top of the function.

backend/compact-connect/lambdas/cloudfront-csp/test/index.test.js
jsandoval81: Removed.

backend/compact-connect/lambdas/cloudfront-csp/test/index.test.js
jsandoval81: Removed.

backend/compact-connect/lambdas/cloudfront-csp/test/index.test.js
jsandoval81: Factored this to a helper function.

backend/compact-connect/lambdas/cloudfront-csp/index.js
jsandoval81: Just so if the lambda event structure changes without us knowing, the Prod CSP headers will be returned; keeping the Prod site functional while we work on an update.

backend/compact-connect/lambdas/cloudfront-csp/index.js
jsandoval81: Yes, we're favoring arrow functions over the `function` keyword (& its associated differences). Our linter styles are using the AirBnB rules as a base, with a small number of modifications.

backend/compact-connect/lambdas/cloudfront-csp/index.js
jsandoval81: The quotes around `${keyword}` are doing the escaping. The lowercasing is just keeping things uniform.

backend/compact-connect/lambdas/cloudfront-csp/index.js
jsandoval81: Good catch. We weren't actually using it, but it should be disallowed nonetheless.

backend/compact-connect/lambdas/cloudfront-csp/README.md
jsandoval81: Good question. This is technically running in the context of a CDN viewer response. I updated this to reflect that.

backend/compact-connect/lambdas/cloudfront-csp/README.md
jsandoval81: Yeah, that was a little light on detail 🙂. I updated it to add detail.

.github/workflows/webroot-deploy-development.yml
jsandoval81: The domains were recently made permanent, allowing us to stably implement CSP. Updating the domains in this deployment pipeline was part of making it all work together.

.github/workflows/check-webroot.yml
jsandoval81: Same as in `webroot-deploy-development.yml` above.

backend/compact-connect/lambdas/cloudfront-csp/index.js
jsandoval81: It'll require the step of setting up the Prod environment. Browsing the board, I didn't see something exactly for that. Do you or @jusdino know if that's part of an existing ticket?

backend/compact-connect/tests/unit/test_app.py
jlkravitz: Got it. Sounds good!

backend/compact-connect/lambdas/cloudfront-csp/test/config/index.js
jlkravitz: Got it, thanks!

backend/compact-connect/lambdas/cloudfront-csp/test/index.test.js
jlkravitz: Thanks!

backend/compact-connect/lambdas/cloudfront-csp/index.js
jlkravitz: makes sense!

backend/compact-connect/lambdas/cloudfront-csp/index.js
jlkravitz: Got it!

backend/compact-connect/lambdas/cloudfront-csp/index.js
```diff
@@ -1,5 +1,337 @@
...
+ */
+const buildSrcString = (name = '', list = []) => {
+    let src = [];
+    let srcString = '';
+
+    if (Array.isArray(list)) {
+        src = src.concat(list);
+    }
+
+    srcKeywordsEscape(src, name);
```
jlkravitz, line 198: Didn't catch this last time, but why not just assert (or similar) that list is an array of strings and remove all the checks in `buildSrcString` and `srcKeywordsEscape`? You may have a good reason but am curious because it seems it would simplify the code significantly by removing many edge cases.

backend/compact-connect/lambdas/cloudfront-csp/index.js
```diff
@@ -1,5 +1,337 @@
...
+                'none',
+            ]),
+            buildSrcString('form-action', [
+                'none',
+            ]),
+            buildSrcString('connect-src', [
+                'self',
+                domains.dataApi,
+                domains.s3Upload,
+                'https://cognito-idp.us-east-1.amazonaws.com',
```
jlkravitz, line 278: can we hoist this URL up like we did in the test suite?

webroot/src/models/State/State.model.ts
jlkravitz: Got it. Can we make a backlog item that references the relevant code, including this code and that in `StateUpload.ts`, to make an update when the library is updated or if we decide to change translation libraries?

jlkravitz (changes requested):
A couple small follow-ups. Close!

webroot/src/models/State/State.model.ts
jsandoval81: Sounds good. I created #118 to track this.

backend/compact-connect/lambdas/cloudfront-csp/index.js
jsandoval81: Created #119 to track.

backend/compact-connect/lambdas/cloudfront-csp/index.js
jsandoval81: This `buildSrcString()` function had some extra cruft that I removed. Basically remnants of when multiple optional lists could be passed in. No longer necessary. The rest I feel kinda keeps each function durable & able to stand on its own.

backend/compact-connect/lambdas/cloudfront-csp/index.js
jsandoval81: Done.

backend/compact-connect/lambdas/cloudfront-csp/index.js
jlkravitz: Makes sense, thanks for the explanation!

backend/compact-connect/lambdas/cloudfront-csp/README.md
```diff
@@ -0,0 +1,54 @@
+# Cloudfront Viewer Response Edge Lambda
```
jlkravitz, line 1: One remaining item- 

Seems like the unsafe-eval is needed for i18n? I get an error on deploy that says:

```EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self'".```

**Review by jlkravitz — CHANGES_REQUESTED** [review]

File: backend/compact-connect/lambdas/cloudfront-csp/README.md (hunk 25 above)
Comment (jsandoval81) [reply backend/compact-connect/lambdas/cloudfront-csp/README.md:1]: Oh, interesting. I'm not seeing that in our development environment. Are you sure you're deploying the frontend with the latest build from `yarn build`?

Comment (jlkravitz) [reply backend/compact-connect/lambdas/cloudfront-csp/README.md:1]: Honestly, that is a fair question. I'll be back...

Comment (jlkravitz) [reply backend/compact-connect/lambdas/cloudfront-csp/README.md:1]: @jsandoval81 Forgot I needed to upload the `dist/` folder to the S3 bucket.

Regardless, the UI loads, but I get the below error:

```
Refused to connect to 'https://7suc1ey74m.execute-api.us-east-1.amazonaws.com/joshua/mock/providers/query' because it violates the following Content Security Policy directive: "connect-src 'self' https://api.compactconnect.org  https://cognito-idp.us-east-1.amazonaws.com".
```

Is this a configuration issue on my end, or is this a case we didn't handle?

jsandoval81: Technically, a config issue in your stack. The CSP is working as expected in this case by refusing to let the frontend connect to your personal server-side (`https://7suc1ey74m.execute-api.us-east-1.amazonaws.com`).

If that's ok for now then it looks like we're good. Otherwise, we'll need to figure out the best way to add your personal environment to the CSP. That would be most straightforward if you had a long-lived domain for the both your frontend & backend.

jusdino: I don't think we decided on a hard line or anything, but I believe we decided that we would handle two cases for UI:
1) Create a domain name and we support the hosted UI infrastructure (and the CSP works)
2) Run the UI from localhost, point to the API in the cloud (CSP is irrelevant)

jlkravitz: Hmm OK. This is important though not an immediate priority but, eventually, it should be possible to stand up a test site in a few minutes with at most a few commands and basic technical knowledge. I'm concerned that this is another barrier to entry for doing so.

I plan on sharing the steps I had to take to stand up the site, which I may convert into a PR on the repo's README for your review. Let's continue this discussion at that time and, for now, operate under the status quo described by @jusdino (create a domain name, or run from localhost).

jlkravitz (approved):
Approving this, but with a note to circle back on https://github.com/csg-org/CompactConnect/pull/116#discussion_r1704686142 in the medium-term.

Issue #105, opened 2024-07-23
 staff user backend authentication implementation


isabeleliassen:
## User Story
As a staff member, I want to be able to see and test CompactConnect so that I can understand how the system works and provide feedback.

### Acceptance criteria
- API allows for compact users (boards, commissions, CPCN) to authenticate and have their permissions controlled per the design from #88 
- users can read compact level data and/or write compact level data depending on user
- machine to machine integration not needed; licensee users not needed
- use hosted amazon login for now

Note: work with UI dev team to prepare for UI integration

PR #122, opened 2024-08-05
Feat/user authentication

jusdino:
## Description List
- Added Staff user pool, authn/authz according to access control design (see included docs)
- Integrated existing API authz

## Testing List
- Documentation/design review
- Code review
- PO Demonstration

Closes #
https://github.com/csg-org/CompactConnect/issues/105


backend/compact-connect/lambdas/staff-user-pre-token/tests/test_user_scopes.py
```diff
@@ -0,0 +1,246 @@
...
+            UserScopes(self._user_sub)
+
+    def test_disallowed_compact_action(self):
+        """
+        If a user's permissions list an invalid compact, we will refuse to give them
+        any scopes at all.
+        """
+        from user_scopes import UserScopes
+
+        # Create a DB record with permissions for an unsupported compact
```
jlkravitz, line 176: comment needs updating

backend/compact-connect/lambdas/staff-user-pre-token/tests/test_user_scopes.py
```diff
@@ -0,0 +1,246 @@
...
+            UserScopes(self._user_sub)
+
+    def test_disallowed_jurisdiction(self):
+        """
+        If a user's permissions list an invalid jurisdiction, we will refuse to give them
+        any scopes at all.
+        """
+        from user_scopes import UserScopes
+
+        # Create a DB record with permissions for an unsupported compact
```
jlkravitz, line 203: comment needs updating

backend/compact-connect/lambdas/staff-user-pre-token/tests/test_user_scopes.py
```diff
@@ -0,0 +1,246 @@
...
+            UserScopes(self._user_sub)
+
+    def test_disallowed_action(self):
+        """
+        If a user's permissions list an invalid action, we will refuse to give them
+        any scopes at all.
+        """
+        from user_scopes import UserScopes
+
+        # Create a DB record with permissions for an unsupported compact
```
jlkravitz, line 229: comment needs updating

backend/compact-connect/lambdas/staff-user-pre-token/tests/test_user_scopes.py
```diff
@@ -0,0 +1,246 @@
...
+        self._table.put_item(
+            Item={
+                'pk': self._user_sub,
+                'createdCompactJurisdiction': 'aslp/aslp',
+                'permissions': {
+                    'aslp': {
+                        'actions': {'read'},
+                        'jurisdictions': {
+                            'al': {'actions': {'write', 'hack'}}
```
jlkravitz, line 238: Hmm, perhaps we should add a 'hack' action. Just for fun. (Just kidding.)

backend/compact-connect/tests/unit/test_app.py
```diff
@@ -95,9 +95,28 @@ def test_synth_sandbox_no_domain(self):
         self._check_no_annotations(app.sandbox_stage.persistent_stack)
         self._check_no_annotations(app.sandbox_stage.ui_stack)
         self._check_no_annotations(app.sandbox_stage.api_stack)
+        self._check_no_annotations(app.sandbox_stage.ingest_stack)
```
jlkravitz, line 98: I've seen this before enough times that I gotta ask...what does this do?

backend/compact-connect/tests/unit/test_app.py
```diff
@@ -95,9 +95,28 @@ def test_synth_sandbox_no_domain(self):
...
 
         self._inspect_api_stack(app.sandbox_stage.api_stack)
 
+    def test_synth_no_ui_raises_value_error(self):
```
jlkravitz, line 102: Thanks for testing this!

backend/compact-connect/docs/design/README.md
```diff
@@ -2,7 +2,33 @@
...
+jurisdictions it supports within each compact is all 50 states, Washington D.C., Puerto Rico, and the Virgin Islands.
+
+### Adding a compact to CompactConnect
+
+When a new compact joins CompactConnect, some configuration has to be done to add them to the system. First, a new
+entry has to be added to the list of supported compacts, found in [cdk.json](../../cdk.json). Each compact is
+represented there with an abbreviation, which determines how the compact will be represented in the API as well as
+in its corresponding Oauth2 access scopes. Because of the way that the scopes are represented, the compact abbreviation
+must not overlap with any jurisdiction abbreviations (which correspond to the jurisdictions' USPS postal abbreviations).
+**Since postal abbreviations are all two letters, make a point to choose a compact abbreviation that is at least four
```
jlkravitz, line 22: Perhaps add a quick note to say, "The compact abbreviation does not need to match the one used by CSG." (Or whatever those alternative abbreviations are used for.)

backend/compact-connect/docs/design/README.md
```diff
@@ -31,6 +57,82 @@ efficient processing. A lambda receives messages from the SQS queue. Each messag
...
+Authentication with the CompactConnect backend will be controlled through Oauth2 via
+[AWS Cognito User Pools](https://github.com/csg-org/CompactConnect). Clients will be divided into two groups, each
+represented by an independent User Pool: [Staff Users](#staff-users) and [Licensee Users](#licensee-users). See
+the accompanying [architecture diagram](./users-arch-diagram.pdf) for an illustration.
+
+### Staff Users
+
+Staff users come with a variety of different permissions, depending on their role. There are Compact Executive
+Directors, Compact ED Staff, Board Executive Directors, Board ED Staff, and CSG Admins, each with different levels
+of ability to read and write data, and to administrate users. Read permissions are granted to a user for an entire
```
jlkravitz, line 75: Question: can a user have write permission without read permission? If not, is that something we're testing for?

backend/compact-connect/docs/design/README.md
```diff
@@ -31,6 +57,82 @@ efficient processing. A lambda receives messages from the SQS queue. Each messag
...
+with their own Cognito user id. That record will be used to generate scopes in the Oauth2 token issued to them on login.
+See [Implementation of scopes](#implementation-of-scopes) for a detailed explanation of the design for exactly how
+permissions will be represented by scopes in an access token. See
+[Implementation of permissions](#implementation-of-permissions) for a detailed explanation of the design for exactly
+how permissions are stored and translated into scopes.
+
+#### Compact Executive Directors and Staff
+
+Compact ED level staff can have permission to read all compact data as well as to create and manage users and their
+permissions. They can grant other users the ability to write data for a particular jurisdiction and to create more
```
jlkravitz, line 87: It sounds like the ED level staff will have read and admin permissions.

But as an admin, one can create a user with write permissions, I think.

In that case, it seems like admin and write are implicitly intertwined, no? (If you have admin permissions, just create an account with write permissions.)

jlkravitz: And to clarify: What are those with "write" permissions able to write?

backend/compact-connect/docs/design/README.md
```diff
@@ -31,6 +57,82 @@ efficient processing. A lambda receives messages from the SQS queue. Each messag
...
+users associated with a particular jurisdiction. They can also delete any user within their compact, so long as that
+user does not have permissions associated with a different compact.
```
jlkravitz, line 89: Something to test, perhaps?

backend/compact-connect/docs/design/README.md
```diff
@@ -31,6 +57,82 @@ efficient processing. A lambda receives messages from the SQS queue. Each messag
...
+API that can verify the tokens issued by a given User Pool and to protect access based on scopes belonging to
+[Resource Servers](https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-pools-define-resource-servers.html)
+associated with that User Pool. In the Staff Users user pool, we represent each compact as its own Resource Server, with
+associated scopes. Unfortunately, because resource servers support only up to 100 scopes each, and we would like to
+control permission to write to or administrate each of more than 50 jurisdictions independently, the combinations would
+require more than 100 scopes per resource server.
+
+To design around the 100 scope limit, we will have to split authorization into two layers: coarse- and fine-grained.
+We can rely on the Cognito authorizers to protect our API endpoints based on fewer coarse-grained scopes, then
+protect the more fine-grained access within the API endpoint logic. The Staff User pool resource servers will are
```
jlkravitz, line 110: Nit, maybe silly question: At this point, why not just do all of the access control within the API endpoints? (I would agree this is better, but I wonder what your rationale would be here.)

backend/compact-connect/docs/design/README.md
```diff
@@ -31,6 +57,82 @@ efficient processing. A lambda receives messages from the SQS queue. Each messag
...
+configured with `read`, `write`, and `admin` scopes. `read` scopes indicate that the user is allowed to read the
+entire compact's licensee data. `write` and `admin` scopes, however, indicate only that the user is allowed to write
+or administrate _something_ in the compact respectively, thus giving them access to the write or administrative
+API endpoints. We will then rely on the API endpoint logic to refine their access based on the more fine-grained
+access scopes.
+
+To compliment each of the `write` and `admin` scopes, there will be at least one, more specific, scope, to indicate
```
jlkravitz, line 117: Should the [diagram](https://github.com/csg-org/CompactConnect/blob/7f02302977dae3c7374e49ee0ea3cf07b1b047e9/backend/compact-connect/docs/design/users-arch-diagram.pdf) have a line under `aslp` for the generic `write` scope?

jlkravitz (changes requested):
Generally looks great! I'll go through the rest of the checklist tomorrow. Thanks for the meeting earlier today- very helpful context as I reviewed this PR!

backend/compact-connect/bin/create_staff_user.py
```diff
@@ -0,0 +1,148 @@
...
+                    'jurisdictions': {
+                        jurisdiction: {'actions': {'write', 'admin'}}
+                    }
+                }
+            }
+        }
+    )
+
+
+def create_cognito_user(*, username: str, email: str):
```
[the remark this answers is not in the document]
jusdino: The * forces the arguments to be keyword-only, so that our interface is more explicit in its use. Helps to prevent accidentally getting arguments out of order or confusion in the case of future revisions to the function signature.

backend/compact-connect/bin/create_staff_user.py
```diff
@@ -0,0 +1,148 @@
...
+    def get_sub_from_attributes(attributes: list):
+        for attribute in attributes:
+            if attribute['Name'] == 'sub':
+                return attribute['Value']
+        raise ValueError('Failed to find user sub!')
```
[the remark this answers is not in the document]
jusdino: `sub` is one of the [standard JSON Web Token claims](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.2). The value is passed around between Cognito, its tokens, and the users table. Because `sub` is its standardized name, I'm using that across the breadth of its use, rather than renaming for the various places it winds up being used.

backend/compact-connect/lambdas/license-data/tests/function/test_data_model/test_license_transformations.py
```diff
@@ -32,7 +32,7 @@ def test_transformations(self):
             'jurisdiction': 'co'
         }
         # Authorize ourselves to write an aslp/co license
-        event['requestContext']['authorizer']['claims']['scope'] = 'openid email aslp/co'
+        event['requestContext']['authorizer']['claims']['scope'] = 'openid email aslp/co.write'
```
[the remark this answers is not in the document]
jusdino: What kind of alternative scopes are you thinking of?

backend/compact-connect/lambdas/staff-user-pre-token/.coveragerc
```diff
@@ -0,0 +1,10 @@
+[run]
```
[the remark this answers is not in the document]
jusdino: yep

backend/compact-connect/lambdas/staff-user-pre-token/config.py
```diff
@@ -0,0 +1,33 @@
...
+    def users_table(self):
+        return boto3.resource('dynamodb').Table(self.users_table_name)
+
+    @property
+    def compacts(self):
+        return set(json.loads(os.environ['COMPACTS']))
+
+    @property
+    def jurisdictions(self):
+        return set(json.loads(os.environ['JURISDICTIONS']))
```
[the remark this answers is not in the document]
jusdino: They're the same. This is how we pass the list in `cdk.json` over into the lambda runtime - via [environment variable](https://github.com/InspiringApps/JCC-CompactConnect/blob/7f02302977dae3c7374e49ee0ea3cf07b1b047e9/backend/compact-connect/stacks/persistent_stack/staff_users.py#L116).

backend/compact-connect/lambdas/staff-user-pre-token/user_scopes.py
```diff
@@ -0,0 +1,82 @@
...
+            logger.error('Authenticated user not found!', exc_info=e, sub=sub)
+            raise RuntimeError('Authenticated user not found!') from e
+        return user_data
+
+    def _process_compact_permissions(self, compact_name, compact_permissions):
+        # Compact-level permissions
+        compact_actions = compact_permissions.get('actions', set())
+
+        # Ensure included actions are limited to supported values
+        disallowed_actions = compact_actions - {'read', 'admin'}
```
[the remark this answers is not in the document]
jusdino: The scope structure and the permissions structure are each a little different from each other. This logic is parsing the permission structure in the database, which is designed to reflect the user's actual permissions by business logic (without any reflection of the coarse vs fine grained scope implementation complexity in the middle). As such, there is no `write` permission at the compact level. Write permissions are always specific only to a jurisdiction.

backend/compact-connect/lambdas/staff-user-pre-token/user_scopes.py
```diff
@@ -0,0 +1,82 @@
...
+            raise ValueError(f'User {compact_name} permissions include disallowed actions: {disallowed_actions}')
+
+        # Read is the only truly compact-level permission
+        if 'read' in compact_actions:
+            self.add(f'{compact_name}/read')
+
+        if 'admin' in compact_actions:
+            # Two levels of authz for admin
+            self.add(f'{compact_name}/admin')
+            self.add(f'{compact_name}/{compact_name}.admin')
```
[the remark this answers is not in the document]
jusdino: `compact_name/admin` is the coarse-grained scope, which will grant access to the admin endpoints. `compact_name/compact_name.admin` is the corresponding fine-grained scope, which will allow a request that affects compact-level administrative actions. Both are required for a compact admin to actually be allowed to do anything.

backend/compact-connect/stacks/persistent_stack/board_users.py
```diff
@@ -42,7 +45,9 @@ def __init__(
         self._add_scope_customization()
         # Do not allow resource server scopes via the client - they are assigned via token customization
         # to allow for user attribute-based access
-        self.ui_client = self.add_ui_client()
+        self.ui_client = self.add_ui_client(callback_urls=[
+            'http://localhost:3018/auth/callback'
```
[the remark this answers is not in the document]
jusdino: It's the tcp port that the UI dev team configured the UI to use, locally.

backend/compact-connect/stacks/persistent_stack/board_users.py
```diff
@@ -10,6 +10,9 @@
 
 
 class BoardUsers(UserPool):
+    """
+    DEPRECATED - see comment above constructor use
```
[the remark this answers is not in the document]
jusdino: At the [constructor use](https://github.com/InspiringApps/JCC-CompactConnect/blob/7f02302977dae3c7374e49ee0ea3cf07b1b047e9/backend/compact-connect/stacks/persistent_stack/__init__.py#L78)

backend/compact-connect/stacks/persistent_stack/staff_users.py
```diff
@@ -0,0 +1,134 @@
...
+        self._add_resource_servers()
+        self._add_scope_customization()
+
+        callback_urls = []
+        if stack.ui_domain_name is not None:
+            callback_urls.append(f'https://{stack.ui_domain_name}/auth/callback')
+        # This toggle will allow front-end devs to point their local UI at this environment's user pool to support
+        # authenticated actions.
+        if environment_context.get('allow_local_ui', False):
+            callback_urls.append('http://localhost:3018/auth/callback')
```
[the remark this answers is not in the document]
jusdino: Real domain names are saved in context. This is specifically the local-ui configuration toggle, for developers running the UI on their own machines.

backend/compact-connect/tests/unit/test_app.py
jusdino: We are using the [cdk-nag](https://github.com/cdklabs/cdk-nag) library to sanity-check our infrastructure for security best practices. When `cdk-nag` finds things like a resource without encryption at rest configured, or other general security rules, it adds a finding to the `annotations` for the stack on CDK synthesis. The `_check_no_annotations()` method is looking at `annotations` and asserting that there are no un-suppressed findings outstanding for that stack.

backend/compact-connect/docs/design/README.md
jusdino: Yes, they can. In fact, when we create a machine-to-machine client, that may happen, depending on how boards end up wanting to use that capability.

backend/compact-connect/docs/design/README.md
jusdino: > And to clarify: What are those with "write" permissions able to write?

They can write license/privilege data for their own jurisdiction.

> In that case, it seems like admin and write are implicitly intertwined, no?

Yes. There is a path to a sort of privilege escalation by virtue of a compact ED creating a board user with write permission for the board's jurisdiction. My understanding of the use cases here is that this is desirable. Compact EDs do not have any need or desire to write data for a jurisdiction themselves, but we aren't specifically aiming to prevent them from being able to do so, if they're determined to. Compact EDs _do_, however, need the ability to delegate write permissions to users in all jurisdictions for their compact, by virtue of creating board ED users.

backend/compact-connect/docs/design/README.md
jusdino: Yes, we should absolutely test these things, when we get to the user management feature(s). We haven't taken that on yet.

backend/compact-connect/docs/design/README.md
jusdino: There are a couple of reasons to do it this way:
1) Generally, we want avoid rolling-our-own security as much as is possible. Hand crafting authn/authz logic is intrinsically more risky than relying on dedicated tools that do so for you, so as a general principal, we should defer as much of our authorization to Cognito as we can. Here, we are keeping the most critical pieces in Cognito: Token verification and coarse grained access control. We will only cover fine-grained access controls in our own logic because we can't fit them into Cognito directly.
2) There is a performance benefit to using a dedicated API Gateway authorizer instead of authorizing in runtime: Latency and cost. API Gateway authorizers are very efficient compute-wise in that they sit in front of the compute that we pay for (the endpoint Lambda function) and gate-keep malicious requests such that our compute is never touched. Authorizers can also cache request authorization results so that they don't need to continually be re-validating recently used tokens. Between the gate-keeping and caching, we end up with a much more performant, cost effective API when we use the AWS tools for authorization.

backend/compact-connect/docs/design/README.md
jusdino: Nope. Note that the data structure in the drawing is illustrating the mapping of the data (which more closely reflects the actual business logic) to the scopes (which are really more of an implementation detail). There is no real generic write permissison, so there is no generic write field in the db record. Instead, the `aslp/write` scope is derived together with its fine-grained companion, `aslp/al.write` from the actual permission node in the data structure (such as `permissions.aslp.jurisdictions.al.write`)

backend/compact-connect/bin/create_staff_user.py
```diff
@@ -0,0 +1,148 @@
...
+    sub = create_cognito_user(username=username, email=email)
+    user_table.put_item(
+        Item={
+            'pk': sub,
+            'createdCompactJurisdiction': f'{compact}/{jurisdiction}',
+            'permissions': {
+                compact: {
+                    'actions': {'read'},
+                    'jurisdictions': {
+                        jurisdiction: {'actions': {'write', 'admin'}}
```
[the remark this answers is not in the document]
jusdino: As it happens, we actually are validating those, [here](https://github.com/InspiringApps/JCC-CompactConnect/blob/7f02302977dae3c7374e49ee0ea3cf07b1b047e9/backend/compact-connect/bin/create_staff_user.py#L118)

backend/compact-connect/bin/create_staff_user.py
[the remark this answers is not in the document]
jlkravitz: ah cool! the more you know.

backend/compact-connect/bin/create_staff_user.py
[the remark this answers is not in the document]
jlkravitz: got it. makes sense!

backend/compact-connect/bin/create_staff_user.py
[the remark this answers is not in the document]
jlkravitz: Right. I think my brain thought of the case where this is used as a module, rather than as a CLI, but hopefully no one will make that mistake 😃

backend/compact-connect/lambdas/license-data/tests/function/test_data_model/test_license_transformations.py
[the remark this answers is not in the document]
jlkravitz: e.g., if the scope is `aslp/al.write`, testing the user can't write to `aslp/co`.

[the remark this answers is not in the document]
jlkravitz: e.g., if the scope is `aslp/al.write`, testing the user can't write to `aslp/co`.

backend/compact-connect/lambdas/staff-user-pre-token/user_scopes.py
```diff
@@ -0,0 +1,82 @@
...
+        try:
+            user_data = config.users_table.get_item(Key={'pk': sub})['Item']
+        except KeyError as e:
+            logger.error('Authenticated user not found!', exc_info=e, sub=sub)
+            raise RuntimeError('Authenticated user not found!') from e
+        return user_data
+
+    def _process_compact_permissions(self, compact_name, compact_permissions):
+        # Compact-level permissions
+        compact_actions = compact_permissions.get('actions', set())
```
jlkravitz, line 41: I don't see an `actions` key in the [diagram](https://github.com/csg-org/CompactConnect/blob/7f02302977dae3c7374e49ee0ea3cf07b1b047e9/backend/compact-connect/docs/design/users-arch-diagram.pdf)?

backend/compact-connect/lambdas/staff-user-pre-token/user_scopes.py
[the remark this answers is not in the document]
jlkravitz: Got it!

backend/compact-connect/lambdas/staff-user-pre-token/user_scopes.py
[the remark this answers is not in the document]
jlkravitz: Is the reason we can't just have `compact_name/compact_name.admin` because the resource servers rely on the coarse grained scope to grant access to the admin endpoints? (In my mind, `compact_name/compact_name.admin` implies `compact_name/admin`, so just clarifying the model.)

backend/compact-connect/docs/design/README.md
jlkravitz: Makes a lot of sense. Thanks for the detailed explanation.

backend/compact-connect/docs/design/README.md
jlkravitz: Right, we discussed this. Bear with me for a second, though...(and thank you in advance)...

Am I right to assume that there are endpoints that are designated as write, read, or admin?

For read and admin, isn't it the case that we let the coarse-grained scope authorize access to those read and admin endpoints respectively, even though access is ultimately determined by the granular scopes?

If that's the case, and assuming there _are_ endpoints that are designated as "write", why would we not follow the same model with write?

What makes "write" different?

backend/compact-connect/docs/design/README.md
jlkravitz: Got it!

backend/compact-connect/docs/design/README.md
jlkravitz: Got it, thanks!

Still a little confused on 'write' access. Does 'write' provide permission to add a new license? Edit existing licenses? If edit, which licenses can they edit? All of them? Is the assumption that if I have "write" access on a compact/jurisdiction, then I also have read access on the compact?

backend/compact-connect/tests/unit/test_app.py
jlkravitz: awesome, love that. thank you!

backend/compact-connect/stacks/persistent_stack/board_users.py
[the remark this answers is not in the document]
jlkravitz: hmm, any way to make this reference an env variable? If someone's running locally and has to use a different port, I'd prefer for this not to break.

jlkravitz (commented):
A few more small q's and comments but should be able to get this merged today!

jlkravitz:
- [ ] stand up the site locally
	- [ ] test all functionality in all major browsers, emphasizing the functionality that this pull request addresses
		- [ ] for public-facing functionality, test in browsers consistent with [public browser use data](https://analytics.usa.gov/)
		- [ ] test in Mobile Safari and Mobile Chrome
	- [ ] use an automated audit tool for code quality and practices (recommended: [Chrome DevTools](https://developers.google.com/web/tools/chrome-devtools/), aka [Lighthouse](https://chrome.google.com/webstore/detail/lighthouse/blipmdconlkpinefehnmjammfjpmpbjk))
		- [ ] look at efficiency of page loads, asset sizes, HTTP connection management, etc.
	- [ ] review for accessibility
		- [ ] use an automated audit tool, such as Chrome Audit or aXe
		- [ ] navigate site only with the keyboard
		- [ ] use VoiceOver or Narrator to navigate the site with audio only, with the display turned off
		- [ ] manually test anything that pa11y cannot test automatically (e.g., contrast of text over images)
- [ ] review static code analysis results
- [ ] run a security audit of dependencies (e.g. `npm audit`) to ensure that there are no vulnerabilities that will be deployed to production (as opposed to vulnerabilities that only have an impact on the development environment)
- [ ] review all tests

backend/compact-connect/lambdas/license-data/tests/function/test_data_model/test_license_transformations.py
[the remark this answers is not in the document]
jusdino: Got it. That's actually handled [over here](https://github.com/InspiringApps/JCC-CompactConnect/blob/6a419706c97e6eee470dc534d2bed7668f50f9c3/backend/compact-connect/lambdas/license-data/tests/unit/test_handlers/test_licenses.py#L45)

backend/compact-connect/lambdas/staff-user-pre-token/user_scopes.py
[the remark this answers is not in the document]
jusdino: Yeah, that's right. The resource servers aren't actually configured with the `compact_name/compact_name.admin` scope, so they rely on the coarse-grained `compact_name/admin` scope to protect the endpoint, then our logic will control the ability to make actual changes to users/permissions related to the compact based on the `compact_name/compact_name.admin` scope.

backend/compact-connect/stacks/persistent_stack/board_users.py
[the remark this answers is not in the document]
jusdino: Sure, we can set it up as an override-able default based on an optional context variable. I'll add that in shortly.

backend/compact-connect/docs/design/README.md
jusdino: So 'write' and 'read' permissions are entirely independent, the way this is set up. If you have 'read' access, you can read all license/privilege data for the whole compact. 'write' permission is specific to a jurisdiction. If a user has 'write' permissions for the Alabama jurisdiction for the Audiology/Speech Language Pathology compact, they will be able to create/update licenses, update privileges, but only within the Alabama jurisdiction, for ASLP.

backend/compact-connect/stacks/persistent_stack/board_users.py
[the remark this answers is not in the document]
jusdino: FYI - I'll be adding this change to the StaffUsers user pool, not the Board Users one, since we're just working on tearing down the BoardUsers, anyway.

backend/compact-connect/docs/design/README.md
jusdino: > Am I right to assume that there are endpoints that are designated as write, read, or admin?

Yes.

> What makes "write" different?

'write' and 'admin' are the same in their approach to authorization, each with a coarse-grained scope that controls endpoint access and a fine-grained scope that ultimately determines real access. The only difference between 'write' and 'admin' that I can think of is that 'admin' has fine-grained scopes for jurisdiction-admin and one for compact-admin, whereas 'write' only has fine-grained scopes for jurisdiction-write. There is no logical write permission equivalent to compact-admin (there is no compact-level data to write).

backend/compact-connect/lambdas/staff-user-pre-token/user_scopes.py
jusdino: This is true. The diagram is more of a conceptual model of how the permissions may be represented via API than an actual schema for the record in the database.

backend/compact-connect/docs/design/README.md
jlkravitz: Ok, I think I'm 80% there but will table this for a future discussion :)

I think part of my confusion is that I assumed the `read` and `write` compact-level actions are what directly indicated the _coarse-grained_ scopes, but I'm realizing that that's not the case. The compact-level actions exist precisely because there is data that is specific to a _compact_ but _not_ to a jurisdiction that could have "read" or "write" permissions. I've been over-conflating the diagram's json schema with the scopes themselves.

jlkravitz (approved):
Looks good. Thanks for answering all of my questions! @isabeleliassen This is good to go. Once this is merged, I will conduct security and accessibility tests on the development branch for last sprint's work.

PR #138, opened 2024-08-13
Update CODE_REVIEW.md

jlkravitz:
Updated to better reflect the how we work document. This is more for me than anyone else, so that I can copy the most common checklist items for PRs.

### DIFF
--- a/docs/CODE_REVIEW.md
+++ b/docs/CODE_REVIEW.md
@@ -85,12 +85,27 @@ We wouldn’t want anyone being mean to us because of an oversight, mistake, or 
 
 ## Checklist
 
-We use this list when performing a code review to ensure that all tasks have been completed.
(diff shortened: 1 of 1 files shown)


jusdino (approved):
This all seems reasonable to me. I think it's mostly your prerogative to decide how to best use your time conducting these code reviews so I'm just looking for anything that looks like a major gap or other large concerns.

jlkravitz:
> This all seems reasonable to me. I think it's mostly your prerogative to decide how to best use your time conducting these code reviews so I'm just looking for anything that looks like a major gap or other large concerns.

Sounds great! That was mostly my intent- share with you (1) since I consider IA to be the working owner of the codebase and (2) to have another eye flag anything that looks suspect. For something like this, I assumed it'd be a non-issue, but I figure it's best to be consistent from a process perspective.

isabeleliassen (approved):

PR #137, opened 2024-08-13
Sprint 5: Development to Production

jlkravitz:
## Waiting on

- [x] syncing development branch back up with main #141

PR #115, opened 2024-07-31
Update HOW_WE_WORK.md

jlkravitz:
Updated `HOW_WE_WORK.md` based on discussion and my experience this past sprint. Feedback welcome!

PR #141, opened 2024-08-13
Merge main with development branch

jlkravitz:
Syncs changes we accidentally merged into `main` back into `development`.


HOW_WE_WORK.md
```diff
@@ -0,0 +1,13 @@
+# How We Work
```
jusdino, line 1: This is an outdated duplicate of [this doc](https://github.com/csg-org/CompactConnect/blob/development/docs/HOW_WE_WORK.md). We should probably remove it.

backend/compact-connect/docs/postman/postman-environment.json
```diff
@@ -1,6 +1,6 @@
 {
 	"id": "65234e00-5ac9-4819-8620-d1b9076dcbce",
-	"name": "JCC License Data - sandbox",
+	"name": "License Data - sandbox",
```
jusdino, line 3: This appears to just be rolling back a change made in a more recent commit.

jusdino (changes requested):
I don't think this PR is accomplishing what you hope. Both of these changes appear to be just rolling back more recent changes, rather than merging in out-of-band changes to `main`:
- The HOW_WE_WORK.md doc was moved into `docs/`, this is just recreating an older version in its original location.
- The postman environment name was updated in a more recent commit

I'd suggest we close this PR as there don't appear to be any other items out of sync between `main` and `development`.

HOW_WE_WORK.md
jlkravitz: I don't think that's quite right, but I may be wrong. #115 was merged into main, rather than development. This document includes the per-sprint language we discussed in that PR, while the one you linked does not.

jlkravitz:
> I don't think this PR is accomplishing what you hope. Both of these changes appear to be just rolling back more recent changes, rather than merging in out-of-band changes to `main`:
> 
> * The HOW_WE_WORK.md doc was moved into `docs/`, this is just recreating an older version in its original location.
> * The postman environment name was updated in a more recent commit
> 
> I'd suggest we close this PR as there don't appear to be any other items out of sync between `main` and `development`.

I don't think that's quite right, but I may be wrong. https://github.com/csg-org/CompactConnect/pull/115 was merged into main, rather than development. This document includes the per-sprint language we discussed in that PR, while the one you linked does not. I made some updates to merge the two with the API design language, too, which was not in `main`.

At the moment #137 has many conflicts, when it shouldn't.

jusdino (approved):
OK, this looks good, now. Thanks for the rehash. 👍

jlkravitz:
Great! @isabeleliassen Feel free to approve and merge when you're able.

isabeleliassen (approved):

PR #157, opened 2024-08-14
Bug/older f string

jusdino:
CodeBuild, where our CDK code is run, is not actually using the latest python version (3.12) on its latest platform version. I discovered this when our CDK app stopped synthesizing this morning when it failed to recognize a python 3.12 f-string feature use, introduced in the monitoring-and-alerting feature. This PR resolves that issue.

### Description List
- ~Remove use of python 3.12 f-string feature use from CDK code~ *specify python version in the build environment*
- Add `python --version` call to ShellStep to explicitly print out python version in use
- Remove main/development PR branch filter for python checks github action (not strictly related to bug-fix but facilitating more consistent check executions)

### DIFF
--- a/.github/workflows/check-python.yml
+++ b/.github/workflows/check-python.yml
@@ -2,9 +2,6 @@ name: Check-Python
 
 on:
   pull_request:
-    branches:
-      - main
-      - development
     paths:
       - backend/**
 
--- a/backend/compact-connect/common_constructs/alarm_topic.py
+++ b/backend/compact-connect/common_constructs/alarm_topic.py
@@ -34,7 +34,7 @@ class AlarmTopic(Topic):
         self.slack_channel_integrations = {}
         for config in slack_subscriptions:
             self.slack_channel_integrations[config['channel_name']] = SlackChannelConfiguration(
-                self, f'{config['channel_name']}-SlackChannelConfiguration',
+                self, f'{config["channel_name"]}-SlackChannelConfiguration',
                 notification_topics=[self],
                 workspace_id=config['workspace_id'],
                 channel_id=config['channel_id']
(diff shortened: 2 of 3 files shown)


jlkravitz (commented):
Could we instead explicitly set the python version per [this guide](https://docs.aws.amazon.com/codebuild/latest/userguide/runtime-versions.html)?

Even better, could we tie the python version to whatever we're expecting? (It seems you can specify arbitrary python versions, but I may be wrong. I read quickly.)

jusdino:
In theory, we could and I would prefer that approach as well. The difficulty comes into the fact that we've got a couple of levels of abstraction built on top of the actual CodeBuild project that don't readily support adding that python version argument. We are defining this synthesis step via the [ShellStep](https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.pipelines/ShellStep.html) construct. I'm not finding a reasonable way to inject that python version argument to the underlying build spec in a way that will not be un-intuitive and/or brittle. Instead, I settled for using no 3.12 features in the CDK code at this time, since we don't especially need them.

jusdino:
Oh, actually, I may have found a better path - let's hold on this one while I check things out.

jlkravitz:
Hmm. This feels brittle in its own  right, no? It's unclear to me which python version the CodeBuild uses by default, so it seems we could run into this issue with any python feature incompatibilities, which we wouldn't discover until runtime. Can we track this in the backlog? I'm fine with merging now to fix this issue.

jusdino:
Ok, here's a better approach. Turns out I was looking in the wrong direction when trying to find a good way to specify that python runtime. This new approach is now tested on the IA pipeline.

jlkravitz (approved):
Looks great! Do we specify a required python version anywhere else in the backend codebase for local development or lambdas, etc? Wondering if we'll run into similar issues moving forward.

jusdino:
Everything else was already defined as Python 3.12 - I made the mistake of assuming that the latest build platform meant Python 3.12 for CodeBuild as well, but it turns out they're still on 3.11 if you don't specify. That's now fixed.

jlkravitz:

PR #212, opened 2024-09-25
Update check-branch.yml to fetch full git history

jlkravitz:
Only the most recent development commit is fetched by default within Github Actions. This clones the entire git history.


jusdino (approved):
Ohhh, that makes sense. Good find!

PR #218, opened 2024-09-30
Feat/add provider users profile endpoint

landonshumway-ia:
This adds a new endpoint to allow provider users that have logged in to see their applicant dashboard.
When a provider user is created, the process that created them will be responsible for setting their provider id 
and their compact as custom attributes on their cognito user identity. These custom attributes will be used to
query the database for their personal information.

### Requirements List
- PR #208 must be merged before this

### Description List
- Added handler to get provider's compact and provider id from their cognito user claims and query for their information.
- Added cdk setup to attach the lambda to the api gateway to process requests.
- Refactor api gateway model definitions to be centralized in one class
- Refactor CDK tests to use robust logical id checks

### Testing List
- `yarn test:unit:all` should run without errors or warnings
- `yarn serve` should run without errors or warnings
- `yarn build` should run without errors or warnings
- Code review

Closes #202

PR #220, opened 2024-10-01
Feat/staff user management api

### LINKED ISSUE #147: backend API for CSG staff user management
## User Story
As a superadmin, I want to be able to see all other access permissions so I can step in to resolve problems when necessary.

## Acceptance Criteria
- superadmins can see all data in the system (read and admin permissions)
- log user actions for traceability--is this possible?
- no need to directly impersonate other users
- no need to create new compacts
- ability to add/remove/disable other levels of users
- if an existing user who is disabled is re-invited, they will be re-enabled
- all invited users including re-invites should include email notification to create a password

Reference #125

jusdino:
## Requirements List
- Added staff user management API for admins and users to self-manage attributes

## Closes #
- Closes #147


jlkravitz:
- [ ] review all tests
- [ ] run a security audit of dependencies (e.g. `npm audit` and `pip audit`) to ensure that there are no vulnerabilities that will be deployed to production (as opposed to vulnerabilities that only have an impact on the development environment)

backend/compact-connect/lambdas/staff-users/utils.py
```diff
@@ -0,0 +1,263 @@
+import json
```
jlkravitz, line 1: I feel like I've seen this code, and some of the code from other staff-user files. Is that true? Are these copies, or is the code being moved?

backend/compact-connect/stacks/api_stack/v1_api/staff_users.py
```diff
@@ -0,0 +1,662 @@
...
+
+        self.log_groups = []
+        env_vars = {
+            **self.stack.common_env_vars,
+            'USER_POOL_ID': persistent_stack.staff_users.user_pool_id,
+            'USERS_TABLE_NAME': persistent_stack.staff_users.user_table.table_name,
+            'FAM_GIV_INDEX_NAME': persistent_stack.staff_users.user_table.family_given_index_name
+        }
+
+        # .../
```
jlkravitz, line 43: What's this comment meant to imply?

backend/compact-connect/stacks/api_stack/v1_api/staff_users.py
```diff
@@ -0,0 +1,662 @@
...
+
+    @property
+    def _common_user_properties(self):
+        return {
+            'attributes': self._attributes_schema,
+            'permissions': self._permissions_schema
+        }
+
+    @property
+    def _user_response_schema(self):
```
jlkravitz, line 630: If I understood correctly, I think @landonshumway-ia did some consolidation of these schemas in #218. Is there a plan to merge?

backend/compact-connect/lambdas/provider-data-v1/tests/unit/test_data_model/test_paginated.py
```diff
@@ -79,27 +79,33 @@ def test_multiple_internal_pages(self):
...
-                    'pk': self._item['pk'],
-                    'sk': self._item['sk']
-                }
+
+            last_key = int(kwargs['dynamo_pagination'].get('ExclusiveStartKey', {}).get('pk', 0))
+            resp = {
+                'Items': [],
+                'Count': 3
+            }
+            # 2 items, starting after last_key
```
jlkravitz, line 96: isn't this 3 items?

backend/compact-connect/lambdas/provider-data-v1/tests/unit/test_data_model/test_paginated.py
```diff
@@ -110,22 +116,28 @@ def get_something(*args, **kwargs):
...
-                    'prevLastKey': last_key
-                },
-            },
-            resp
-        )
-        # 2 calls, each returning 6 items, will fill out the page size.
+        # We are requesting 10 users, starting with exclusive key 1, and filtering out all odds client-side. This
+        # should result in three queries to the DB, with the last record included in the response having a pk of 20:
+        #
+        # | Query | DB sequence | PK | Ret Sequence |  Filter   | last_key |
```
jlkravitz, line 122: I'm confused by this table. What is DB sequence and return Sequence? 

Also, where are the records being filtered in the tests? I can't seem to find that.

backend/compact-connect/lambdas/staff-users/tests/function/test_data_model/test_client.py
```diff
@@ -0,0 +1,442 @@
...
+        self._create_board_staff_users(compacts=['aslp', 'octp'])
+
+        from data_model.client import UserClient
+
+        client = UserClient(self.config)
+
+        # Provide a client filter that will filter out users without permissions in the jurisdiction we're looking for
+        resp = client.get_users_sorted_by_family_name(  # pylint: disable=missing-kwoa
+            compact='aslp',
+            # All three jurisdictions, this time
```
jlkravitz, line 123: update comment?

backend/compact-connect/lambdas/staff-users/tests/function/test_data_model/test_client.py
```diff
@@ -0,0 +1,442 @@
...
+            jurisdictions=['oh']
+        )
+
+        # We created two board users that have aslp permissions in oh so those are what we should get back
+        self.assertEqual(2, len(resp['items']))
```
jlkravitz, line 128: Why is this just 2 vs line 94, which is 2*the number of jurisdictions?

backend/compact-connect/lambdas/staff-users/tests/function/test_data_model/test_client.py
```diff
@@ -0,0 +1,442 @@
...
+
+        resp = client.update_user_attributes(
+            user_id=user_id,
+            attributes={
+                'givenName': 'Bob',
+                'familyName': 'Smith'
+            }
+        )
+        self.assertEqual(1, len(resp))
+        user = resp[0]
```
jlkravitz, line 267: why would it ever return more than one item (the user)?

backend/compact-connect/lambdas/staff-users/tests/function/test_data_model/test_client.py
```diff
@@ -0,0 +1,442 @@
...
+            {
+                'actions': {'read'},
+                'jurisdictions': {
+                    'oh': {'write', 'admin'}
+                }
+            },
+            resp['permissions']
+        )
+
+    def test_create_existing_user(self):
```
jlkravitz, line 325: A bit confused re: expected behavior here?

backend/compact-connect/lambdas/staff-users/tests/function/test_handlers/test_get_user.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def test_get_user_outside_jurisdiction(self):
+        self._load_user_data()
+
+        from handlers.users import get_one_user
+
+        with open('tests/resources/api-event.json', 'r') as f:
+            event = json.load(f)
+
+        # The user has admin permission for aslp/ne, user does not have aslp/ne permissions
```
jlkravitz, line 97: Do you mean ...
```suggestion
        # The user has admin permission for aslp/ne, user does not have aslp/oh permissions
```

backend/compact-connect/lambdas/staff-users/tests/function/test_handlers/test_get_user.py
```diff
@@ -0,0 +1,107 @@
...
+
+        body = json.loads(resp['body'])
+
+        self.assertEqual(
+            expected_user,
+            body
+        )
+
+    def test_get_user_jurisdiction_admin(self):
+        self._load_user_data()
```
jlkravitz, line 60: I assume this user has `aslp/oh` permissions? Can we add a brief comment somewhere?

backend/compact-connect/lambdas/staff-users/tests/function/test_handlers/test_get_users.py
```diff
@@ -0,0 +1,118 @@
...
+        # The user has admin permission for all of octp (but not aslp)
+        event['requestContext']['authorizer']['claims']['scope'] = 'openid email octp/admin octp/octp.admin'
+        event['pathParameters'] = {
+            'compact': 'aslp'
+        }
+        event['body'] = None
+
+        resp = get_users(event, self.mock_context)
+
+        self.assertEqual(403, resp['statusCode'])
```
jlkravitz, line 118: Why not just return an empty list in this case?

backend/compact-connect/lambdas/staff-users/tests/function/test_handlers/test_patch_user.py
```diff
@@ -0,0 +1,69 @@
...
+class TestPatchUser(TstFunction):
+    def test_patch_user(self):
+        self._load_user_data()
+
+        from handlers.users import patch_user
+
+        with open('tests/resources/api-event.json', 'r') as f:
+            event = json.load(f)
+
+        # The user has admin permission for all of aslp
```
jlkravitz, line 18: Do we (and should we) test ability of non-admin users to patch other user data?

backend/compact-connect/lambdas/staff-users/tests/unit/test_api_handler.py
```diff
@@ -0,0 +1,140 @@
...
+        def lambda_handler(event: dict, context: LambdaContext):
+            raise RuntimeError('Egads!')
+
+        with open('tests/resources/api-event.json', 'r') as f:
+            event = json.load(f)
+
+        with self.assertRaises(RuntimeError):
+            lambda_handler(event, self.mock_context)
+
+    def test_null_headers(self):
```
jlkravitz, line 126: What is this conceptually testing, and why should it still be successful? (Can we maybe add this as a brief comment?)

backend/compact-connect/lambdas/staff-users/tests/unit/test_collect_changes.py
```diff
@@ -0,0 +1,161 @@
...
+            },
+            resp
+        )
+
+    def test_disallowed_jurisdiction_changes(self):
+        from exceptions import CCAccessDeniedException
+        from utils import collect_changes
+
+        with self.assertRaises(CCAccessDeniedException):
+            collect_changes(
```
jlkravitz, line 69: Doesn't `collect_changes` just produce a new object, in that it's a pure function? I'd expect the function to throw these errors if it was actually _making_ the changes in the DB, but not just if it were creating an object for updating. (This might just be a design decision you made, which is totally fine. Just clarifying my understanding.)

jlkravitz (changes requested):
woo! what an adventure. generally: thank you for the thorough comments and testing. The permissions are complex, but the tests make me feel confident that they work properly, and the comments always seemed to anticipate many of my questions :)

Despite that, still several questions...but hopefully we can resolve everything quickly!

backend/compact-connect/common_constructs/stack.py
```diff
@@ -64,6 +64,14 @@ def __init__(self, *args, standard_tags: StandardTags, **kwargs):
...
                     'id': 'HIPAA.Security-LambdaConcurrency',
                     'reason': 'The lambdas in this app will share account-wide concurrency limits'
+                },
+                {
+                    'id': 'AwsSolutions-APIG4',
+                    'reason': 'Temporarily turning off auth for OWASP Zap'
+                },
+                {
+                    'id': 'AwsSolutions-COG4',
+                    'reason': 'Temporarily turning off auth for OWASP Zap'
```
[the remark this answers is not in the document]
jusdino: Ut oh. Those are strays from when I was helping you with Zap. I thought I'd removed that.

backend/compact-connect/lambdas/provider-data-v1/data_model/query_paginator.py
```diff
@@ -102,6 +102,7 @@ def _generate_pages(self, last_key: str | None, page_size: int, args, kwargs):
 
             raw_resp = self._caught_query(*args, dynamo_pagination=dynamo_pagination, **kwargs)
             count += raw_resp['Count']
+            last_key = raw_resp.get('LastEvaluatedKey')
```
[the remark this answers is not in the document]
jusdino: Yep. I tacked on a test and fix while I worked.

backend/compact-connect/lambdas/staff-user-pre-token/tests/test_user_scopes.py
```diff
@@ -18,13 +18,12 @@ def test_compact_ed_user(self):
...
+                'pk': f'USER#{self._user_sub}',
+                'sk': 'COMPACT#aslp',
+                'compact': 'aslp',
                 'permissions': {
-                    'aslp': {
-                        'actions': {'read', 'admin'},
-                        'jurisdictions': {}
-                    }
+                    'actions': {'read', 'admin'},
+                    'jurisdictions': {}
```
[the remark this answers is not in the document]
jusdino: Yes. Essentially, in order to take advantage of DynamoDB's UpdateItem API in a way that is both simple and idempotent, I needed to flatten the fields out a little.

backend/compact-connect/lambdas/staff-users/data_model/client.py
```diff
@@ -0,0 +1,294 @@
...
+        """
+        logger.info('Getting staff users by family name')
+
+        filter_expression = None
+        if jurisdictions:
+            # Form an attribute_exists(<attr>) OR attribute_exists(<attr>) OR ... expression for each jurisdiction
+            iter_jurisdictions = iter(jurisdictions)
+            jurisdiction = next(iter_jurisdictions)
+            filter_expression = Attr(f'permissions.jurisdictions.{jurisdiction}').exists()
+            while jurisdiction :=  next(iter_jurisdictions, None):
```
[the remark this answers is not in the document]
jusdino: Good point. I don't remember why I moved to a `while`, but I don't think the reason is valid anymore.

backend/compact-connect/lambdas/staff-users/data_model/client.py
```diff
@@ -0,0 +1,294 @@
...
+            'delete': []
+        }
+        expression_attribute_names = {}
+        expression_attribute_values = {}
+
+        if compact_action_additions:
+            update_expression_parts['add'].append('#permissions.#actions :addActions')
+            expression_attribute_names['#permissions'] = 'permissions'
+            expression_attribute_names['#actions'] = 'actions'
+            expression_attribute_values[':addActions'] = compact_action_additions
```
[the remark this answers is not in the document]
jusdino: Yeah... Gnarly language but pretty handy.

backend/compact-connect/lambdas/staff-users/data_model/schema/user.py
```diff
@@ -0,0 +1,167 @@
...
+                        for action in jurisdiction_permissions
+                    }
+                }
+        data['permissions'] = user_permissions
+
+        return data
+
+    @post_dump  # Note _post_ dump, so after any type conversions happen, in this case
+    def transform_to_dynamo_permissions(self, data, **kwargs):  # pylint: disable=unused-argument
+        # { "permissions": { "aslp": { ... } } } -> { "compact": "aslp", "permissions": { ... } }
```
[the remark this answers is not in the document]
jusdino: Thank Amazon Q 😁 . AI suggested this succinct way of describing what was going on here.

backend/compact-connect/stacks/persistent_stack/__init__.py
```diff
@@ -81,13 +82,15 @@ def __init__(
...
+
         self.staff_users = StaffUsers(
-            self, 'StaffUsers',
+            self, 'StaffUsersGreen',
             cognito_domain_prefix=staff_prefix if environment_name == 'prod'
             else f'{staff_prefix}-{environment_name}',
             environment_name=environment_name,
             environment_context=environment_context,
             encryption_key=self.shared_encryption_key,
+            # user_invitation=UserInvitationConfig(...),
```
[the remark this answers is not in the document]
jusdino: Not anymore. Will remove.

backend/compact-connect/stacks/persistent_stack/__init__.py
```diff
@@ -81,13 +82,15 @@ def __init__(
             user_pool_email_settings = UserPoolEmail.with_cognito()
 
         staff_prefix = f'{app_name}-staff'
+
         self.staff_users = StaffUsers(
-            self, 'StaffUsers',
+            self, 'StaffUsersGreen',
```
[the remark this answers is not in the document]
jusdino: I had to rename the user pool to avoid resource naming conflicts when the deploy creates a new user pool with the current updates. 'Green' here is a reference to a kind of 'Blue/Green deploy' strategy.

backend/compact-connect/lambdas/staff-users/utils.py
jusdino: Copies. A lot of the code in here is nearly or completely identical to the code in the `provider-data-v1` lambda. We can't (easily) reuse the code across those lambdas because they are separate run-times, so our options are:
1) Copy/paste some common code across the independent runtimes or
2) Add extra infrastructure and complexity specifically for the purpose of sharing common code across lambdas

Option (2) comes with a high level of additional complexity in infrastructure and especially in how we automate testing of the lambdas with multiple source bundles. It's doable but, so long as the shared/duplicated code is not huge, it probably isn't worth going down that road.

backend/compact-connect/stacks/api_stack/v1_api/staff_users.py
jusdino: This and the next couple of comments are just noting which url path segment the methods are creating. So this section is creating `POST` and `GET` methods right at the level of the API resource that was passed into the constructor.

backend/compact-connect/stacks/api_stack/v1_api/staff_users.py
jusdino: Yes. I would like to update this to adopt @landonshumway-ia's newer pattern but didn't do that for this PR, since I was working this in parallel with him establishing the pattern. Same goes for his new CDK tests.

backend/compact-connect/lambdas/provider-data-v1/tests/unit/test_data_model/test_paginated.py
jusdino: Yep. My bad.

backend/compact-connect/lambdas/provider-data-v1/tests/unit/test_data_model/test_paginated.py
jusdino: Oh, I didn't end up doing client-side filtering here so I should remove that part of the comment.

As for the table, this is trying to walk through the sequence of queries to the DB vs the items our backend returns to the client, to demonstrate expected internal behavior. `Query` is the number of the query the backend makes to the DB. `DB sequence` is the position in the returned records each item has for that query:
```
        # | Query | DB sequence | PK |
        # |-------|-------------|----|
        # |   1   |     1       |  2 | # First item returned in the first query to the DB
        # |   1   |     2       |  3 | # Second item returned in the first query to the DB
        # |   1   |     3       |  4 |
        # |   2   |     1       |  5 | # First item returned in the second query to the DB
        # |   2   |     2       |  6 | # Second item returned in the second query to the DB
        ...
```

Similarly, `Ret Sequence` is then denoting the position in the array of results actually returned from the backend to the client.

backend/compact-connect/lambdas/staff-users/tests/function/test_data_model/test_client.py
jusdino: In this case, we are filtering the results down to a single jurisdiction (`oh`) but in line 94, we are looking at each of three different jurisdictions (there are only three jurisdictions (`oh`, `ne`, `ky`) in this test suite, to keep size reasonable).

backend/compact-connect/lambdas/staff-users/tests/function/test_data_model/test_client.py
jusdino: A user has a record in the DB for each compact they have permissions for. In this case, the user is only in `aslp`, so they should only get one record back.

backend/compact-connect/lambdas/staff-users/tests/function/test_data_model/test_client.py
jusdino: In the case that two compact admins invite the same individual to their respective compacts, we want there to only be a single Cognito user created, but two database records for that user (one for each compact). The idea is that the actual creation of a new user vs adding permissions to an existing user should be be invisible to admins across compacts. So what we're looking at here is that we have two sets of permissions (DB records, internally) but that they share the same `userId`.

backend/compact-connect/lambdas/staff-users/tests/function/test_handlers/test_get_user.py
jusdino: I sure do 🤦

backend/compact-connect/lambdas/staff-users/tests/function/test_handlers/test_get_user.py
jusdino: Yes and yes.

backend/compact-connect/lambdas/staff-users/tests/function/test_handlers/test_get_users.py
jusdino: Because in this case, the caller is explicitly requesting to see users that they are not allowed to see. In the similar test case for a jurisdiction admin requesting to see users, they are still requesting users for inside their own compact, so there is nothing illegal about their request, so we just filter out the ones they shouldn't see.

backend/compact-connect/lambdas/staff-users/tests/function/test_handlers/test_patch_user.py
jusdino: You mean like an admin trying to patch user attributes instead of permissions? I suppose we could, though it would take some serious refactoring of the app to the extent that it would be malicious for something like that to become possible since:
1) the API schema does not allow for user attributes to even be provided in the request body for this endpoint;
2) this endpoint doesn't even look for `attributes` if they got past schema validation, so it would be ignored; and
3) the `UserClient` method for updating attributes is separate from the one that is used to update permissions and is not called here anyway.

backend/compact-connect/lambdas/staff-users/tests/unit/test_api_handler.py
jusdino: Sure, I can add a comment. API Gateway has the annoying behavior that, if what is normally a map in its event payload is entirely empty, it will sometimes set the field to `null` instead of just letting it be empty or absent. This test is checking that we've handled that case for headers, just in case that happens.

backend/compact-connect/lambdas/staff-users/tests/unit/test_collect_changes.py
jusdino: The design choice I made was to tightly couple access control with the logic that interprets the changes, so that it would be harder to accidentally leave an opening in the permission model compared to if the two evaluations were independent. In the case that one of the changes identified is not allowed based on the provided scopes, the function will raise an exception. I should probably rename this function to make that behavior more clear.

backend/compact-connect/stacks/persistent_stack/__init__.py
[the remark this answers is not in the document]
jlkravitz: Got it thanks!

backend/compact-connect/lambdas/staff-users/utils.py
jlkravitz: Totally agree with your assessment. Just wanted to validate my deja vu.

backend/compact-connect/stacks/api_stack/v1_api/staff_users.py
jlkravitz: got it. thanks for changing to base-url. that makes sense!

backend/compact-connect/stacks/api_stack/v1_api/staff_users.py
jlkravitz: got it.

backend/compact-connect/lambdas/provider-data-v1/tests/unit/test_data_model/test_paginated.py
jlkravitz: Got it, helpful. Thank you!

And re: client side filtering. That makes sense... I was going a little wild trying to find where that was happening. :)

backend/compact-connect/lambdas/staff-users/tests/unit/test_collect_changes.py
jlkravitz: Got it! Thanks for the explanation. Makes sense.

backend/compact-connect/lambdas/staff-users/tests/unit/test_api_handler.py
jlkravitz: Got it!

backend/compact-connect/lambdas/staff-users/tests/function/test_handlers/test_patch_user.py
jlkravitz: Ah sorry, I mean a non-admin (e.g., a provider) trying to hit this endpoint.

backend/compact-connect/lambdas/staff-users/tests/function/test_data_model/test_client.py
jlkravitz: ah got it. so still that same calculation, just putting the number down explicitly

backend/compact-connect/lambdas/staff-users/tests/function/test_data_model/test_client.py
jlkravitz: Makes sense. Mind adding a brief comment? I think that's not 100% clear from this. A trimmed down version of that explanation is great.

backend/compact-connect/lambdas/staff-users/tests/function/test_handlers/test_patch_user.py
jusdino: Oh, right. Yes, yes we should.

jlkravitz (approved):

isabeleliassen (approved):

PR #221, opened 2024-10-02
Bugfix: validate licenseType on POST licenses schema

jusdino:
I discovered a bug in the API validation while I was doing unrelated things. This closes a testing gap and fixes the bug.

### Description List
- Add function tests of the post_licenses handler
- Move license type validation to parent schema class (`LicensePostSchema`)


jlkravitz (approved):
lgtm! good catch.

Issue #227, opened 2024-10-03
Automated ZAP scanning


jlkravitz:
Re-open of #28

## Full User Story

As CSG technical staff, I want to see the results of ZAP security testing automatically on all PRs to main.

### Acceptance Criteria
- IA and CSG agree on parameters and configuration for ZAP security testing
- When any PR opens against main, ZAP automatically runs on https://app.test.compactconnect.org/ (I think this is what makes most sense given current setup). ZAP runs on all web pages as an authenticated user.
- ZAP failures are presented as Github Action failures

### Discussion items
- Can this run on all PRs without PR preview environments?
- How do we handle false-positives? Documentation? Suppress manually for future runs? Something else?

### Useful links / document
* 18F implemented automated ZAP testing (and other automated tests) on their US Tax Court repository: https://github.com/ustaxcourt/ef-cms

PR #231, opened 2024-10-08
Sprint 9

jlkravitz:


.github/pull_request_template.md
```diff
@@ -8,6 +8,7 @@
 - `yarn test:unit:all` should run without errors or warnings
```
jlkravitz, line 8: It seems SSN should be obscured as a password field given its sensitivity, no? Or, even, better all but the last 4 digits are obscured.

.github/pull_request_template.md
```diff
@@ -8,6 +8,7 @@
...
 - `yarn serve` should run without errors or warnings
```
jlkravitz, line 9: * The search box seems to go off of the main part of the page. Intentional, or is this just a temp design? The page also "jumps" when I open and close the search modal.
image

* Similarly on mobile...
image

.github/pull_request_template.md
```diff
@@ -8,6 +8,7 @@
...
 - `yarn build` should run without errors or warnings
```
jlkravitz, line 10: * First/last name search does not work for me. Intended/expected?

.github/pull_request_template.md
```diff
@@ -8,6 +8,7 @@
...
+- For API configuration changes: CDK tests added/updated in `backend/compact-connect/tests/unit/test_api.py`
```
jlkravitz, line 11: Home page pops up with search, but what if I just want to look at the list of providers? It's not obvious how to do that. (Was it an intended design choice to show the search modal on load?) From an accessibility POV, it's unclear how to close the modal and show the results, too. When I use voiceover, nothing obviously allows me to see results.

jlkravitz (commented):
Made comments on arbitrary files so that they can be individually resolved. Few small questions/issues. @jsandoval81 @jusdino

.github/pull_request_template.md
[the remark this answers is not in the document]
jusdino: Part of my current ticket (#227) includes creating a security profile that is friendlier to automated security scanning, so hopefully that covers your needs 🤞 .

.github/pull_request_template.md
jusdino: I had you set up as an Occupational Therapy user but the UI feature to support multiple compacts is actually still in the PR you're reviewing from John. I just went in and added you to the Audiology/Speech Language Pathology compact as well so you can look at those records in the meantime. Does that resolve the issue for you?

.github/pull_request_template.md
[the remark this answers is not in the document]
jlkravitz: that's great, thanks!

.github/pull_request_template.md
jlkravitz: Hmm, I'm actually testing on my own deployment on which I'm set up as an ASLP user. Same issue on the test site, too, though.

See - https://scribehow.com/shared/Search_for_Licensing_Data_by_Last_Name__T5LwE1-qTBOkN3cj7TRL1w

.github/pull_request_template.md
jsandoval81: I'm actually not sure the best practice here, where a search input is being used. We _can_ mask, but it's not part of the current designs. If you think we should, a new ticket is probably the way to go.

.github/pull_request_template.md
jsandoval81: I think was was the lo-fi design at the time, though those seem to be hidden in Figma now. The hi-fi design calls for a UX that is slightly different, so we'll be touching this again when the hi-fi implementation for this screen happens. If you want to request temp changes to the current lo-fi implementation, we can create a ticket.

.github/pull_request_template.md
jsandoval81: It appears from the designs that an initial search is required. I think this may have stemmed from the lack of paging support and wanting to force users to search first rather than scanning pre-loaded records. It looks like that approach has carried into the hi-fi designs. If you think that's something we should discuss and possibly change, we could create a discussion ticket (or otherwise remember to bring up at our next touch point).

.github/pull_request_template.md
jusdino: Ok, I see that behavior in test as well. I'll have to do more digging. This is definitely not expected behavior and, most perplexing, I have test cases covering this and they do not behave this way.

jusdino: Found the issue. Short version, the code is fine but the feature update didn't get deployed to the CSG `test` environment yet because some merge wonkiness (resulting a missing test dependency in a lambda) caused the deploy to fail. I'll open a bugfix PR to get that dependency added back in.

jusdino: (See #241 )

.github/pull_request_template.md
jlkravitz: No need, especially if we're changing soon. I expect as we move more and more towards the vertical slice model, we'll have fewer of these things come up anyways.

.github/pull_request_template.md
jlkravitz: Got it. In that case, why does the "search" button appear at all while the search modal is up? I can currently click it to close the modal, but it then says "no results" which is confusing. For the sake of this sprint's PR, can we do something simple like disable the "search" button when the search modal is active? That'll make it more intuitive and accessible for screen readers (and, I hope, is quick to implement.)

jsandoval81: @jlkravitz I can start working on something that addresses that, and is also maybe is in the direction of the upcoming hi-fi. I've created #242 

Which branch should I cut that from and merge to? There is already some significant work on `development` since then. For branch history sake it will probably be simplest to treat this as new work in the current sprint. Thoughts?

jlkravitz:
@isabeleliassen Unless you have objections, since we can't merge this in until several upstream changes in `development` are merged, I'm proposing we skip the `development` -> `main` merge for Sprint 9. My thinking is that (1) I know how I'll avoid this issue moving forward, (2) the `main` branch is not actually used by users right now, and (3) several Sprint 10 changes have already been merged into development that would make the review repetitive (vs. just waiting until the end of this sprint).

Let me know what you think!

jsandoval81:
PR #244 Is open and based off latest `development`.

isabeleliassen:
@jlkravitz we are good to skip this merge.

PR #274, opened 2024-10-23
Bug/staff user query error

### LINKED ISSUE #268: Bug: query error updating staff user permissions

jusdino:
We identified a bug in the PATCH staff user method, where the updates fail if both `true` and `false` are provided in the same `actions` block in the request body. This update addresses that bug.

Closes #268

### DIFF
--- a/backend/compact-connect/lambdas/staff-users/data_model/client.py
+++ b/backend/compact-connect/lambdas/staff-users/data_model/client.py
@@ -108,6 +108,7 @@ class UserClient:
         """
         logger.info('Updating staff user permissions', user_id=user_id)
 
+        # Creating a mutable collection so the handlers can add their collected changes
         update_expression_parts = {'add': [], 'delete': []}
 
         # DynamoDB does not support both ADD and DELETE on the same String Set in a single UpdateItem call, so we will
--- a/backend/compact-connect/lambdas/staff-users/data_model/client.py
+++ b/backend/compact-connect/lambdas/staff-users/data_model/client.py
@@ -109,20 +109,50 @@ class UserClient:
         logger.info('Updating staff user permissions', user_id=user_id)
 
         update_expression_parts = {'add': [], 'delete': []}
+
+        # DynamoDB does not support both ADD and DELETE on the same String Set in a single UpdateItem call, so we will
+        # split additions and removals into two calls to prevent a conflict.
+        resp = self._handle_user_permission_additions(
+            user_id=user_id,
+            compact=compact,
+            update_expression_parts=update_expression_parts['add'],
+            compact_action_additions=compact_action_additions,
+            jurisdiction_action_additions=jurisdiction_action_additions,
+        )
+        resp = (
+            self._handle_user_permission_removals(
+                user_id=user_id,
+                compact=compact,
+                update_expression_parts=update_expression_parts['delete'],
+                compact_action_removals=compact_action_removals,
+                jurisdiction_action_removals=jurisdiction_action_removals,
+            )
+            or resp
+        )
+
+        if not (update_expression_parts['add'] or update_expression_parts['delete']):
+            logger.warning('No changes provided for update_user_permissions')
+            raise CCInvalidRequestException('No changes requested')
+
+        return self.schema.load(resp['Attributes'])
+
+    def _handle_user_permission_additions(
+        self,
+        *,
+        user_id: str,
+        compact: str,
+        update_expression_parts: list,
+        compact_action_additions: set,
+        jurisdiction_action_additions: dict,
+    ) -> dict | None:
         expression_attribute_names = {}
         expression_attribute_values = {}
 
         if compact_action_additions:
-            update_expression_parts['add'].append('#permissions.#actions :addActions')
+            update_expression_parts.append('#permissions.#actions :addActions')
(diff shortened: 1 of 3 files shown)


backend/compact-connect/lambdas/staff-users/data_model/client.py
```diff
@@ -109,61 +109,116 @@ def update_user_permissions(
...
+        resp = (
+            self._handle_user_permission_removals(
+                user_id=user_id,
+                compact=compact,
+                update_expression_parts=update_expression_parts['delete'],
+                compact_action_removals=compact_action_removals,
+                jurisdiction_action_removals=jurisdiction_action_removals,
+            )
+            or resp
+        )
```
landonshumway-ia, line 131: Nit: Can we assign these responses to two different variables, and then merge the result together? It would make the code easier to grok.

landonshumway-ia (approved):
Looks good. Just one nit request but doesn't need to hold up this PR IMO

backend/compact-connect/lambdas/staff-users/data_model/client.py
jusdino: I thought about that, but I don't actually want to merge the results. Each includes a complete copy of the user record and only the last response is valid, so we're just selecting the most recent response to return. Can you think of a more obvious way to present that? I'd be happy to update.

backend/compact-connect/lambdas/staff-users/data_model/client.py
```diff
@@ -109,61 +109,116 @@ def update_user_permissions(
         logger.info('Updating staff user permissions', user_id=user_id)
 
         update_expression_parts = {'add': [], 'delete': []}
+
+        # DynamoDB does not support both ADD and DELETE on the same String Set in a single UpdateItem call, so we will
+        # split additions and removals into two calls to prevent a conflict.
```
jlkravitz, line 114: idea being that a delete will undo an add if both are in the string?

backend/compact-connect/lambdas/staff-users/data_model/client.py
```diff
@@ -109,61 +109,116 @@ def update_user_permissions(
...
+                user_id=user_id,
+                compact=compact,
+                update_expression_parts=update_expression_parts['delete'],
+                compact_action_removals=compact_action_removals,
+                jurisdiction_action_removals=jurisdiction_action_removals,
+            )
+            or resp
+        )
+
+        if not (update_expression_parts['add'] or update_expression_parts['delete']):
```
jlkravitz, line 133: why can't this be checked above, before the calls to `_handle_user_permission...`?

backend/compact-connect/lambdas/staff-users/tests/function/test_data_model/test_client.py
```diff
@@ -194,6 +195,29 @@ def test_update_user_permissions_compact_to_board_admin(self):
         # Checking that we're getting the whole object, not just changes
         self.assertFalse({'type', 'userId', 'compact', 'attributes', 'permissions', 'dateOfUpdate'} - resp.keys())
 
+    def test_update_user_permissions_no_change(self):
```
jlkravitz, line 198: To clarify- Is this testing something different from the original issue?

backend/compact-connect/lambdas/staff-users/tests/function/test_handlers/test_patch_user.py
```diff
@@ -40,6 +40,71 @@ def test_patch_user(self):
...
+        event['requestContext']['authorizer']['claims']['scope'] = (
+            'openid email octp/admin octp/octp.admin octp/oh.admin'
+        )
+        event['pathParameters'] = {'compact': 'octp', 'userId': '648864e8-10f1-702f-e666-2e0ff3482502'}
+        event['body'] = json.dumps(
+            {
+                'permissions': {
+                    'octp': {
+                        'actions': {'read': True, 'admin': False},
+                        'jurisdictions': {'oh': {'actions': {'write': True, 'admin': False}}},
```
jlkravitz, line 75: To confirm, this line is the core of what's being tested?

backend/compact-connect/lambdas/staff-users/data_model/client.py
jusdino: Strictly speaking, it's not possible for a request to include both the addition and removal because of the structure of the API but, yes, I have the delete after add just so that, if there is ever a conflict, the DELETE will 'win'.

backend/compact-connect/lambdas/staff-users/data_model/client.py
jusdino: The logic to compile the additions/removals collections into update expression parts is in their respective `_handle_user_permission...` methods, so the simplest check for whether no changes were requested is by looking at the two resulting expression parts, after those handlers are executed.

backend/compact-connect/lambdas/staff-users/tests/function/test_data_model/test_client.py
jusdino: Correct. I was just reviewing the overall test coverage in this area and thought we would benefit from including this as well.

jlkravitz (approved):
@isabeleliassen good to merge

isabeleliassen (approved):

PR #329, opened 2024-11-12
Feat/privilege selection tweaks

### LINKED ISSUE #297: Existing privilege handling changes
Double check logic for purchase privilege button to not allow privilege purchase button if no active license.
Gray out privilege purchase choice if existing privilege for that state has same expiration date as active license.
Also, adjust licensee dahsboard to only show most recent privilege and license for each state

Three bug fix additions to this ticket related to the Select Privilege screen: 
1) Clicking state privilege choice check box exactly does not trigger checked state
2) Jurisprudence modal confirmation can be worked around by using keyboard navigation because the "next" button is enabled while the modal is open
3) The jurisprudence validation was tied directly to the clicking of the state option so clicking the exout did not reset properly

ChiefStief:
### Requirements List
- A user with one privilege that is not eligible for renewal

### Description List
- Upaded models and tests to set number params as number only rather than number or null as there seems to be a ts compiler bug involving values set as null initially (
- refactored privilege select screen to have more calculation inside the ts layer and outside the template layer
- Fixed all the bugs listed in the ticket
- Changed the grayed out behavior to match ticket description

### Testing List
- Code review
- Check out the Privilege selection screen in both the real and mockAPI and confirm that it has been updated to match the changes and fixes described in the ticket

Closes #297 

Notes: The second active license is not reflected in what is grayed out in my display. However, this is an impossible edge case in the final version the user should not be able to get to this screen if they are in such an error state

### DIFF
--- a/webroot/src/pages/SelectPrivileges/SelectPrivileges.ts
+++ b/webroot/src/pages/SelectPrivileges/SelectPrivileges.ts
@@ -311,6 +311,7 @@ export default class SelectPrivileges extends mixins(MixinForm) {
         if (newValue === true) {
             if (stateAbbrev) {
                 this.jurisprudencePendingConfirmation = stateAbbrev;
+                this.formData.jurisprudenceConfirmations[stateAbbrev].value = false;
             }
         }
     }
--- a/webroot/src/pages/SelectPrivileges/SelectPrivileges.ts
+++ b/webroot/src/pages/SelectPrivileges/SelectPrivileges.ts
@@ -63,6 +63,10 @@ export default class SelectPrivileges extends mixins(MixinForm) {
         return this.currentCompact?.compactCommissionFee || null;
     }
 
+    get currentCompactCommissionFeeDisplay(): string {
+        return this.currentCompactCommissionFee?.toFixed(2) || '0.00';
+    }
+
     get userStore(): any {
         return this.$store.state.user;
     }
@@ -99,6 +103,24 @@ export default class SelectPrivileges extends mixins(MixinForm) {
 
         licenseList.forEach((license) => {
             if (license?.issueState?.abbrev) {
+                stateList.push(license?.issueState?.abbrev);
+            }
+        });
+
+        return stateList;
+    }
+
+    get disabledPrivilegeStateChoices(): Array<string> {
+        const licenseList = this.licenseList.concat(this.privilegeList);
+
+        const stateList: Array<string> = [];
+
+        licenseList.forEach((license) => {
+            if (
+                license?.issueState?.abbrev
+                && license?.expireDate === this.activeLicense?.expireDate
+                && license.statusState === LicenseStatus.ACTIVE
+            ) {
                 stateList.push(license?.issueState?.abbrev);
             }
         });
@@ -157,6 +179,27 @@ export default class SelectPrivileges extends mixins(MixinForm) {
             }
 
             return includes;
+        });
+    }
+
+    get selectedStatePurchaseDataDisplayList(): Array<object> {
+        return this.selectedStatePurchaseDataList.map((option) => {
+            let feeDisplay = '';
+            let militaryDiscountAmountDisplay = '';
+
+            if (option?.fee) {
+                feeDisplay = option?.fee?.toFixed(2);
+            }
+
+            if (option?.militaryDiscountAmount) {
+                militaryDiscountAmountDisplay = option?.militaryDiscountAmount?.toFixed(2);
+            }
+
+            return {
+                ...option,
+                feeDisplay,
+                militaryDiscountAmountDisplay
+            };
         });
     }
 
@@ -228,12 +271,14 @@ export default class SelectPrivileges extends mixins(MixinForm) {
         return date;
     }
 
-    get subTotalList(): Array<number> {
+    get subTotalListDisplay(): Array<string> {
         return this.selectedStatePurchaseDataList?.map((purchaseInfo) => {
             const militaryDiscount = purchaseInfo.isMilitaryDiscountActive && purchaseInfo.militaryDiscountAmount
                 ? purchaseInfo.militaryDiscountAmount : 0;
 
-            return ((purchaseInfo?.fee || 0) + (this.currentCompactCommissionFee || 0) - (militaryDiscount));
+            const total = ((purchaseInfo?.fee || 0) + (this.currentCompactCommissionFee || 0) - (militaryDiscount));
+
+            return total.toFixed(2);
         });
     }
 
@@ -382,7 +427,7 @@ export default class SelectPrivileges extends mixins(MixinForm) {
     }
 
     checkIfStateSelectIsDisabled(state): boolean {
-        return this.alreadyObtainedPrivilegeStates.includes(state.id);
+        return this.disabledPrivilegeStateChoices.includes(state.id);
     }
 
     @Watch('alreadyObtainedPrivilegeStates.length') reInitForm() {
--- a/webroot/src/pages/SelectPrivileges/SelectPrivileges.ts
+++ b/webroot/src/pages/SelectPrivileges/SelectPrivileges.ts
@@ -334,7 +334,10 @@ export default class SelectPrivileges extends mixins(MixinForm) {
     }
 
     deselectState(state) {
-        this.formData.stateCheckList.find((checkBox) => (checkBox.id === state?.jurisdiction?.abbrev)).value = false;
+        const stateAbbrev = state?.jurisdiction?.abbrev;
+
+        this.formData.stateCheckList.find((checkBox) => (checkBox.id === stateAbbrev)).value = false;
+        delete this.formData.jurisprudenceConfirmations[stateAbbrev];
     }
 
     submitUnderstanding() {
@@ -357,7 +360,7 @@ export default class SelectPrivileges extends mixins(MixinForm) {
         }
     }
 
-    checkState(stateFormInput) {
+    toggleStateSelected(stateFormInput) {
         const newStateFormInputValue = !stateFormInput.value;
         const stateAbbrev = stateFormInput.id;
 
--- a/webroot/src/models/PrivilegePurchaseOption/PrivilegePurchaseOption.model.spec.ts
+++ b/webroot/src/models/PrivilegePurchaseOption/PrivilegePurchaseOption.model.spec.ts
@@ -36,10 +36,10 @@ describe('PrivilegePurchaseOption model', () => {
         expect(privilegePurchaseOption.jurisdiction.id).to.equal(null);
         expect(privilegePurchaseOption.jurisdiction.abbrev).to.equal(null);
         expect(privilegePurchaseOption.compactType).to.equal(null);
-        expect(privilegePurchaseOption.fee).to.equal(null);
+        expect(privilegePurchaseOption.fee).to.equal(0);
         expect(privilegePurchaseOption.isMilitaryDiscountActive).to.equal(false);
         expect(privilegePurchaseOption.militaryDiscountType).to.equal(null);
-        expect(privilegePurchaseOption.militaryDiscountAmount).to.equal(null);
+        expect(privilegePurchaseOption.militaryDiscountAmount).to.equal(0);
         expect(privilegePurchaseOption.isJurisprudenceRequired).to.equal(false);
     });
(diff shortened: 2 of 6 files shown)


webroot/src/components/CollapseCaretButton/CollapseCaretButton.vue
```diff
@@ -12,8 +12,19 @@
         tabIndex="0"
         class="collapse-caret-button"
     >
-        <img v-if="isUp" class="caret-img" src="@assets/icons/up-caret.svg" alt="Collapse Button Logo" />
-        <img v-else class="caret-img" src="@assets/icons/down-caret.svg" alt="Expand Button Logo" />
+        <svg
```
jsandoval81, line 15: Not in this file, but rather in the TS file there is a missing `emits` declaration in `@Component`; resulting in the following console warning:

Screenshot 2024-11-20 at 10 58 05 AM

webroot/src/pages/SelectPrivileges/SelectPrivileges.ts
```diff
@@ -89,6 +83,10 @@ export default class SelectPrivileges extends mixins(MixinForm) {
         return name;
     }
 
+    get isPhone() {
+        return this.$matches.phone.only;
+    }
+
     get privilegeList(): Array<License> {
```
jsandoval81, line 90: Not sure if this is the correct spot, but should be sort the UI list alphabetically?

Screenshot 2024-11-20 at 10 58 55 AM

webroot/src/components/SelectedStatePurchaseInformation/SelectedStatePurchaseInformation.less
```diff
@@ -0,0 +1,111 @@
...
+    padding: 1rem 1.6rem;
+    border-radius: 20px;
+    color: white;
+    background-color: @primaryColor;
+
+    @media @tabletWidth {
+        margin: 1rem 1rem;
+    }
+
+    .jurisprudence-modal {
```
jsandoval81, line 19: I'm seeing the modal be 100% width on mobile. From the designs, it looks like it should be more like ~90% (ish).

Screenshot 2024-11-20 at 10 55 28 AM

webroot/src/components/SelectedStatePurchaseInformation/SelectedStatePurchaseInformation.less
```diff
@@ -0,0 +1,111 @@
...
+
+                    @media @tabletWidth {
+                        margin: 0 -1.2rem 0 0;
+                    }
+                }
+            }
+        }
+
+        &.sub-row {
+            color: rgba(255, 255, 255, 0.5);
```
jsandoval81, line 73: Ideally we stick with LESS variables. An interpolated form of `color: ~"rgba(@{white}, 0.5)";` might work. If not there is always the LESS function `color: fade(@white, 50%);`

High-level, transparent white on our eventual blue doesn't pass WCAG AA:
- https://webaim.org/resources/contrastchecker/?fcolor=FFFFFF80&bcolor=2459A9

75% seems to be passing:
- https://webaim.org/resources/contrastchecker/?fcolor=FFFFFFBF&bcolor=2459A9

jsandoval81 (changes requested):
Those changes for mobile are looking nice! 🚀 

I wasn't able to test on anything but Chromium due to the cognito settings of the dev environment provider pool. After your updates, I'll try to do more cross-browser testing (if that gets resolved).

webroot/src/components/CollapseCaretButton/CollapseCaretButton.vue
```diff
@@ -12,8 +12,19 @@
...
+            v-if="isUp" width="24" height="14" viewBox="0 0 24 14" role="img"
+            fill="none" xmlns="http://www.w3.org/2000/svg">
+            <title>Collapse Button Logo</title>
+            <path d="M2 11.6342L11.9054 1.72852L21.8111 11.6342"
+            stroke-width="3" stroke-linecap="round" stroke-linejoin="round"/>
+        </svg>
+        <svg v-else width="24" height="14"
```
jsandoval81, line 22: One more thing: If you remove the hardcoded `width` & `height` attributes here and instead set `width` & `height` in your call-site CSS, then future components can also use this icon with their own width / height as needed.

It's really the `viewBox` that's the key from the SVG side. 

You can also remove the `xmlns` attribute if you want.

webroot/src/components/SelectedStatePurchaseInformation/SelectedStatePurchaseInformation.ts
```diff
@@ -0,0 +1,208 @@
...
+        InputButton,
+        Modal,
+        CollapseCaretButton
+    },
+    emits: ['exOutState']
+})
+class SelectedStatePurchaseInformation extends Vue {
+    // PROPS
+    @Prop({ required: true }) selectedStatePurchaseData?: PrivilegePurchaseOption;
+    @Prop({ default: new FormInput({ value: false }) }) jurisprudenceCheckInput?: FormInput | any;
```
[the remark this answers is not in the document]
ChiefStief: Screenshot 2024-11-20 at 11 49 41 AM (2)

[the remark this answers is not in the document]
ChiefStief: And in `SelectPrivileges.ts` I am instantiating the `FormInput` with a boolean value, so it should never get set with string. Also based on the FormInput model, it shouldn't be an issue to set the value as a boolean:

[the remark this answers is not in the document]
jsandoval81: Got it. That's annoying. I don't see an obvious problem there or in the model. Good call stepping past that and not letting TS become a velocity killer there. Maybe just add an end-of-line comment explaining that TS wasn't recognizing the full interface and needed an `any` stop-gap.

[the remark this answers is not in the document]
ChiefStief: When I've researched similar issues before, they actually turned out to be ts bugs, I think.

The new one that I keep running into is the `'Property does not exist on type 'never''` when we have `type: Class | null` and thing?.property

I've worked around that by switching to `type: Class` and instantiating to `thing = new Class({})`

I prefer the old pattern and I feel it makes things more clear. Just bringing that up while we are on the topic of my recent ts struggles / if you have an opinion.

webroot/src/pages/FinalizePrivilegePurchase/FinalizePrivilegePurchase.ts
```diff
@@ -139,11 +139,20 @@ export default class FinalizePrivilegePurchase extends mixins(MixinForm) {
...
+        let states = this.$tm('common.states') as Array<any>;
+
+        if (typeof states[0]?.abbrev === 'function') {
+            const normalize = ([value]) => value;
+
+            states = states.map((st) => ({
+                abbrev: st.abbrev({ normalize }),
+                full: st.full({ normalize }),
+            }));
+        }
```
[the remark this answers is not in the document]
ChiefStief: Oh got it, that's better than what I did!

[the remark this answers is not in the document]
ChiefStief: So this is another example of the ts never bug: 
Screenshot 2024-11-20 at 5 00 11 PM (2)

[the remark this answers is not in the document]
ChiefStief: This single ts bug has been a thorn in my side many times, I would like to maybe add add a global ignore or rule, I have refactored a fair amount to avoid it. How do you feel about that?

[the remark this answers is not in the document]
ChiefStief: I guess we do this: https://stackoverflow.com/questions/75677237/silence-typescript-error-property-does-not-exist-on-type-never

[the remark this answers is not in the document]
ChiefStief: Actually, back to the actual code block we are talking about: that code doesn't seem to work because before normalization the abbrev is not accessible so we cant do: `const model = new State({ abbrev: state.abbrev });`

webroot/src/pages/SelectPrivileges/SelectPrivileges.ts
```diff
@@ -52,7 +50,26 @@ export default class SelectPrivileges extends mixins(MixinForm) {
     // Computed
     //
     get purchaseDataList(): Array<PrivilegePurchaseOption> {
-        return this.currentCompact?.privilegePurchaseOptions || [];
+        const privilegePurchaseOptions = [...this.currentCompact?.privilegePurchaseOptions || []];
+
+        privilegePurchaseOptions.sort((a, b) => {
+            let toReturn = 0;
+            const nameA = (a as PrivilegePurchaseOption).jurisdiction?.abbrev;
+            const nameB = (b as PrivilegePurchaseOption).jurisdiction?.abbrev;
```
jsandoval81, line 58: Would it make more sense to type the function arguments, rather than use the inline casting?

Also, it might make more sense to sort on what the user sees (state name) rather than the abbreviation. When doing that, it's usually best to also compare lower-cased versions of each string; that way if a state name is ever incorrectly updated to "arkansas" it won't get pushed to the top or bottom of the list.

webroot/src/pages/SelectPrivileges/SelectPrivileges.ts
```diff
@@ -110,6 +127,24 @@ export default class SelectPrivileges extends mixins(MixinForm) {
         return stateList;
     }
 
+    get disabledPrivilegeStateChoices(): Array<string> {
+        const licenseList = this.licenseList.concat(this.privilegeList);
+
+        const stateList: Array<string> = [];
```
jsandoval81, line 133: Minor: Can remove the empty line between variable declarations.

jsandoval81 (changes requested):
Everything is shaping up solid. I also tested on Safari & mobile Safari in the simulator and things are looking good 👍 

Just a couple minor items below. Might also be good to merge in latest `development`.

webroot/src/pages/SelectPrivileges/SelectPrivileges.ts
ChiefStief: Oh whoops I should be sorting on the name and will do

jsandoval81 (approved):

jsandoval81:

webroot/src/pages/LicenseeDashboard/LicenseeDashboard.ts
```diff
@@ -65,11 +74,51 @@ export default class LicenseeDashboard extends Vue {
     }
 
     get privilegeList(): Array<License> {
-        return this.licensee?.privileges || [];
+        const privilegeList: Array<License> = [];
```
jlkravitz, line 77: Can you add a function-level comment explaining why this extra logic is needed? Why can't we just use `this.licenseePrivileges` directly? Took me a minute to understand that 1. the privilege list can contain multiple privileges of the same state (e.g., if one is expired (?)) and 2. this function returns the most recent privilege of the duplicates.

webroot/src/pages/LicenseeDashboard/LicenseeDashboard.ts
```diff
@@ -65,11 +74,51 @@ export default class LicenseeDashboard extends Vue {
...
+                privilegeList[previousEntryOfStateIndex] = privilege;
+            }
+        });
+
+        return privilegeList;
     }
 
     get licenseList(): Array<License> {
-        return this.licensee?.licenses || [];
+        const licenseList: Array<License> = [];
```
jlkravitz, line 101: Same thing here.

jlkravitz (commented):
Small nit, but otherwise looks great!

webroot/src/pages/LicenseeDashboard/LicenseeDashboard.ts
jlkravitz: Got it- helpful context! I appreciate it.

jlkravitz (approved):

isabeleliassen (approved):

PR #326, opened 2024-11-12
Feat/allow renewal of privileges

landonshumway-ia:
We need to allow licensees to renew privileges if their license expiration date has been updated. We still prevent them
from purchasing the same privilege if their license has not been updated since their last purchase.

As part of this change, we have updated the data model for privilege records to include a `dateOfRenewal` field, similar
to licenses. The sorting key for privilege and license records will now be suffixed with this 'dateOfRenewal' date stamp. So for example, supposing a privilege was purchased on 2024-11-08, this SK:
```
aslp#PROVIDER#privilege/ne
```

Will now be this tracked as this SK:
```
aslp#PROVIDER#privilege/ne#2024-11-08
```

### Requirements List
- No new requirements. The change should be backwards compatible with the existing schema.

### Description List
- Add logic to allow user to renew their existing privileges if their license expiration date has been updated.
- move common test resources into common python folder
- Update privilege and license sorting keys to include 'dateOfRenewal' date stamp suffix.

### Testing List
- `yarn test:unit:all` should run without errors or warnings
- `yarn serve` should run without errors or warnings
- `yarn build` should run without errors or warnings
- For API configuration changes: CDK tests added/updated in `backend/compact-connect/tests/unit/test_api.py`
- Code review

Closes #295

PR #336, opened 2024-11-15
Feat/state status

### LINKED ISSUE #296: licenses and privileges past the expiration dates automatically become inactive
- may only have backend component
- check is once per day

landonshumway-ia:
The system needs to be aware of when a license or privilege is inactive. Rather than depend on scheduled jobs to update the status of these records, the status will be calculated whenever these records are loaded from the DB.

We still need to track the status of the license as determined by the jurisdiction which uploaded the license data into the system. This introduces a new required field for licenses 'jurisdictionStatus' which we will track in the system.

There is a chance that this status may not be updated by a jurisdiction before it expires according to the 'dateOfExpiration' field on the license in our system, so whenever we load the records for a license or privilege, we calculate the status that is returned to the client by checking the expiration against the current moment (we have determined to use the US Eastern Time zone when determining the date. Though management expects this to likely change in the future)

### Requirements List
- This introduces a new required field for licenses 'jurisdictionStatus', which is a backwards incompatible change with the
current records in the database. Given that we are currently only using test records, we can clear the DB and load 

### Description List
- Add schema loading hooks to calculate the status of licenses and privileges whenever we read them from the DB.
- Move common testing resources into the common folder to reduce code duplication.
- Update Python Runtimes to latest
- Add required 'jurisdictionStatus' field to license records, 
- No longer writing license 'status' field to DB, as it will be calculated when read.

### Testing List
- Code review

Closes #296


landonshumway-ia:
This branch is based off of my previous ticket branch #326. Once that PR has been merged in I will rebase this branch to reduce the diff

landonshumway-ia:
Cherry-picked commits from my previous feature branch into #343 to deal with merge conflicts

PR #343, opened 2024-11-19
Feat/Calculate license and privilege status based on expiration

### LINKED ISSUE #296: licenses and privileges past the expiration dates automatically become inactive

landonshumway-ia:
The system needs to be aware of when a license or privilege is inactive. Rather than depend on scheduled jobs to update the status of these records, the status will be calculated whenever these records are loaded from the DB.

We still need to track the status of the license as determined by the jurisdiction which uploaded the license data into the system. This introduces a new required field for licenses 'jurisdictionStatus' which we will track in the system.

There is a chance that this status may not be updated by a jurisdiction before it expires according to the 'dateOfExpiration' field on the license in our system, so whenever we load the records for a license or privilege, we calculate the status that is returned to the client by checking the expiration against the current moment (we have determined to use the US Eastern Time zone when determining the date. Though management expects this to likely change in the future)

### Requirements List
- This introduces a new required field for licenses 'jurisdictionStatus', which is 🚨🚨 **a backwards incompatible change** 🚨🚨  with the
current records in the database. Given that we are currently only using test records, we can clear the DB and load new test data.

### Description List
- Add schema loading hooks to calculate the status of licenses and privileges whenever we read them from the DB.
- Move common testing resources into the common folder to reduce code duplication.
- Update Python Runtimes to latest
- Add required 'jurisdictionStatus' field to license records, 
- No longer writing license 'status' field to DB, as it will be calculated when read.

### Testing List
- Code review

Closes #296


backend/compact-connect/lambdas/common-python/cc_common/data_model/client.py
```diff
@@ -193,7 +193,6 @@ def create_privilege(self, *, compact: str, jurisdiction: str, provider_id: str)
                                     'providerId': provider_id,
                                     'compact': compact,
                                     'jurisdiction': jurisdiction,
-                                    'status': provider_data['status'],
```
jusdino, line 196: These dates 👇  will need to be in Eastern, now, right? Actually, I see a bunch of places that are UTC dates that will need to be updated for consistency.

backend/compact-connect/lambdas/provider-data-v1/tests/function/test_data_model/test_provider_transformations.py
```diff
@@ -5,7 +5,7 @@
 from boto3.dynamodb.conditions import Key
 from moto import mock_aws
 
-from tests.function import TstFunction
+from .. import TstFunction
```
jusdino, line 8: If we switch to relative imports to avoid test package name conflicts, we should probably do it everywhere.

backend/compact-connect/lambdas/common-python/cc_common/config.py
```diff
@@ -115,5 +116,13 @@ def users_table_name(self):
     def fam_giv_index_name(self):
         return os.environ['FAM_GIV_INDEX_NAME']
 
+    @property
+    def expiration_date_resolution_timezone(self):
+        """
+        This is the timezone used to determine the expiration dates of licenses and privileges.
+        We anticipate that this may change in the future, so we have a configuration value for it.
+        """
+        return timezone('US/Eastern')
```
jusdino, line 125: Just to avoid Daylight Savings Time headaches, I'd suggest that we consolidate around UTC-4, rather than actual Eastern time. That puts our system at the latest American-ish time zone, but one that doesn't jump around twice a year.

jusdino (changes requested):
Good stuff! There a few tweaks to make:

backend/compact-connect/lambdas/provider-data-v1/handlers/licenses.py
```diff
@@ -2,13 +2,13 @@
...
 from cc_common.config import config, logger
-from cc_common.data_model.schema.license import LicensePostSchema
+from cc_common.data_model.schema.license import LicenseIngestSchema
 from cc_common.exceptions import CCInternalException, CCInvalidRequestException
 from cc_common.utils import api_handler, authorize_compact_jurisdiction
 from event_batch_writer import EventBatchWriter
 from marshmallow import ValidationError
 
-schema = LicensePostSchema()
+schema = LicenseIngestSchema()
```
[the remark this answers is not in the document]
landonshumway-ia: Yep, that was the bug your functional test caught 😄

backend/compact-connect/lambdas/common-python/cc_common/data_model/client.py
landonshumway-ia: so in the case of creating the privilege record, we are just copying the expiration date value from the license record. I thought we would only be using the UTC-4 timezone for checking if a license has expired.

landonshumway-ia: I've updated all references to UTC in the runtime code to use the config timezone instead.

backend/compact-connect/lambdas/common-python/cc_common/config.py
jusdino: Now that we're using a basic UTC offset, we [don't actually need pytz anymore](https://docs.python.org/3.6/library/datetime.html#datetime.timezone.utcoffset).

jusdino (approved):
Excellent!

landonshumway-ia:
@jlkravitz This is now ready for CSG review. Thanks

backend/compact-connect/lambdas/common-python/cc_common/data_model/client.py
```diff
@@ -293,10 +238,19 @@ def create_provider_privileges(
...
                     batch.put_item(Item=privilege_record)
+
+            # finally we need to update the provider record to include the new privilege jurisdictions
+            # batch writer can't perform updates, so we'll use a transact_write_items call
+            self.config.provider_table.update_item(
+                Key={'pk': f'{compact_name}#PROVIDER#{provider_id}', 'sk': f'{compact_name}#PROVIDER'},
+                UpdateExpression='ADD #privilegeJurisdictions :newJurisdictions',
+                ExpressionAttributeNames={'#privilegeJurisdictions': 'privilegeJurisdictions'},
+                ExpressionAttributeValues={':newJurisdictions': set(jurisdiction_postal_abbreviations)},
+            )
```
jlkravitz, line 249: Why did we not require this before?

backend/compact-connect/lambdas/common-python/cc_common/data_model/schema/provider.py
```diff
@@ -95,3 +105,22 @@ def populate_fam_giv_mid(self, in_data, **kwargs):  # noqa: ARG001 unused-argume
...
+        # along with the expiration date of the license
+        in_data['status'] = (
+            'active'
+            if (
+                in_data['jurisdictionStatus'] == 'active'
+                and date.fromisoformat(in_data['dateOfExpiration'])
+                > datetime.now(tz=config.expiration_date_resolution_timezone).date()
+            )
+            else 'inactive'
+        )
```
jlkravitz, line 124: Any easy way to reduce duplication of `remove_status_field_if_present`, `pre_load_initialization`, and `_calculate_status` across the license, privilege, and provider models?

jlkravitz (commented):
Looks great! Couple small nits/questions.

backend/compact-connect/lambdas/common-python/cc_common/data_model/schema/provider.py
landonshumway-ia: @jlkravitz Great call. I've introduced a common schema which handles this common serialization/deserialization logic and updated the license, privilege, and provider models to inherit it.

isabeleliassen (approved):

PR #368, opened 2024-12-03
Feat/Record licensee id in privilege transactions

### LINKED ISSUE #287: associate provider ID with payment transactions
- currently authorize.net would only capture billing name
- we ultimately need to be able to generate a list of who (name) has gotten a privilege in a state within a certain time period such as a month
- would we want to know their home state?

landonshumway-ia:
For reporting purposes, we must capture the user id of the licensee in the transaction record in authorize.net. Our user ids are generated as UUIDv4 strings, which are 36 characters long and contain dashes. Originally, we had intended to use the customer id field authorize.net provides for storing these. However, that field is limited to 20 alphanumeric characters (no dashes). The only field that will support the length and dashes of our user ids is the order description field. This field is readonly once the transaction is created, allows up to 255 characters, and is not restricted to alphanumeric characters. 

This PR updates the Purchase client to pass in the licensee id into the order description, so it can be referenced by our reporting which will be implemented in a future ticket.

### Requirements List
- no new requirements. Using existing information.

### Description List
- Add required 'licensee_id' field when processing privilege purchases.

### Testing List
- Code review

Closes #287

### DIFF
--- a/backend/compact-connect/lambdas/purchases/purchase_client.py
+++ b/backend/compact-connect/lambdas/purchases/purchase_client.py
@@ -214,8 +214,9 @@ class AuthorizeNetPaymentProcessorClient(PaymentProcessorClient):
         # Create order information
         order = apicontractsv1.orderType()
         # We store the LICENSEE ID in the description field, since the ID is a UUID that is 36 characters long
-        # and this is the only field that can store that length of data
-        order.description = f'LICENSEE#{licensee_id}#'
+        # and this is the only field that can store that length of data. The description field can hold up to 255
+        # characters.
+        order.description = json.dumps({'LICENSEE': licensee_id})
 
         line_items = apicontractsv1.ArrayOfLineItem()
         for jurisdiction in selected_jurisdictions:
--- a/backend/compact-connect/lambdas/purchases/purchase_client.py
+++ b/backend/compact-connect/lambdas/purchases/purchase_client.py
@@ -118,6 +118,11 @@ class AuthorizeNetPaymentProcessorClient(PaymentProcessorClient):
         if hasattr(response, 'transactionResponse') and hasattr(response.transactionResponse, 'errors'):
             error_code = response.transactionResponse.errors.error[0].errorCode
             error_message = response.transactionResponse.errors.error[0].errorText
+            if error_code =='11' or error_message == 'A duplicate transaction has been submitted.':
+                # This occurs if the user submitted duplicate transactions within the duplicate window
+                # we log the warning and return it as a user error
+                logger.warning(logger_message, error_code=error_code, error_message=error_message)
+                raise CCInvalidRequestException("Duplicate transaction detected for previously successful transaction.")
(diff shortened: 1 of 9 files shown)


backend/compact-connect/lambdas/purchases/purchase_client.py
```diff
@@ -208,6 +211,12 @@ def process_charge_on_credit_card_for_privilege_purchase(
         payment = apicontractsv1.paymentType()
         payment.creditCard = credit_card
 
+        # Create order information
+        order = apicontractsv1.orderType()
+        # We store the LICENSEE ID in the description field, since the ID is a UUID that is 36 characters long
+        # and this is the only field that can store that length of data
+        order.description = f'LICENSEE#{licensee_id}#'
```
jlkravitz, line 218: Is this extensible if we decide to add more data to the description field in the future? What are the tradeoffs of doing it this way, vs. as json (e.g.)?

jlkravitz (commented):
One question, but otherwise looks good!

backend/compact-connect/lambdas/purchases/purchase_client.py
landonshumway-ia: That's a great question and I think the main answer is character count. We can only store 255 chars in this field, and to support json we would need double quotes around the keys and string values, which adds an extra 4 chars per field. We would also need to track brackets. This current setup stores 46 chars, which leaves room for 209 more chars, or approximately 4-5 more fields. We don't anticipate needing to store any more fields at this point, since the licensee Id is enough to fetch all the related records for the licensee on our side of the system.

landonshumway-ia: However, doing the math, if we were to convert this to into a json formatted string it would put us at 52 characters. That leaves us with 203 characters, which is arguable still plenty of headroom for a couple more variables if needed. I think I prefer formatting it as a json string, as it makes it more clear that this is a data field used by an automation system. With that in mind, I'm going to update this to be a json formatted string.

landonshumway-ia:
I've rebased the branch with development and updated our API spec to latest, hence the large diff. The actual code changes are still relatively small.

backend/compact-connect/lambdas/purchases/purchase_client.py
landonshumway-ia: @jlkravitz so I just tested making a transaction using a json serialized string, and authorize.net removed the brackets around the object. That was unexpected behavior which is not documented in their API. 
Screenshot 2024-12-05 at 12 29 51 PM

Obviously, this will not work, so I'm going to revert back to using the # delimiter.

jlkravitz: Strange! But sounds good.

jlkravitz (approved):
Looks great! Thanks for adding the comment about our consideration of JSON. Helpful for our future brains.

isabeleliassen (approved):

PR #382, opened 2024-12-12
Feat/read permission updates

### LINKED ISSUE #207: Read Permissions (backend)
- state administrators can view all the data for those that have a license or privilege in their state (this includes a license that is not the "home state" license)
- state administrators can see public information for those that are in their state. This includes license and discipline, DOB, address, and phone/email (SSN the only item considered private and is addressed in a separate ticket)
- compact administrators can view all data for all their profession
- convert read permission from compact level to state level (maybe?)
- scoped down as read all in your state and read none outside of your state
- Realistic test data in preprod/csg test env to ensure limited read permissions functional

landonshumway-ia:
We need to support specific read access for our staff users when searching for providers in their respective 
compacts/jurisdictions. This makes the following updates to our permissions model

- The former '{compact}/read' scope has been replaced with a '{compact}/readGeneral' scope, which is granted implicitly to all staff users in the system, and allows them to call the query and GET provider endpoints to get generally available (aka not private) data from the system.
- A new action has been added at the compact and jurisdiction levels, 'readPrivate' which must be explicitly granted to a user and allows a user to view the PII for a provider record within their respective compact/jurisdiction that they have permissions for.

With the former read permission split out into 'readGeneral' and 'readPrivate', this allows us to support the following requested access patterns:
- state administrators can view all the data for those that have a license or privilege in their state (this includes a license that is not the "home state" license)
- state administrators can see public information for those that are in their state. This includes license and discipline, DOB, address, and phone/email (SSN the only item considered private)
- compact administrators can view all data for all their profession

### Requirements List
- Note that the authorizer scope change will cause all active JWTs at the time of deployment to become invalid. Any users logged into the system at time of deployment will need to log out and log back in again.

### Description List
- see above

### Testing List
- Code review

Closes #207


landonshumway-ia:
Rebased and moved into #387

PR #388, opened 2024-12-14
Feat/auth tweaks

### LINKED ISSUE #321: Fix auth to work as expected
- After research, chosen first stage for this is to chain logout redirect urls to be able to log any user out using just cognito from both pools
- Explore customization options for login page, at least to confirm viability
- Handle log in such that it functions with two user pool set up
- Also implement mockAPI support including setting mock tokens on login (disambiguation link click on mock) and handling logout such that errors aren't thrown in a mock setting

ChiefStief:
### Requirements List
- Python 3.12+
- A working sandbox account, once you have access you can get it set up by:
0) Create AWS CLI profile that has access to your sandbox and login as that
1) Follow the modified readme in compact-connect 
2) Run the create_staff_user.py script
3) Validate emails to your ia account in ses so that you can get your temp passwords for your licensee user.
4) Manually create a licensee user in the relevant user pool and associated db records (licensee and license). I can help with this when needed. 
5) You may need to do your initial deploy twice as there is a race condition that fails the first time but is met the second time

### Description List
- Added mock login support
- Added route guarding based on role
- Modified logout flow to hit both hosted logout pages by hopping to and from logout page that will handle logout cases
- Fixed 401 interceptor handler to actually fully logout

### Testing List
- For API configuration changes: CDK tests back/bin/run_tests.sh
- Code review
- Try logging in as a licensee, staff, both and logout and verify logging out logs out all users 
- Try the mockAPI flow in the same way
- Produce a 401 error by modifying your access or id token (staff and licensee respectively) and refreshing and confirm the 401 causes the user to totally logout
- Try going to routes that should not be available to a logged in user (both staff and licensee) and confirm they are redirected to their part of the site

Closes #321


landonshumway-ia (approved):
Looks great! Just one optional suggestion.

backend/compact-connect/stacks/persistent_stack/staff_users.py
```diff
@@ -68,13 +68,29 @@ def __init__(
...
+        # authenticated actions.
+        if environment_context.get('allow_local_ui', False):
+            local_ui_port = environment_context.get('local_ui_port', '3018')
+            logout_urls.append(f'http://localhost:{local_ui_port}/Login')
+            logout_urls.append(f'http://localhost:{local_ui_port}/Logout')
+        if not logout_urls:
+            raise ValueError(
+                "This app requires a logout url for its logout function. Either provide 'domain_name' or set "
+                "'allow_local_ui' to true in this environment's context."
+            )
```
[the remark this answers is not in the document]
ChiefStief: So Im on the fence about whether to alter to the comment.

I feel that there are potentially pros and cons,  the cons might be me misinterpreting things, let me know. Id imaging I would want to move the callback URL code just above it as well, if I do?

Pros:
Remove repeated code.

Cons:
UserPool construct becomes not a general class construct but more opinionated into how this project is specifically configured. Because these are common_constructs just for this project in this case that may not be a real con.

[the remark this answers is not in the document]
landonshumway-ia: I see your point with the con. Perhaps there is another place we could put the common code for generating this project specific logout urls then? I'll leave it up to your call. If you're good with how it is. You can resolve this comment.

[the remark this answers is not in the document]
ChiefStief: I think I want to leave it for now because it will be easier to change if we want the user pools to not function identically in the future. 

That being said, there is repeated code / logic not just between user pools but between callback urls and logout urls that might be better as a general thing because it seems like we will often generally run into wanting to allow domain name redirection and localhost redirection.

Ill make a very low priority tech ticket to do refactor that + and revisit the repeated code with callbackUrls + logOutUrls by that time if they still havent changed it might make sense to cement / abstract that away.

Does that sound good to you?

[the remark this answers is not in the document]
landonshumway-ia: Yep that sounds reasonable. Thanks.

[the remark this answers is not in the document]
jusdino: The user pool construct is intended to be specialized/opinionated, so I'd be in favor of consolidating.

webroot/src/components/App/App.ts
```diff
@@ -123,6 +123,7 @@ class App extends Vue {
     async setCurrentCompact(): Promise<void> {
         const { authType } = this.globalStore;
         const { currentCompact, model: user } = this.userStore;
+
```
jsandoval81, line 126: Remove empty line, especially since it's the only edit in this file.

webroot/src/pages/Login/Login.ts
```diff
@@ -58,10 +61,44 @@ export default class Login extends Vue {
...
+        this.$router.push({ name: 'Home' });
+    }
+
+    async mockLicenseeLogin(): Promise<void> {
+        const data = {
+            access_token: 'dflkjdfhwhefjhkfdkfshdlkjfs',
+            token_type: 'Bearer',
+            expires_in: '100000000',
+            id_token: 'idToken',
+            refresh_token: 'refreshToken'
```
jsandoval81, line 96: Minor: Would it be better if these token mock values were simpler and also easily confirmed as mock values and also the user type? E.g. `mock_staff_access_token`, `mock_licensee_id_token`, etc?

webroot/src/pages/Logout/Logout.ts
```diff
@@ -91,6 +90,10 @@ export default class Logout extends Vue {
         }
     }
 
+    get isLoggedIn(): boolean {
```
jsandoval81, line 93: Move up with the rest of the computed props.

jsandoval81 (changes requested):
Looking really solid 👍 

I just had a small number of minor code items in the frontend. I'll DM you about a functionality review.

backend/compact-connect/stacks/persistent_stack/staff_users.py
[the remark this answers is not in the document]
ChiefStief: Well it appears Im outvoted. Ill make the callback url and logout url consolidation change now, but not the url generation function I mentioned

landonshumway-ia (approved):
Backend changes look good. Nice work 👍

jsandoval81:
Frontend changes look good. Great job 🚀

jsandoval81:

backend/compact-connect/common_constructs/user_pool.py
```diff
@@ -117,19 +117,53 @@ def __init__(  # pylint: disable=too-many-arguments
...
+        if environment_context.get('allow_local_ui', False):
+            local_ui_port = environment_context.get('local_ui_port', '3018')
+            callback_urls.append(f'http://localhost:{local_ui_port}/auth/callback')
+        if not callback_urls:
+            raise ValueError(
+                "This app requires a callback url for its authentication path. Either provide 'domain_name' or set "
+                "'allow_local_ui' to true in this environment's context."
+            )
+
+        logout_urls = []
```
jlkravitz, line 149: Would `auth_urls` be a more appropriate name since this includes both Login and Logout URLs?

backend/compact-connect/common_constructs/user_pool.py
```diff
@@ -117,19 +117,53 @@ def __init__(  # pylint: disable=too-many-arguments
...
+            local_ui_port = environment_context.get('local_ui_port', '3018')
+            logout_urls.append(f'http://localhost:{local_ui_port}/Login')
+            logout_urls.append(f'http://localhost:{local_ui_port}/Logout')
+        if not logout_urls:
+            raise ValueError(
+                "This app requires a logout url for its logout function. Either provide 'domain_name' or set "
+                "'allow_local_ui' to true in this environment's context."
+            )
+        # Do not allow resource server scopes via the client - they are assigned via token customization
+        # to allow for user attribute-based access
```
jlkravitz, line 165: Trying to parse this comment. What code does it apply to?

backend/compact-connect/common_constructs/user_pool.py
jlkravitz: Hmm, I guess `OAuthSettings` calls it `logout_urls`. I'm confused, then, why we need to include the Login URLs here?

webroot/src/pages/Home/Home.ts
```diff
@@ -41,7 +41,8 @@ class Home extends Vue {
 
         if (currentCompact) {
             const compactType = currentCompact.type;
-            const authType = this.$store.getters['user/highestPermissionAuthType']();
+            const authType = authStorage.getItem(tokens.staff.AUTH_TYPE)
+                || authStorage.getItem(tokens.licensee.AUTH_TYPE);
```
jlkravitz, line 45: Design question: Is there ever a time where `tokens.staff.AUTH_TYPE` is not of type staff, and `tokens.licensee.AUTH_TYPE` is not of type licensee?

webroot/src/router/index.ts
```diff
@@ -46,8 +49,12 @@ router.beforeEach(async (to, from, next) => {
 
         if (!isLoggedIn) {
             next({ name: 'Logout' });
-        } else {
+        } else if ((isLicenseeRoute && isStaffRoute)
+        || (isLicenseeRoute && authStorage.getItem(tokens.licensee.AUTH_TYPE) === AuthTypes.LICENSEE)
+        || (isStaffRoute && authStorage.getItem(tokens.staff.AUTH_TYPE) === AuthTypes.STAFF)) {
```
jlkravitz, line 54: Similar question here. This is slightly confusing to me, because the all-caps `AUTH_TYPE` implies a constant to me.

webroot/src/store/user/user.actions.ts
```diff
@@ -192,6 +196,21 @@ export default {
         dispatch('sorting/resetStoreSorting', null, { root: true });
         dispatch('reset', null, { root: true });
     },
+    clearAllNonAccessTokens: () => {
+        /* istanbul ignore next */
```
jlkravitz, line 200: What does this do? Code coverage related?

webroot/src/store/user/user.spec.ts
```diff
@@ -349,19 +355,9 @@ describe('User Store Actions', async () => {
...
     });
-    it('should successfully clear session stores', () => {
-        const dispatch = sinon.spy();
-
-        actions.clearSessionStores({ dispatch });
-
-        expect(dispatch.callCount).to.equal(5);
-    });
-    it('should successfully use the authType getter for licensee auth', () => {
+    it('should successfully clear update auth tokens by clearing existing ones except for auth token, then storing', () => {
```
jlkravitz, line 358: "should successfully **clear update auth** tokens..."  typo?

jlkravitz: Relatedly, if you're replacing the auth tokens why would you not want to clear the existing ones?

webroot/src/store/user/user.state.ts
```diff
@@ -25,6 +27,8 @@ export interface State {
 export const state: State = {
     model: null,
     isLoggedIn: (!!authStorage.getItem(tokens.staff.AUTH_TOKEN) || !!authStorage.getItem(tokens.licensee.AUTH_TOKEN)),
+    isLoggedInAsLicensee: Boolean(authStorage.getItem(tokens.licensee.AUTH_TYPE) === AuthTypes.LICENSEE),
+    isLoggedInAsStaff: Boolean(authStorage.getItem(tokens.staff.AUTH_TYPE) === AuthTypes.STAFF),
```
jlkravitz, line 31: Similar question here re: `AUTH_TYPE` on the staff and licensee objects.

webroot/src/pages/Logout/Logout.ts
```diff
@@ -33,54 +38,59 @@ export default class Logout extends Vue {
...
+        return this.userStore.isLoggedIn;
     }
 
     //
     // Methods
     //
     async logout(): Promise<void> {
-        const isLoggedInAsLicenseeOnly = this.$store.getters['user/highestPermissionAuthType']() === AuthTypes.LICENSEE;
+        if (this.isLoggedIn) {
+            const isRemoteLoggedInAsLicenseeOnly = !authStorage.getItem(tokens.staff.AUTH_TOKEN);
```
jlkravitz, line 79: What is 'remote' supposed to mean in this context?

webroot/src/pages/Login/Login.ts
```diff
@@ -58,10 +61,44 @@ export default class Login extends Vue {
...
+    }
+
     //
     // Methods
     //
     redirectToHostedLogin(): void {
         window.location.replace(this.hostedLoginUriStaff);
     }
+
+    async mockStaffLogin(): Promise<void> {
```
jlkravitz, line 75: Super excited to have this in place for testing purposes!

jlkravitz (commented):
- [x] review the pull request to get oriented
	- [x] read the description of the pull request, which should summarize the changes made
	- [x] read through every task on the Scrum board that's encompassed by this pull request
- [ ] review all tests
- [ ] run a security audit of dependencies (e.g. `npm audit` and `pip audit`) to ensure that there are no vulnerabilities that will be deployed to production (as opposed to vulnerabilities that only have an impact on the development environment)

backend/compact-connect/common_constructs/user_pool.py
ChiefStief: In the `logout_url` query param we specify where we want to be redirected to after visiting the hosted logout page. But, the `logout_url` we put there must be one of the allowed `logout_urls` listed above or else cognito will throw an error.

We are using the post logout redirect setting to chain logouts and effectively logout from both user pools if needed by visiting both hosted logout urls. 

- We need to be able to logout out of all user pools if the user clicks logout because we have no way of stopping a user from logging in to both user pools simultaneously due the hosted login pages being for separate user pools / not being aware of each other.

- We need to log out of all user pools by chaining redirects because the only way to fully logout of a cognito userpool is by visiting its logout url, so we need to rapidly visit all of them on "logout." Other methods leave the user logged in in various ways. I can elaborate more if desired.

- That all being said it is very unlikely a user will be logged in to both user pools simultaneously as it requires bookmarking one of the hosted login pages / visiting it after already logged in to one user pool, which a user would not naturally do through using the app.

- The weirdness and complexity of all of these flows comes from the potential edge case of a user logging in to both user pools simultaneously combined with cognito's opinionated logout functionality that disables us from merely logging the user out of the other user pool when they login to the new one.

webroot/src/store/user/user.spec.ts
ChiefStief: We retain the access token from both user pools to maintain an indicator that we are still logged in to that user pool remotely. We need this indicator so that the app knows to remotely log the user out of all logged in user pools.

We clear all the other auth tokens (refresh token and auth type most importantly) to make the app function in an expected way (most recent logged in user is logged in, caused by changing auth_type). Clearing the refresh_token will _**most likely**_ revoke access of the initial user pool when the refresh window expires (the definition of logout to cognito). _**Most likely**_ is used above because within the first hour of login a user can revisit the log in page, pass through login using a previously set http only cookie without providing credentials so you aren't fully logged out.

As stated above, the only way to remove this pass through http only cookie and revoke the refresh is by visiting the logout url

webroot/src/pages/Logout/Logout.ts
ChiefStief: Logged out from the perspective of cognito, not just locally. Local logout would be deleting the tokens. Remote logout would be making the tokens invalid / revoking them + removing http only cookies.

webroot/src/pages/Home/Home.ts
ChiefStief: Actually yes, the way `auth_storage` is evaluated `tokens.staff.AUTH_TYPE` and `tokens.licensee.AUTH_TYPE` are actually stored in exactly the same place as each other in session / local storage. 

So if you set authStorage.setItem(tokens.licensee.AUTH_TYPE,  'licensee');  and then authStorage.getItem(tokens.staff.AUTH_TYPE) would equal 'licensee' because you are actually retrieving the same value you just set.)

backend/compact-connect/common_constructs/user_pool.py
jlkravitz: Got it. This is unintuitive enough that I think it might warrant documenting in a README. I think it might also make sense to document something on your other comment re: being logged in remotely vs locally: "Logged out from the perspective of cognito, not just locally. Local logout would be deleting the tokens. Remote logout would be making the tokens invalid / revoking them + removing http only cookies."

Thoughts?

webroot/src/pages/Home/Home.ts
jlkravitz: Hmm, got it. So then why do you have to call `getItem` on both variables here, given they refer to the same thing?

backend/compact-connect/common_constructs/user_pool.py
ChiefStief: Ok, Ill create a readme documenting the nuances of cognito, how our two user pool set up meshes with those nuances and explaining our mitigation efforts.

webroot/src/pages/Home/Home.ts
ChiefStief: I could do that get once, save it to a variable and then refer to that variable. I could add a comment too explaining that they are the same value.

It might also make more sense to move the auth_type value up out of the staff / licensee objects, so just authStorage.getItem(AUTH_TYPE) as this is more clear that this value is global?

webroot/src/router/index.ts
ChiefStief: `tokens.staff.AUTH_TYPE` and `tokens.licensee.AUTH_TYPE` both evaluate to the constant `auth_type`.

This is a bit obfuscated with the current set up. Would the fix mentioned above "It might also make more sense to move the auth_type value up out of the staff / licensee objects, so just authStorage.getItem('auth_type') as this is more clear that this value is global?" cancel out your concern here?

backend/compact-connect/common_constructs/user_pool.py
jlkravitz: Awesome, thank you!

webroot/src/pages/Home/Home.ts
jlkravitz: Yeah I think that makes more sense from my perspective, and would clarify the variable's role!

webroot/src/router/index.ts
jlkravitz: Yeah I think so - the more we can abstract the complexity here away into a single place, the better. And then, if a comment is still needed, we can add one in one place.

backend/compact-connect/common_constructs/user_pool.py
ChiefStief: I think this comment was erroneously copy pasted

webroot/src/store/user/user.actions.ts
ChiefStief: Yes that makes that line not count in the code coverage tester we are using.

jsandoval81: @jlkravitz For more context, we don't have a full mock replacement of sessionStorage / localStorage in our test runner at the moment. We only have a very small number of difficult-to-test cases for that, so we only need a very small number of coverage ignores.

ChiefStief:
@jlkravitz I added to the existing webroot readme and addressed all your comments, so the PR should be ready for a full re-review.

webroot/README.md
```diff
@@ -227,3 +228,26 @@ Note that testing the **built** app locally will require a running web server; f
...
+    - The downstream effect of this point is that the app needs to handle users being logged in to both user pools in an expected and secure way:
+        - When a user logs out they must be logged out from all user pools they are logged into
+        - When a user logs in to the second user pool, the app must treat them as only logged into the second user pool
+
+- **How the app handles this situation:**
+    - Firstly, the user being logged into both pools is very unlikely as there is not natural way to do this within the app, they would need to manually visit the
+    second hosted login page after logging in normally to the first
+    - When the user logs in as to the second user pool we record that the app should treat them as the second user type and save the initial access token
+        - When the user logs out we check the existence of the access tokens to see which user pools we need to log out, and then chain logout redirects
+        to visist all necessary logout pages
```
jlkravitz, line 252:
```suggestion
        to visit all necessary logout pages
```

webroot/src/store/user/user.actions.ts
jlkravitz: Got it. Thanks for the context!

jlkravitz (approved):
One typo which I'll manually commit! @isabeleliassen This is good to merge.

isabeleliassen (approved):

jlkravitz (approved):

Issue #325, opened 2024-11-12
Read Permissions (frontend)


isabeleliassen:
See #207 
- update user invite and user permission edit UI (permission options increased when inviting user)
- no need to update license search UI

Updates below based on conversation between @jsandoval81 and @jlkravitz https://github.com/csg-org/CompactConnect/pull/401#discussion_r1890621678

- Update API docs to reflect that the top-level actions prop can only conditionally be included if user is compact ED
- Update frontend data layer to prevent top-level actions prop from being included in payload if logged-in user is not a compact ED

Issue #313, opened 2024-11-05
Apply hi-fi styling to frontend


jsandoval81:
- Nav menu UI
- General color scheme
- General fonts
- Page-by-page color & font audit
- a11y testing

PR #401, opened 2024-12-17
Sprint 13

jlkravitz:


.github/workflows/check-lambda-js.yml
```diff
@@ -10,15 +10,16 @@ on:
       - main
       - development
       - ia-web-development
+      - ia-development
```
jlkravitz, line 13: # (Maybe?) Backend Bug

Path: `/octp/Users`
Logged in as user created with: `bin/create_staff_user.py -e joshua+csg-compact-ed@magicclay.io -f kravitz -g joshua -t compact-ed -c octp -j oh`

I added a new user. Then, I updated their permissions. I get back a 502 Bad Gateway response.

Request Body: 
```json
{
    "octp": {
        "actions": {
            "read": false,
            "admin": false
        },
        "jurisdictions": {
            "ar": {
                "actions": {
                    "write": false,
                    "admin": false
                }
            },
            "mt": {
                "actions": {
                    "write": false,
                    "admin": false
                }
            }
        }
    }
}
```

### Error message
Gateway logs–
```(7242f70d-a907-4440-bad3-caa5203d98d7) Lambda execution failed with status 200 due to customer function error: {'permissions': defaultdict(<class 'dict'>, {'octp': {'value': {'actions': ['Missing data for required field.']}}})}. Lambda request id: ea61a756-9304-4a6c-bdad-ef1708f7f21f```

Lambda logs–
```python
[ERROR] ValidationError: {'permissions': defaultdict(<class 'dict'>, {'octp': {'value': {'actions': ['Missing data for required field.']}}})}
Traceback (most recent call last):
  File "/opt/python/aws_lambda_powertools/logging/logger.py", line 450, in decorate
    return lambda_handler(event, context, *args, **kwargs)
  File "/opt/python/cc_common/utils.py", line 73, in caught_handler
    'body': json.dumps(fn(event, context), cls=ResponseEncoder),
  File "/opt/python/aws_lambda_powertools/logging/logger.py", line 450, in decorate
    return lambda_handler(event, context, *args, **kwargs)
  File "/opt/python/cc_common/utils.py", line 149, in authorized
    return fn(event, context)
  File "/var/task/handlers/users.py", line 64, in get_users
    resp['users'] = [user_api_schema.load(user) for user in users]
  File "/opt/python/marshmallow/schema.py", line 722, in load
    return self._do_load(
  File "/opt/python/marshmallow/schema.py", line 909, in _do_load
    raise exc
```

What follows is confusing, because it now seems like this API request fails with a gateway error even on a new user. Almost seems like the database is in a corrupt state.

jlkravitz (commented):
Still going through the checklist, but wanted to surface a couple of issues I came across in the meantime.

jlkravitz:
A few lighthouse performance issues (these have been around for a bit, and I put these down more as a starting point for discussion – not as definite "ToDo" items).

image

.github/workflows/check-lambda-js.yml
```diff
@@ -10,15 +10,16 @@ on:
...
     paths:
-      - backend/compact-connect/lambdas/cloudfront-csp/**
+      - backend/compact-connect/lambdas/nodejs/**
```
jlkravitz, line 16: A few lighthouse performance issues (these have been around for a bit, and I put these down more as a starting point for discussion – not as definite "ToDo" items).

image

.github/workflows/check-lambda-js.yml
```diff
@@ -10,15 +10,16 @@ on:
...
 
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
 
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
```
jlkravitz, line 20: # Accessibility: Keyboard Navigation

When I select "Add state/jurisdiction", a new row appears but no item is selected for input. When I "TAB", it proceeds to the "Cancel" button, instead of the "State/Jurisdiction" dropdown, as I would expect.

image

.github/workflows/check-lambda-js.yml
```diff
@@ -10,15 +10,16 @@ on:
...
 jobs:
```
jlkravitz, line 21: # Accessibility: Keyboard Navigation

Path: `/Account`
I would expect the "Save changes" button to be highlighted after tabbing from the "Last Name" input box. Instead, it highlights a visually external box. I have to tab again to get to the button itself.

image

.github/workflows/check-lambda-js.yml
```diff
@@ -10,15 +10,16 @@ on:
...
-  CheckLambdaJS:
+  CheckLambdas:
```
jlkravitz, line 22: # Password update validation

Path: `/Account`

The validation messages are all 'green', but the box is red. It's not clear why.

image

.github/workflows/check-lambda-js.yml
```diff
@@ -10,15 +10,16 @@ on:
...
     runs-on: ubuntu-latest
```
jlkravitz, line 24: # Accessibility: Keyboard Navigation

When the screen is small enough to collapse the menu into an expandable menu, I can no longer navigate the menu with my keyboard (the menu is skipped when I tab through items). I realize this is technically mobile view, but the menu seems to transition at a width that I might still use on a laptop.

.github/workflows/check-lambda-js.yml
```diff
@@ -36,28 +37,30 @@ jobs:
       - name: Setup Node
```
jlkravitz, line 37: # Accessibility: Keyboard Navigation

On Safari, the menu is not navigable by keyboard at all.

.github/workflows/check-lambda-js.yml
```diff
@@ -36,28 +37,30 @@ jobs:
...
         uses: actions/setup-node@v1
         with:
-          node-version: '20.15.1'
+          node-version: '22.1.0'
```
jlkravitz, line 40: # Backend Error

1. It seems like [`licenses.csv`](https://github.com/csg-org/CompactConnect/blob/0978070151d9947bd27ec20da463fdaf787ab48d/backend/compact-connect/lambdas/python/common/tests/resources/licenses.csv) is not a valid CSV for uploading.
2. Despite that, the `/licenses/bulk-upload` endpoint returns a status 200 response (I think this is expected, since it should only be returning an upload URL), and the upload URL returns a status 204 response. It seems like this might be expected, but I don't see any validation errors in the DataEventTable (which I think I would expect).
3. I see the below log in AWS under the `Sandbox-PersistentStack-BulkUploadsBucketV1ParseOb` lambda.

```json
{
    "level": "INFO",
    "location": "process_bulk_upload_file:135",
    "message": "Invalid license in line 5 uploaded: {'licenseType': ['Must be one of: occupational therapist, occupational therapy assistant.']}",
    "timestamp": "2024-12-18 19:02:59,165+0000",
    "service": "service_undefined",
    "cold_start": true,
    "function_name": "Sandbox-PersistentStack-BulkUploadsBucketV1ParseOb-PdhaLidEQgJO",
    "function_memory_size": "1024",
    "function_arn": "arn:aws:lambda:us-east-1:010526247786:function:Sandbox-PersistentStack-BulkUploadsBucketV1ParseOb-PdhaLidEQgJO",
    "function_request_id": "35e5010e-96f0-43a5-9854-724ae6f9e0f4",
    "valid_data": {
        "licenseType": "speech-language pathologist",
        "status": "active",
        "givenName": "María",
        "middleName": "José",
        "familyName": "Carreño Quiñones",
        "dateOfIssuance": "2024-06-30",
        "dateOfRenewal": "2024-06-30",
        "dateOfExpiration": "2024-06-30"
    },
    "exception": "Traceback (most recent call last):\n  File \"/var/task/handlers/bulk_upload.py\", line 126, in process_bulk_upload_file\n    validated_license = schema.load({'compact': compact, 'jurisdiction': jurisdiction, **raw_license})\n                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n  File \"/opt/python/marshmallow/schema.py\", line 722, in load\n    return self._do_load(\n           ^^^^^^^^^^^^^^\n  File \"/opt/python/marshmallow/schema.py\", line 909, in _do_load\n    raise exc\nmarshmallow.exceptions.ValidationError: {'licenseType': ['Must be one of: occupational therapist, occupational therapy assistant.']}",
    "exception_name": "ValidationError",
    "stack_trace": {
        "type": "ValidationError",
        "value": "{'licenseType': ['Must be one of: occupational therapist, occupational therapy assistant.']}",
        "module": "marshmallow.exceptions",
        "frames": [
            {
                "file": "/var/task/handlers/bulk_upload.py",
                "line": 126,
                "function": "process_bulk_upload_file",
                "statement": "validated_license = schema.load({'compact': compact, 'jurisdiction': jurisdiction, **raw_license})"
            },
            {
                "file": "/opt/python/marshmallow/schema.py",
                "line": 722,
                "function": "load",
                "statement": "return self._do_load("
            },
            {
                "file": "/opt/python/marshmallow/schema.py",
                "line": 909,
                "function": "_do_load",
                "statement": "raise exc"
            }
        ]
    },
    "xray_trace_id": "1-67631c60-4d55644656a1a98f65211ae9"
}
```

.github/workflows/check-lambda-js.yml
jsandoval81: I think this might be the longstanding Safari issue of not allowing tab nav to link / anchor elements by default. Unfortunately this is only something that can be changed in the client-side browser settings. There is technically a app code hack-around, but it negatively impacts screen reader browsers.

https://www.a11yproject.com/posts/macos-browser-keyboard-navigation/

_(Firefox used to have this issue on MacOS too, but I think has since fixed it)_

.github/workflows/check-lambda-js.yml
jsandoval81: I think this might also be the longstanding Safari issue of not allowing full tab nav by default (same as above). Though I think this one only relies on the MacOS System Setting.

.github/workflows/check-lambda-js.yml
```diff
@@ -10,15 +10,16 @@ on:
       - main
```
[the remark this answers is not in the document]
jsandoval81: @jlkravitz Do you mind creating a ticket for this? Speaking w/ @isabeleliassen & @StacyNGriffin this week it sounds like we'll want to have the option to work these at a lower priority than current backlog features.

.github/workflows/check-lambda-js.yml
jsandoval81: Do you mind creating a ticket for this one?

.github/workflows/check-lambda-js.yml
jsandoval81: Do you mind creating a ticket for this one? It might also be good to have the repro steps (unless it basically happens in all cases).

.github/workflows/check-lambda-js.yml
jlkravitz: That sounds good to me! Let's make a ticket for Cloudfront compression and cache policy.

.github/workflows/check-lambda-js.yml
jlkravitz: This was actually on Chrome!

jsandoval81: I see. I think I misunderstood. I thought you were saying you couldn't tab through the nav items. But it sounds like the entire nav menu isn't togglable on mobile w/ keyboard, correct?

If that's the case, that should be solved by the hi-fi design implementation that's currently in progress #313, which includes an overhaul of the nav layout.

.github/workflows/check-lambda-js.yml
jsandoval81: I created #408 

cc @isabeleliassen

.github/workflows/check-lambda-js.yml
jlkravitz: Basically, when I reduce the width on my computer such that the nav collapses, the three horizontal bars – which toggle the navigation items – is itself not toggle-able. If I click the three bars and open the navigation bar with my mouse,  then I can toggle through the items themselves using the keyboard.

image

.github/workflows/check-lambda-js.yml
```diff
@@ -10,15 +10,16 @@ on:
...
       - development
```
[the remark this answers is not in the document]
jlkravitz: The UI _does_ disable that option, but it seems to send to the backend admin-related permissions (even if they do not change the status quo), which the backend rejects. So, at the moment, I do not seem to be able to change any user permissions of any kind as a Board ED.

.github/workflows/check-lambda-js.yml
jsandoval81: Yep, I see now. That will be overhauled in the in-progress #313.

.github/workflows/check-lambda-js.yml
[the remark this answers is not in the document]
jsandoval81: Ah, I think then **1)** the API docs need to be updated, and / or **2)** the backend could be updated to be idempotent, and / or **3)** the frontend can conditionally exclude request props.

Let me know if that tracks with your thinking.

If item 3 is urgent, it can be a new ticket. Otherwise it will be related to upcoming #325 (currently top of the backlog) and a note can be added to that ticket.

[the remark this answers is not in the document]
jlkravitz: Let's add to that ticket. @jusdino informed me that this was an intentional API decision, so we should probably do #1 and #3 on your list.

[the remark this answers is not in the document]
jsandoval81: Gotcha. I can't actually edit #325. Do you mind adding the following to the original ticket notes:

- Update API docs to reflect that the top-level `actions` prop can only conditionally be included if user is compact ED
- Update frontend data layer to prevent top-level `actions` prop from being included in payload if logged-in user is not a compact ED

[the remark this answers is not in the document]
jlkravitz: You got it!

.github/workflows/check-lambda-js.yml
jlkravitz: Ah- didn't know about that script. Helpful!

.github/workflows/check-lambda-js.yml
```diff
@@ -36,28 +37,30 @@ jobs:
...
+          node-version: '22.1.0'
 
       # Use any cached yarn dependencies (saves build time)
       - uses: actions/cache@v2
         with:
           path: '**/node_modules'
           key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}
 
-      # =======================================================================
-      # =                      Cloudfront CSP Lambda                          =
```
jlkravitz, line 48: ## UI Bug

image

Steps to reproduce-

1. Log in as Board ED
2. Go to `/aslp/Licensing`
3. Click "Search" to show all providers.
4. Navigate to another page, e.g., `/Account` by clicking on the "Account" menu item.
5. Navigate back to `/aslp/Licensing` by clicking on the "Licensing" menu item.

Issue #396, opened 2024-12-16
Move SSN to a dedicated, restricted Table (backend)


jusdino:
Remove the SSN/Provider id records from the provider data table and move them to a dedicated, more restricted table, with a dedicated KMS key.

- Add a resource policy statement to the table that denies all actions that allow getting or writing more than one record at a time.
- Add a resource policy statement to the table and key that denies access to all principles except:
- The ingest lambda service role
- The SSN endpoint lambda service role

PR #411, opened 2024-12-18
Feat/ssn table

jusdino:
### Description List
- Added minor improvements to API/Lambda logging to make tracing requests easier
- Moved SSN/provider_id association records into a separate DynamoDB table
- Encrypted that table with a very restrictive key
- Added a resource policy to the table that prevents bulk data read/write
- Integrated required permissions across license ingest and API query workflows

### Testing List
- `bin/run_tests.sh`
- Please manually ETE test by:
  - Deleting all provider data from your sandbox
  - Upload a mock data CSV via the UI, for two different states (requires a user with write permission in two states)
  - Wait ~5 minutes
  - Query for licenses
  - Query for a particular SSN
  - Verify that the provider record has two licenses associated (one for each state)
- Code review

Closes #
#396


backend/compact-connect/common_constructs/queued_lambda_processor.py
```diff
@@ -51,14 +51,45 @@ def __init__(
...
+        #     ),
+        # )
+        #
+        # Except that we are granting the lambda permission to consume SQS messages via resource policy
+        # on the queue, rather than the more conventional approach of principal policy on the IAM role.
+        #
+        # We use a lower-level add_event_source_mapping method here so that we can control how those
+        # permissions are granted. In this case, we need to grant permissions via resource policy on
+        # the Queue rather than principal policy on the role to avoid creating a dependency from the
+        # role on the queue.
```
landonshumway-ia, line 70: This detail is great in explaining how it is working, but I still don't understand why we need to avoid creating that dependency. Does it create a circular dependency between stacks? Could we make that detail a little more clear so devs understand the consequence of not doing it this way?

backend/compact-connect/lambdas/python/common/cc_common/data_model/client.py
```diff
@@ -44,7 +44,7 @@ def get_or_create_provider_id(self, *, compact: str, ssn: str) -> str:
         # This is an 'ask forgiveness' approach to provider id assignment:
         # Try to create a new provider, conditional on it not already existing
         try:
-            self.config.provider_table.put_item(
+            self.config.ssn_table.put_item(
                 Item={
                     'pk': f'{compact}#SSN#{ssn}',
                     'sk': f'{compact}#SSN#{ssn}',
```
landonshumway-ia, line 50: If the pk and sk are the same, what is the purpose of the inverted GSI? Is the GSI going to be needed for future ssn work?

backend/compact-connect/stacks/persistent_stack/ssn_table.py
```diff
@@ -0,0 +1,154 @@
...
+            **kwargs,
+        )
+        self.inverted_index_name = 'inverted'
+
+        self.add_global_secondary_index(
+            index_name=self.inverted_index_name,
+            partition_key=Attribute(name='sk', type=AttributeType.STRING),
+            sort_key=Attribute(name='pk', type=AttributeType.STRING),
+            projection_type=ProjectionType.ALL,
+        )
```
landonshumway-ia, line 84: It would be useful to document here the purpose of this GSI and the access patterns it supports.

landonshumway-ia (commented):
Question about the inverted GSI that we will likely want to document the answer to.

backend/compact-connect/lambdas/python/common/cc_common/data_model/client.py
jusdino: Yeah, I'm anticipating that we will update the key structure and use the inverted key so that we can support both exchanging ssn->provider_id as well as provider_id->ssn

landonshumway-ia (approved):

jlkravitz:
- [ ] review all tests

backend/compact-connect/common_constructs/python_function.py
```diff
@@ -72,19 +75,22 @@ def __init__(
...
-                    'id': 'AwsSolutions-IAM4',
-                    'applies_to': [
-                        'Policy::arn:<AWS::Partition>:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole',
-                    ],
-                    'reason': 'The BasicExecutionRole policy is appropriate for these lambdas',
-                },
-            ],
-        )
+
+        # We only need to suppress this if a role is being created for this lambda
```
jlkravitz, line 79: Do you mean to say, if a role is _not_ being created for this lambda? (I am assuming so since the check is that `role is None`.)

backend/compact-connect/common_constructs/slack_channel_configuration.py
```diff
@@ -37,7 +37,7 @@ def _configure_chatbot_role(self):
             suppressions=[
                 {
                     'id': 'AwsSolutions-IAM4',
-                    'appliesTo': ['Policy::arn:<AWS::Partition>:iam::aws:policy/job-function/ViewOnlyAccess'],
+                    'applies_to': ['Policy::arn:<AWS::Partition>:iam::aws:policy/job-function/ViewOnlyAccess'],
```
jlkravitz, line 40: I assume this was a typo? What were the implications of this typos, if any? (Genuine question because I don't understand AWS well enough to know. I realize this is a suppression, but I imagine an equivalent typo somewhere else could open a security hole.)

backend/compact-connect/stacks/ingest_stack.py
```diff
@@ -29,18 +29,22 @@ def _add_v1_ingest_chain(self, persistent_stack: ps.PersistentStack):
...
             index=os.path.join('handlers', 'ingest.py'),
             handler='ingest_license_message',
+            role=persistent_stack.ssn_table.ingest_role,
             timeout=Duration.minutes(1),
             environment={
                 'EVENT_BUS_NAME': persistent_stack.data_event_bus.event_bus_name,
                 'PROVIDER_TABLE_NAME': persistent_stack.provider_table.table_name,
+                'SSN_TABLE_NAME': persistent_stack.ssn_table.table_name,
+                'SSN_INVERTED_INDEX_NAME': persistent_stack.ssn_table.inverted_index_name,
+                **self.common_env_vars,
```
jlkravitz, line 39: This a dupe of L40 below?

backend/compact-connect/stacks/persistent_stack/ssn_table.py
```diff
@@ -0,0 +1,157 @@
...
+            encryption_key=self.key,
+            billing_mode=BillingMode.PAY_PER_REQUEST,
+            removal_policy=removal_policy,
+            point_in_time_recovery=True,
+            partition_key=Attribute(name='pk', type=AttributeType.STRING),
+            sort_key=Attribute(name='sk', type=AttributeType.STRING),
+            resource_policy=PolicyDocument(
+                statements=[
+                    PolicyStatement(
+                        # No actions that involve reading/writing more than one record at a time. In the event of a
```
jlkravitz, line 53: Thanks for the comments and for thinking so deeply about security on this one!

backend/compact-connect/stacks/persistent_stack/ssn_table.py
```diff
@@ -0,0 +1,157 @@
...
+                ]
+            ),
+            **kwargs,
+        )
+
+        # We create a GSI here in anticipation of a future change, where we will need to facilitate a lookup
+        # of provider_id -> ssn, in addition to our current ssn -> provider_id pattern.
+        self.inverted_index_name = 'inverted'
+        self.add_global_secondary_index(
+            index_name=self.inverted_index_name,
```
jlkravitz, line 82: Dumb question...but where does the 'inverted' part of the key show up in the codebase?

backend/compact-connect/tests/app/base.py
```diff
@@ -128,6 +129,70 @@ def _inspect_persistent_stack(
...
+                'Version': '2012-10-17',
+            },
+        )
+        # Ensure we're using our locked down KMS key for encryption
+        self.assertEqual(
+            ssn_table_template['SSESpecification'],
+            {'KMSMasterKeyId': {'Fn::GetAtt': [ssn_key_logical_id, 'Arn']}, 'SSEEnabled': True, 'SSEType': 'KMS'},
+        )
+        self.compare_snapshot(
+            ssn_table_template['ResourcePolicy']['PolicyDocument'], 'SSN_TABLE_RESOURCE_POLICY', overwrite_snapshot=True
```
jlkravitz, line 194: Should this be `overwrite_snapshot=False`?

backend/compact-connect/tests/app/base.py
```diff
@@ -128,6 +129,70 @@ def _inspect_persistent_stack(
...
+            persistent_stack.ssn_table.api_query_role.node.default_child
+        )
+        ssn_table_template = self.get_resource_properties_by_logical_id(
+            persistent_stack.get_logical_id(persistent_stack.ssn_table.node.default_child),
+            persistent_stack_template.find_resources(CfnTable.CFN_RESOURCE_TYPE_NAME),
+        )
+        ssn_key_template = self.get_resource_properties_by_logical_id(
+            ssn_key_logical_id, persistent_stack_template.find_resources(CfnKey.CFN_RESOURCE_TYPE_NAME)
+        )
+        # This naming convention is important for opting into future organization access logging
```
jlkravitz, line 149: Can you elaborate on this one? Not certain I understand what you mean by "future organization access logging."

jlkravitz (commented):
Looks great! Thanks for being so detail-oriented on this one. Few questions/comments.

backend/compact-connect/common_constructs/python_function.py
```diff
@@ -31,8 +32,9 @@ def __init__(
         construct_id: str,
         *,
         lambda_dir: str,
-        log_retention: RetentionDays = RetentionDays.ONE_MONTH,
+        log_retention: RetentionDays = RetentionDays.INFINITE,
```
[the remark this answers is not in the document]
jusdino: Yeah, intentional. In some semi-related conversations about security/legal concerns, we discussed the possibility that system logs might become evidence in a legal action, which changes the retention strategy to more than one for operational support. I flipped this over with that in mind. We could fine-tune some more later, but this seemed like the more appropriate base pattern at this point.

backend/compact-connect/common_constructs/python_function.py
jusdino: Oh, I can see that this would be confusing. In the case that `role` is `None`, a new role will be created for this particular lambda on construction. If `role` is not `None`, that means that a role is being provided from elsewhere, rather than created, in which case, any Nag suppression would need to be handled wherever the role was created.

backend/compact-connect/common_constructs/slack_channel_configuration.py
jusdino: Yes, this was a typo I noticed while looking things over. The way these suppressions work is to basically 'turn off' any findings from the rule sets we are using that match the criteria of the suppression. In this case the intended suppression was findings:
- for the resource provided in the first argument
- for the rule called [AwsSolutions-IAM4](https://github.com/cdklabs/cdk-nag/blob/main/RULES.md#errors)
- that are specifically referencing this `ViewOnlyAccess` policy in the `applies_to` field.

The `applies_to` field is a further filter criteria that allows us to be very specific in what we are suppressing. By accidentally miss-naming the field, I was effectively making the suppression broader than I intended because that additional filter criteria was missed. Fixing that typo adds the criteria back in the more specific way I had intended.

backend/compact-connect/stacks/ingest_stack.py
jusdino: Huh. Yes it is. Will fix.

backend/compact-connect/stacks/persistent_stack/ssn_table.py
jusdino: It doesn't yet. This index is one we will end up using in the future. I wanted to create it ahead of time so we didn't have to add an index to an existing table later on.

backend/compact-connect/tests/app/base.py
jusdino: Yes it should. Will fix.

backend/compact-connect/tests/app/base.py
jusdino: [This ticket](https://github.com/orgs/csg-org/projects/1/views/7?pane=issue&itemId=91125439&issue=csg-org%7CCompactConnect%7C397) will probably involve creating an [organization trail](https://docs.aws.amazon.com/awscloudtrail/latest/userguide/creating-trail-organization.html) in CloudTrail to create an audit log of data events on this table. We would do an organization trail rather than a cloud trail in the app environment account to enhance security - better to have these audit trails out of reach if an account compromise happens. One way to simplify cross-account dependencies is to use naming conventions, rather than explicit references to particular resources.

With a naming convention established, rather than having to maintain a list of resource identifiers across our test/prod accounts, and bringing them into configuration an the management account, we can just create an organization trail that logs data events for resources that follow a pattern like `ends with '-DataEventsLog'`, which will immediately start logging data events for resources like our SSN table without needing any complicated multi-account deployment strategies.

backend/compact-connect/common_constructs/slack_channel_configuration.py
jlkravitz: Got it! I trust you're right, but just confirming...it looks like the docs actually _do_ use `appliesTo`? https://github.com/cdklabs/cdk-nag

backend/compact-connect/stacks/persistent_stack/ssn_table.py
jlkravitz: Or sorry-- I meant to say, "Where, if anywhere, is the _value_ of the GSI set? i.e., where it sets which field(s) are indexed."  

Maybe the answer is the same, still?

backend/compact-connect/tests/app/base.py
jlkravitz: Got it. That makes sense. Could we add a little more detail to this comment, if only to clarify that "organization access logging" means "AWS account event logging" (or similar)?

backend/compact-connect/common_constructs/python_function.py
jlkravitz: Ah ok. If this understanding is correct, maybe we can update the comment to say "If We only need to suppress this if a role is being created for this lambda

```suggestion
        # Since this lambda won't be assigned a custom role, it will automatically create one with a BasicExecutionRole policy.
```

backend/compact-connect/stacks/persistent_stack/ssn_table.py
jusdino: That's what this function call is doing - declaring a GSI and defining its partition key, sort key, and projected attributes.

backend/compact-connect/common_constructs/slack_channel_configuration.py
jusdino: Yeah, that's the fun of using polyglot libraries (and why I missed the typo in the first place). The library and its docs are written for TypeScript, but, when they get built for their target languages, the interface is updated to match the conventions of the target language (so, `snake_case` for Python).

jusdino:
@jlkravitz, alrighty, this one is back in your court.

backend/compact-connect/stacks/persistent_stack/ssn_table.py
jlkravitz: Ah, I missed the flipped sk/pk fields. Small nit, but can we just add a brief comment saying that to facilitate the reverse lookup, we reverse the primary and sort keys?

jlkravitz (approved):
@isabeleliassen This is good to go.

isabeleliassen (approved):

Issue #412, opened 2024-12-19
Fix backend staff user empty action handling


jusdino:
See [this sprint PR comment](https://github.com/csg-org/CompactConnect/pull/401#discussion_r1890640689) for details. It appears there are some holes in how the API is handling cases around when users are created with no compact actions and/or jurisdictions included with no actions.

PR #413, opened 2024-12-19
Test and fix empty action handling

jusdino:
### Requirements List
- Bug fix for #412 

### Description List
- Revised empty action handling in the staff-user permissions api

Closes #
#412

### DIFF
--- a/backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/user.py
+++ b/backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/user.py
@@ -18,6 +18,19 @@ class CompactPermissionsRecordSchema(Schema):
         allow_none=False,
     )
 
+    @post_dump
+    def drop_empty_actions(self, data, **kwargs):  # noqa: ARG002 unused-kwargs
+        """
+        DynamoDB doesn't like empty sets, so we will make a point to drop an actions field entirely,
+        if it is empty.
+        """
+        if not data.get('actions', {}):
+            data.pop('actions', None)
+        empty_jurisdictions = [jurisdiction for jurisdiction, actions in data['jurisdictions'].items() if not actions]
+        for jurisdiction in empty_jurisdictions:
+            del data['jurisdictions'][jurisdiction]
+        return data
+
 
 class UserAttributesSchema(Schema):
     email = String(required=True, allow_none=False, validate=Length(1, 100))
@@ -63,7 +76,7 @@ class CompactActionPermissionAPISchema(Schema):
     actions = Dict(
         keys=String(),  # Keys are actions
         values=Boolean(),
-        required=True,
+        required=False,
         allow_none=False,
     )
 
@@ -90,7 +103,7 @@ class UserAPISchema(Schema):
     dateOfUpdate = Raw(required=True, allow_none=False)
     attributes = Nested(UserAttributesSchema(), required=True, allow_none=False)
     permissions = Dict(
-        keys=String(validate=OneOf(config.compacts)),
+        keys=String(validate=OneOf(config.compacts)),  # Key is one compact
         values=Nested(CompactPermissionsAPISchema(), required=True, allow_none=False),
         validate=Length(equal=1),
     )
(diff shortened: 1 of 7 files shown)


backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/user.py
```diff
@@ -18,6 +18,19 @@ class CompactPermissionsRecordSchema(Schema):
...
+    def drop_empty_actions(self, data, **kwargs):  # noqa: ARG002 unused-kwargs
+        """
+        DynamoDB doesn't like empty sets, so we will make a point to drop an actions field entirely,
+        if it is empty.
+        """
+        if not data.get('actions', {}):
+            data.pop('actions', None)
+        empty_jurisdictions = [jurisdiction for jurisdiction, actions in data['jurisdictions'].items() if not actions]
+        for jurisdiction in empty_jurisdictions:
+            del data['jurisdictions'][jurisdiction]
```
jlkravitz, line 31: For consistency, can we do this?
```suggestion
           data['jurisdictions'].pop(jurisdiction, None)
```

backend/compact-connect/lambdas/python/staff-users/tests/function/test_handlers/test_patch_user.py
```diff
@@ -105,6 +105,43 @@ def test_patch_user_document_path_overlap(self):
...
+    def test_patch_user_add_to_empty_actions(self):
+        from handlers.users import patch_user, post_user
+
+        with open('tests/resources/api-event.json') as f:
+            event = json.load(f)
+
+        with open('tests/resources/api/user-post.json') as f:
+            api_user = json.load(f)
+        # Create a user with no compact read or admin, no actions in a jurisdiction
+        api_user['permissions'] = {'aslp': {'jurisdictions': {}}}
```
jlkravitz, line 117: Wouldn't we want to instead (or additionally) test a permission structure like the below (all false)?

Perhaps (1) create user with no permissions, (2) add permissions, (3) patch user with something like the below, expecting a user again with no permissions?

(This more closely mirrors my test case, but you understand the codebase here better so I defer to your expertise.)

```json
{
    "permissions": {
        "aslp": {
            "actions": {
                "read": false,
                "admin": false
            },
            "jurisdictions": {
                "ak": {
                    "actions": {
                        "write": false,
                        "admin": false
                    }
                },
                "id": {
                    "actions": {
                        "write": false,
                        "admin": false
                    }
                },
                "ut": {
                    "actions": {
                        "write": false,
                        "admin": false
                    }
                }
            }
        }
    }
}
```

jlkravitz (commented):
One small comment, one question!

backend/compact-connect/lambdas/python/staff-users/tests/function/test_handlers/test_patch_user.py
jusdino: Sure, adding this test case wouldn't hurt. I'll add it in.

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/user.py
jusdino: We could, though they are different because we are expecting different behaviors there:
`data.pop(X, None)` will not produce a `KeyError` in the event that `X` is not in `data`, but `del data[X]` will. I'm effectively using `data.pop(X, None)` as a more concise way of doing:
```python
try:
    del data[X]
except KeyError:
    pass
```

I'd rather not suppress a `KeyError` in the latter call because I expect that the key _should_ be in the `data` object and it would be a bug if I were wrong.

jlkravitz: Ah that makes sense. Thanks for clarifying!

backend/compact-connect/lambdas/python/staff-users/tests/function/test_handlers/test_patch_user.py
jlkravitz: Sounds good. Once that's done, we can get this merged in, followed by #401

jusdino:
@jlkravitz, new test case added ✅

landonshumway-ia (approved):

jlkravitz (approved):
Looks great!

isabeleliassen (approved):

PR #422, opened 2025-01-02
Feat/Add Attestations API Endpoint

### LINKED ISSUE #304: attestations saved (backend)
- backend will store current version of the attestation text and provide via endpoint
- backend will store version of the attestation that was agreed to at runtime for each purchase
- include scope of practice attestation if possible
- include personal information attestations if possible

landonshumway-ia:
We are required to track the historical version of attestations to show what attestation text a user agreed to when they performed an action in the system. This adds the schema design for attestation version records, which we will be storing in the Compact Configuration Table. This adds a new endpoint `GET /v1/compacts/{compact}/attestations/{attestationId}` that retrieves the latest version of an attestation text by its ID.

We need the system to likely support multiple languages, which means attestations need to specify a locale, as legal language is very specific and likely cannot be auto-translated. This adds the locale to the SK and record data to keep locales decoupled from one another.

The Attestation records will be stored according to the following pk and sk pattern:
```
pk: COMPACT#{compact}#ATTESTATIONS
sk: COMPACT#{compact}#LOCALE#{locale}#ATTESTATION#{attestation_type}#VERSION#{version}
```
and will contain the following fields given in the example below:
```
{
  "type": "attestation",
  "attestationId": "jurisprudence-confirmation",
  "displayName": "Jurisprudence Confirmation",
  "description": "For displaying the jurisprudence confirmation",
  "compact": "aslp",
  "version": "1",
  "dateCreated": "2024-06-06T23:59:59+00:00",
  "dateOfUpdate": "2024-06-06T23:59:59+00:00",
  "text": "You attest that you have read and understand the jurisprudence requirements for all states you are purchasing privileges for.",
  "required": true,
  "locale": "en"
}
```

### Requirements List
- None, this is a backwards compatible change

### Description List
- Added `AttestationRecordSchema` for serializing/deserializing attestation records with version tracking
- Added `CompactConfigurationClient` with a method to fetch the latest version of an attestation using DynamoDB's sort key functionality
- Added attestation lambda directory
- Added GET endpoint for getting the latest version of an attestation record.

### Testing List
- Smoke test updated to verify feature against sandbox environment
- Code review

Closes #304


backend/compact-connect/stacks/api_stack/v1_api/api_model.py
```diff
@@ -582,6 +562,25 @@ def post_purchase_privileges_request_model(self) -> Model:
...
                         },
                     ),
+                    'attestations': JsonSchema(
+                        type=JsonSchemaType.ARRAY,
+                        description='List of attestations that the user has agreed to',
+                        items=JsonSchema(
+                            type=JsonSchemaType.OBJECT,
+                            required=['attestationId', 'version'],
+                            properties={
+                                'attestationId': JsonSchema(
```
jusdino, line 572: Let's set a max length here

backend/compact-connect/stacks/transaction_monitoring_stack/transaction_history_processing_workflow.py
```diff
@@ -201,17 +201,20 @@ def __init__(
...
+        #     self,
+        #     f'{compact}-StateMachineExecutionFailedAlarm',
+        #     metric=state_machine.metric_failed(),
+        #     evaluation_periods=1,
+        #     threshold=1,
+        #     actions_enabled=True,
+        #     alarm_description=f'{state_machine.node.path} failed to collect transactions',
+        #     comparison_operator=ComparisonOperator.GREATER_THAN_OR_EQUAL_TO_THRESHOLD,
+        #     treat_missing_data=TreatMissingData.NOT_BREACHING,
+        # ).add_alarm_action(SnsAction(persistent_stack.alarm_topic))
```
jusdino, line 217: Can we leave the alarm and just remove the action? I'd like to still see the CloudWatch alarm if we go looking for it.

jusdino (changes requested):
Looks great! A couple of questions mostly:

backend/compact-connect/lambdas/python/purchases/handlers/privileges.py
```diff
@@ -109,30 +130,90 @@ def _determine_military_affiliation_status(provider_records: list[dict]) -> bool
...
+    or validation rules are not met
+    """
+    # Get all latest attestations for this compact
+    latest_attestations = config.compact_configuration_client.get_attestations_by_locale(compact=compact)
+
+    # TODO: We were asked not to enforce attestations until the frontend is updated # noqa: FIX002
+    #       to pass them in the request body. For now, we will simply check whatever
+    #       attestation is passed in to make sure it is the latest version. This conditional
+    #       should be removed once the frontend is updated to pass in all required attestations.
+    if os.environ.get('ENFORCE_ATTESTATIONS') == 'true':
```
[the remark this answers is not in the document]
landonshumway-ia: It is intended to only be temporary, so I really don't want to add it as a config element. This should be an eye sore for us to rip out as soon as possible.

backend/compact-connect/lambdas/python/purchases/handlers/privileges.py
```diff
@@ -20,6 +21,26 @@
 from cc_common.utils import api_handler
 from purchase_client import PurchaseClient
 
+# List of attestations that are always required
+REQUIRED_ATTESTATION_IDS = [
```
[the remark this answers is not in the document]
landonshumway-ia: Yes, essentially the UI will be well aware of each attestation id value and these id values, as the attestations we are displaying on their respective pages are part of the design. Because of this, the ids will be hardcoded on the UI side for pulling the respective text, and the version number that is returned for each attestation will be the dynamic value which our API will track.

jusdino (approved):
Great! Ready for you, @jlkravitz .

backend/compact-connect/lambdas/python/custom-resources/handlers/compact_config_uploader.py
```diff
@@ -49,6 +54,69 @@ def upload_configuration(properties: dict):
...
 
+def _upload_attestation_configuration(compact_configuration: dict) -> None:
+    """Upload attestation configurations to the provider table.
+    :param compact_configuration: The compact configuration
+    """
+    attestation_record_schema = AttestationRecordSchema()
+    for compact in compact_configuration['compacts']:
+        compact_name = compact['compactName']
+        if 'attestations' not in compact:
+            continue
```
jlkravitz, line 65: Could this ever happen? Is `attestations` not a required field?

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/__init__.py
```diff
@@ -180,4 +180,6 @@ def _generate_providers(self, *, home: str, privilege: str, start_serial: int, n
                 license_expiration_date=date(2050, 6, 6),
                 compact_transaction_id='1234567890',
                 existing_privileges=[],
+                # this is defined in the expected privilege json file
```
jlkravitz, line 183: What does this comment mean? (What is 'this'?)

backend/compact-connect/stacks/api_stack/v1_api/api_model.py
```diff
@@ -582,6 +562,27 @@ def post_purchase_privileges_request_model(self) -> Model:
...
+                                    type=JsonSchemaType.STRING,
+                                    max_length=100,
+                                    description='The ID of the attestation',
+                                ),
+                                'version': JsonSchema(
+                                    type=JsonSchemaType.STRING,
```
jlkravitz, line 578: If an option, why not store the version as an integer rather than as a string?

jlkravitz (commented):
looks good! couple nits

backend/compact-connect/lambdas/python/custom-resources/handlers/compact_config_uploader.py
landonshumway-ia: Good call out, initially I had it as an optional field, but there's no reason it shouldn't be required. All of our current configuration files have it included, so I will remove this check and we'll make it required.

backend/compact-connect/stacks/api_stack/v1_api/api_model.py
landonshumway-ia: We can... it just makes the serialization/deserialization a little more tricky to mess with. DynamoDB stores all numbers as a 'Decimal' type regardless if it is an integer or float, which will cause the marshmallow deserializing to blow up when we load the record from Dynamo, since the Decimal type is not a valid integer type. We can obviously work with this by always casting the version Decimal type to a integer as we pull it out of the table, it just is something that is a little less type safe compared to keeping it as a string.

That being said, if we would like to have it as an integer, it can certainly be done. I prefer to keep it as a string, since we do not intend to perform any math operations on it. It really is just a way for us to sort the sort keys from highest versions to lowest.

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/__init__.py
landonshumway-ia: I've updated the comment to hopefully make it more clear

jlkravitz (commented):

isabeleliassen (approved):

jlkravitz (approved):
@isabeleliassen Good to merge. Sorry about that!

PR #426, opened 2025-01-07
Feat/military affiliation

### LINKED ISSUE #281: military affiliation confirmation (and option to select a license--maybe)

ChiefStief:
### Requirements List
- A working LicenseeUser account

### Description List
- Added in Military Status page
- Added in Update Military Status page
- Added in store + network modules needed to facilitate pages' functionality
- Updated Licensee model and serializer to capture military affiliations api implementation
- Added military affiliations model and serializer
- Edited `FinalizePrivilegePurchase` page and `SelectedStatePurchaseInformation` component to refer to both the state and user's military status to make a military discount determination
- Fixed bug in `InputRadioGroup` component causing multi-line labels to break layout
- Added / updated tests as needed

### Testing List
- Code review
- On both mock and real api and in all screen sizes:
1) click on the view military status button to view your status page
2) Click edit info to go the doc upload page
3) Fill out the form, confirm it does not allow submission with no select choice and no / invalid document selection
4) Submit
5) View your updated military status (real api only)
6) Go through the privilege purchase flow and notice military affiliation discount is being applied
7) Return to the military status screen; try ending your military status
8) Status should now be inactive
9) Go back through privilege purchase flow, confirm military affiliation is no longer being applied

Closes #281 

Notes: 

- The real API system only saves one document upload submission per day. It will replace a prior submission in the day with your most recent one
- The document upload is eventually consistent so your status may not immediately update. I've found it to take ~1 second but theoretically it could be longer.


webroot/src/pages/MilitaryStatus/MilitaryStatus.ts
```diff
@@ -0,0 +1,207 @@
...
+        return this.currentCompact?.type || null;
+    }
+
+    get userStore(): any {
+        return this.$store.state.user;
+    }
+
+    get user(): LicenseeUser {
+        return this.userStore?.model;
+    }
```
jsandoval81, line 103: Minor: The pattern up to now has been to declare computeds like stores, compact, user permissions & "high-level" things near the top of the computed block; as a means of making it easier to find injections / dependencies. Then to have component-generated content further down.

webroot/src/pages/MilitaryStatus/MilitaryStatus.ts
```diff
@@ -0,0 +1,207 @@
...
+        let affiliations: any = [];
+
+        if (this.licensee && this.licensee?.militaryAffiliations) {
+            affiliations = (this.licensee.militaryAffiliations as Array<MilitaryAffiliation>).map((f) => {
+                if (f.fileNames) {
+                    return { name: f.fileNames[0] || null, date: f.dateOfUploadDisplay() };
+                }
+
+                return { name: '' || null, date: '' };
+            });
```
jsandoval81, line 147: Minor: Prefer well-described variable names; e.g. `file` (?) over `f`. Or looking closer, it may be that the array is an array of military affiliations; e.g. `affiliation` or `militaryAffiliations`.

webroot/src/pages/MilitaryStatus/MilitaryStatus.ts
```diff
@@ -0,0 +1,207 @@
...
+    //
+    initFormInputs(): void {
+        const initFormData: any = {
+            submitEnd: new FormInput({
+                isSubmitInput: true,
+                id: 'submit-end',
+            }),
+        };
+
+        this.formData = reactive(initFormData);
```
jsandoval81, line 164: Could this just be: `this.formData = reactive({...});`? This is what we do everywhere else and it would allow us to not need to declare the `any` type.

Also, I noticed we aren't initiating `this.watchFormInputs();` here. Is that on purpose?

webroot/src/pages/MilitaryStatus/MilitaryStatus.ts
```diff
@@ -0,0 +1,207 @@
...
+        await this.$store.dispatch('user/getLicenseeAccountRequest');
+    }
+
+    async sortingChange() {
+        console.log('sort changed');
+    }
+
+    async paginationChange() {
+        console.log('page changed');
+    }
```
jsandoval81, line 197: Should these be removed?

webroot/src/pages/MilitaryStatus/MilitaryStatus.vue
```diff
@@ -0,0 +1,120 @@
...
+                        :listSize="this.affiliations.length"
+                        :sortOptions="sortOptions"
+                        :sortChange="sortingChange"
+                        :pageChange="paginationChange"
+                        :excludeSorting="true"
+                        :excludeTopPagination="true"
+                        :excludeBottomPagination="true"
+                        :isServerPaging="false"
+                        :emptyListMessage="$t('military.noUploadedDocuments')"
+                        :isLoading="$store.state.styleguide.isLoading"
```
jsandoval81, line 48: `isLoading` is probably meant to observe a different store here.

webroot/src/pages/UpdateMilitaryStatus/UpdateMilitaryStatus.less
```diff
@@ -0,0 +1,74 @@
...
+            margin-bottom: 1.4rem;
+            font-size: @h1FontSize;
+        }
+
+        .core-info-block {
+            display: flex;
+            flex-direction: column;
+            padding: 2rem 2rem 0 2rem;
+            border-radius: 14px;
+            background-color: white;
```
jsandoval81, line 34: Minor: Prefer color variables.

webroot/src/pages/UpdateMilitaryStatus/UpdateMilitaryStatus.ts
```diff
@@ -0,0 +1,157 @@
...
+                    maxSizeMbAll: 100,
+                },
+            }),
+            submit: new FormInput({
+                isSubmitInput: true,
+                id: 'submit',
+            }),
+        };
+
+        this.formData = reactive(initFormData);
```
jsandoval81, line 101: Same as earlier comment - existing patterns seem to be just as effective, and not require `any`.

Also, making sure `this.watchFormInputs()` is excluded on purpose.

webroot/src/pages/LicenseeDashboard/LicenseeDashboard.ts
```diff
@@ -172,6 +172,12 @@ export default class LicenseeDashboard extends Vue {
         console.log('Starting Privilege Purchase Flow!');
     }
 
+    viewMilitaryStatus() {
+        this.$router.push({
+            name: 'MilitaryStatus'
+        });
+    }
+
```
jsandoval81, line 180: Does this need the `compact` param?

webroot/src/styles.common/_colors.less
```diff
@@ -58,6 +58,7 @@
 @midOrange: @__orange;
 @darkOrange: darken(@midOrange, 15%);
 @lightOrange: lighten(@midOrange, 40%);
+@errorRed: #E04545;
```
jsandoval81, line 61: Minor: Move this to the `// Reds` section.

Also, is this just a new color in the designs? Or did you have to create it to meet a11y? Ideally, we could use the same red we've been using for errors, rather than adding a new error-red just for military UI.

jsandoval81 (changes requested):
Looking solid 👍 

This was a mostly code review. I'll take a deeper dive in the UI on the next pass.

webroot/src/components/Forms/_mixins/form.mixin.ts
```diff
@@ -53,7 +53,7 @@ class MixinForm extends Vue {
     get joiMessages(): any {
         const messages = {
             string: {
-                'string.empty': this.$t('inputErrors.required'),
```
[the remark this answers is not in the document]
ChiefStief: No that's a total mistake. I didn't mean to delete that, good catch

webroot/src/pages/MilitaryStatus/MilitaryStatus.ts
ChiefStief: Maybe, I was unsure about this.

Essentially, the ListContainer component requires these functions be passed in as props. I didn't want to make those props non-required + change ListContainer logic to allow for this, so I just did this. Its a bit of a punt / Im happy to address that better if you would like.

jsandoval81: If the API supported paging, these would at least trigger a re-fetch of the data. Does it make sense to have them just do that for now? If not, you could just have them `return false` & add comments that the API doesn't support sort or paging.

webroot/src/styles.common/_colors.less
ChiefStief: Was just a new color on the designs.

I've deleted this here and replaced with @midRed in references to this

webroot/src/pages/LicenseeDashboard/LicenseeDashboard.ts
ChiefStief: I just added it, but its been working fine.

I think it may either assume same params if not provided or only edit the url left as far as specified / necessary. I have a vague memory of reading about this.

Either way better to just to explicitly do it correctly.

webroot/src/pages/MilitaryStatus/MilitaryStatus.ts
ChiefStief: That wasn't intentional in either place.

I actually missed this in the SelectPrivileges form too. I'm not sure its important in this spot because there are no fields or success / failure messages, but should probably be used in the other one.

webroot/src/pages/UpdateMilitaryStatus/UpdateMilitaryStatus.ts
ChiefStief: Not intentional, though my form does seem to validate as I go.

I'm going to add this in for the post submit functionality, as it irons out those wrinkles.

webroot/src/pages/MilitaryStatus/MilitaryStatus.ts
```diff
@@ -0,0 +1,212 @@
...
+
+    editInfo() {
+        if (this.currentCompactType) {
+            this.$router.push({
+                name: 'UpdateMilitaryStatus',
+                params: { compact: this.currentCompactType }
+            });
+        }
+    }
+}
```
jsandoval81, line 212: It looks like we'll need to implement focus trapping to fully support keyboard nav. There's a recent example of this in the `SelectedStatePurchaseInformation` component. It's just a single method that is configured with the first & last inputs, and then a `@keydown.tab` binding in the Modal component in the template.

This likely goes for any modals moving forward.

webroot/src/pages/MilitaryStatus/MilitaryStatus.ts
```diff
@@ -0,0 +1,212 @@
...
+                    if (militaryAffiliation.fileNames) {
+                        return {
+                            name: militaryAffiliation.fileNames[0] || null,
+                            date: militaryAffiliation.dateOfUploadDisplay()
+                        };
+                    }
+
+                    return { name: '' || null, date: '' };
+                });
+        }
```
jsandoval81, line 153: We generally try to avoid early returns. Also the default of `return { name: '' || null, date: '' };` it looks like `name` will always fallback to `null`. Overall is there a way to clean up the body of the map function?

E.g.
```
.map((militaryAffiliation: MilitaryAffiliation) => {
    const affiliationDisplay = { name: '', date: '' };

    if (militaryAffiliation.fileNames?.length) {
        affiliationDisplay.name = militaryAffiliation.fileNames[0];
        affiliationDisplay.date = militaryAffiliation.dateOfUploadDisplay();
    }

    return affiliationDisplay;
});
```

Adding the `MilitaryAffiliation` typing this way may also remove the need for the `as` casting.

webroot/src/pages/UpdateMilitaryStatus/UpdateMilitaryStatus.ts
```diff
@@ -0,0 +1,161 @@
...
+    // Methods
+    //
+    initFormInputs(): void {
+        this.formData = reactive({
+            affiliationType: new FormInput({
+                id: 'affiliation-type',
+                name: 'affiliation-type',
+                shouldHideLabel: true,
+                label: computed(() => this.$t('military.affiliationType')),
+                validation: Joi.string().required().messages({ 'string.empty': this.joiMessages.string['string.empty'] }),
```
jsandoval81, line 77: As it stands, the "required" message is kind of awkward:
Screenshot 2025-01-10 at 9 23 11 AM


I think this should just be `.messages({ ...this.joiMessages.string })` like all the rest of our forms. That would result in the more standard message:
Screenshot 2025-01-10 at 9 37 48 AM

jsandoval81 (changes requested):
Functionality looks solid. Just a few trailing items that are pretty minor.

jsandoval81 (approved):

jsandoval81:

webroot/src/models/Licensee/Licensee.model.ts
```diff
@@ -186,6 +187,18 @@ export class Licensee implements InterfaceLicensee {
     public statusDisplay(): string {
         return this.$t(`licensing.statusOptions.${this.status}`);
     }
+
+    public isMilitary(): boolean {
+        // The API does not return the affiliations in the get-all endpoint therefore all users will appear unaffiliated
+        // if only that endpoint has been called
```
jlkravitz, line 193: I'm confused by this comment - is this a comment on the function's logic, or broader conceptual documentation about the API? If the latter, can we hoist it up somewhere more appropriate - perhaps top-level, a README, or some other file?

jlkravitz: OK, I think I have a better understanding from reading further down. Might it make sense to just add a brief comment on this model file explaining that this model technically models two versions of a license - that returned from a get-all response and that returned from a get-one response?

webroot/src/models/MilitaryAffiliation/MilitaryAffiliation.model.ts
```diff
@@ -0,0 +1,70 @@
...
+    public dateOfUpdate? = null;
+    public dateOfUpload? = null;
+    public documentKeys? = null;
+    public fileNames? = null;
+    public status? = null;
+
+    constructor(data?: InterfaceMilitaryAffiliationCreate) {
+        const cleanDataObject = deleteUndefinedProperties(data);
+
+        Object.assign(this, cleanDataObject);
```
jlkravitz, line 40: Can we document why we have to do this? (Is this standard in our codebase? It's possible I've just missed it in the past.)

webroot/src/pages/MilitaryStatus/MilitaryStatus.ts
```diff
@@ -0,0 +1,229 @@
...
+    get sortOptions(): Array<any> {
+        // Sorting not API supported
+        return [];
+    }
+
+    get affiliations(): Array<any> {
+        let affiliations: any = [];
+
+        if (this.licensee && this.licensee?.militaryAffiliations) {
+            affiliations = (this.licensee.militaryAffiliations as Array<MilitaryAffiliation>)
```
jlkravitz, line 142: why is casting required here?

webroot/src/pages/MilitaryStatus/MilitaryStatus.ts
```diff
@@ -0,0 +1,229 @@
...
+                .map((militaryAffiliation: MilitaryAffiliation) => {
+                    const affiliationDisplay = { name: '', date: '' };
+
+                    if (militaryAffiliation.fileNames && (militaryAffiliation.fileNames as Array<string>).length) {
```
jlkravitz, line 146: and here?

jlkravitz (commented):
looks great! a few small nits.

webroot/src/pages/MilitaryStatus/MilitaryStatus.ts
ChiefStief: It turns out casting is not required on this line given other changes. Why it was used initially will be explained in the comment below explaining why that line does need casting.

webroot/src/pages/MilitaryStatus/MilitaryStatus.ts
ChiefStief: Casting is needed here to avoid the error: `Property 'length' does not exist on type 'never'.`

TS has an annoying well documented behavior initially discussed here: https://github.com/microsoft/TypeScript/issues/9998 

Some digging would be required on your end to explain the relevance so I will link this SO post that explains the situation pretty well: https://stackoverflow.com/questions/75677237/silence-typescript-error-property-does-not-exist-on-type-never

Essentially, TS guesses the type. The type was initially null and TS doesn't figure out that it is now an array. Further it can't be null there because of the totality of the conditional: `if (militaryAffiliation.fileNames && (militaryAffiliation.fileNames as Array<string>).length)`

jlkravitz: Got it - thanks for the explanation!

webroot/src/models/MilitaryAffiliation/MilitaryAffiliation.model.ts
jlkravitz: @ChiefStief Flagging this before we merge. If this is something we've been doing across our codebase, we can ignore!

ChiefStief: @jlkravitz Sorry I was just confirming why we do this. This is something we do across the codebase.

This is just a sanity thing so that we can cleanly define what the properties are going to be with the interfaces and don't end up with undefined properties on our objects. Sometimes, undefined properties can cause more complexity so we just get rid of them.

jlkravitz: Got it -all good. Thanks for the explanation!

jlkravitz (approved):

isabeleliassen (approved):

PR #428, opened 2025-01-08
Frontend/permission api updates

### LINKED ISSUE #325: Read Permissions (frontend)

jsandoval81:
### Requirements List
- _None_

### Description List
- Change compact-level `read` permission to `readPrivate`
- Add state-level `readPrivate` permission
- Omit compact permission in server payload if current user isn't compact admin (rather than sending `false`)
- Don't re-send permissions as `false`
- Updated licensee model tests for slightly improved coverage

### Testing List
- Code review
- Make sure to have 3 users created for yourself (this will be helpful with testing many PRs; should be able to use UI if needed):
    - A compact admin
    - A state admin (same compact)
    - A non-state admin (same compact)
- Ensure compact admin can set / update all types of permissions for users
- Ensure state admin can set / update all types of permissions for users in their state(s)
- Ensure non-state admin can't perform user management

Closes #325


ChiefStief (approved):

jsandoval81:

webroot/src/components/Users/UserRowEdit/UserRowEdit.ts
```diff
@@ -458,15 +471,44 @@ class UserRowEdit extends mixins(MixinForm) {
         };
         const stateKeys = Object.keys(formValues).filter((key) => key.startsWith('state-option'));
 
+        // Server endpoints may not be idempotent so the frontend needs to handle statefulness;
+        // e.g. if the user's server permisson is already false, we may get a server error if we try to send false again.
```
jlkravitz, line 475: If I recall from discussions with Justin, I don't think this is really about idempotency. I believe the reason is because board ED users cannot change compact-level permissions, so any value (even if it's the same) is considered a requested change, which is rejected by the API.

@jusdino Is that accurate? Anything else to add?

Assuming this is right, could we update this comment and the code/variables to more accurately reflect this intent?

webroot/src/components/Users/UserRowEdit/UserRowEdit.ts
```diff
@@ -458,15 +471,44 @@ class UserRowEdit extends mixins(MixinForm) {
...
-
-            compactData.states.push({
+            const stateData = {
                 abbrev: stateAbbrev,
                 ...this.setStatePermission(statePermission),
-            });
+            };
+
+            // Server endpoints may not be idempotent so the frontend needs to handle statefulness;
+            // e.g. if the user's server permisson is already false, we may get a server error if we try to send false again.
```
jlkravitz, line 496: Same thing here.

webroot/src/models/StaffUser/StaffUser.model.ts
```diff
@@ -89,9 +90,19 @@ export class StaffUser extends User implements InterfaceStaffUserCreate {
             } else {
                 // Otherwise look for state-level permissions
                 states?.forEach((statePermission) => {
-                    const { isWrite, isAdmin: isStateAdmin } = statePermission;
+                    const {
+                        isReadPrivate: isStateReadPrivate,
+                        isWrite: isStateWrite,
+                        isAdmin: isStateAdmin
```
jlkravitz, line 96: These variable renames are helpful to distinguish state and compact-level permissions!

jlkravitz (commented):
One comment!

webroot/src/components/Users/UserRowEdit/UserRowEdit.ts
jsandoval81: The board ED user situation is handled up near line ~365; to not have board EDs attempt to send those props at all, even if false.

This section here is related to some testing & conversations we've had internally about not sending false if the value is already false (which is a different - but understandably similar-sounding item 🙂). I think the API doc updates for that are planned to be in an upcoming backend PR.

jlkravitz: Ahh, ok. I guess that then begs the question as to why server endpoints may not be idempotent? That seems unnecessarily brittle. I am sure there are reasons beyond my understanding, but is that reasoning something we can document somewhere?

jsandoval81: Yep. I think some updates to the swagger docs are coming soon. They got a little decoupled from the original code updates but are on their way. I think they'll also be accompanied by some testing / confirmation as to the actual behavior. These frontend lines in the meantime felt like an ok stretch to address any gaps.

jlkravitz: Got it!

jlkravitz (approved):

isabeleliassen (approved):

PR #430, opened 2025-01-08
Frontend/Fix pw input validation messages

### LINKED ISSUE #410: Password update validation bug
Path: `/Account`

The validation messages are all 'green', but the box is red. It's not clear why.

To reproduce,

1. Navigate to "Account" page
2. Enter current password (in my case, Josh12345678)
3. Enter new password (in my case, Josh123456789)
4. All errors are green, but the textbox is red and I cannot submit.

Similarly, if I type a new password as "12345678901", all of the validation items are green, even though there is most definitely not a special, lowercase, or uppercase character in the input box.

image

_Originally posted by @jlkravitz in https://github.com/csg-org/CompactConnect/pull/401#discussion_r1890689105_

jsandoval81:
### Requirements List
- _None_

### Description List
- Move message augmentation to after message evaluation
- Update frontend to match latest backend pw requirements (length only)
- Add tests

### Testing List
- Code review
- Fill out the change-password section of the account page to ensure correct validation behavior
    - No need to actually submit the change

Closes #410

### DIFF
--- a/webroot/src/components/ChangePassword/ChangePassword.ts
+++ b/webroot/src/components/ChangePassword/ChangePassword.ts
@@ -102,9 +102,6 @@ class ChangePassword extends mixins(MixinForm) {
                     .min(12)
                     .doesNotInclude([
                         this.user.email,
-                        this.user.firstName,
-                        this.user.lastName,
-                        'password',
                     ])
                     .messages({
                         ...this.joiMessages.string,
--- a/webroot/src/components/ChangePassword/ChangePassword.ts
+++ b/webroot/src/components/ChangePassword/ChangePassword.ts
@@ -100,10 +100,6 @@ class ChangePassword extends mixins(MixinForm) {
                 validation: joiPassword
                     .string()
                     .min(12)
-                    .minOfSpecialCharacters(1)
-                    .minOfLowercase(1)
-                    .minOfUppercase(1)
-                    .minOfNumeric(1)
                     .doesNotInclude([
                         this.user.email,
                         this.user.firstName,
--- a/webroot/src/components/Forms/InputPassword/InputPassword.spec.ts
+++ b/webroot/src/components/Forms/InputPassword/InputPassword.spec.ts
@@ -7,10 +7,15 @@
 
 import { expect } from 'chai';
 import { mountShallow, mountFull } from '@tests/helpers/setup';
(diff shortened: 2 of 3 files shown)


webroot/src/components/ChangePassword/ChangePassword.ts
```diff
@@ -100,10 +100,6 @@ class ChangePassword extends mixins(MixinForm) {
                 validation: joiPassword
                     .string()
                     .min(12)
-                    .minOfSpecialCharacters(1)
-                    .minOfLowercase(1)
-                    .minOfUppercase(1)
-                    .minOfNumeric(1)
                     .doesNotInclude([
                         this.user.email,
```
ChiefStief, line 104: This ticket as described looks good. I noticed that the common string validator is case sensitive, so when I put "dana" in as part of my password it lets me. Not sure if you want to address this here or at all.

ChiefStief: Makes sense, on the MVP point. All that works for me.

ChiefStief (approved):

jsandoval81:

webroot/src/components/ChangePassword/ChangePassword.ts
```diff
@@ -100,15 +100,8 @@ class ChangePassword extends mixins(MixinForm) {
                 validation: joiPassword
                     .string()
                     .min(12)
-                    .minOfSpecialCharacters(1)
```
jlkravitz, line 103: Why are we removing these requirements, yet still testing them below?

(I know we had briefly discussed updated NIST guidelines several weeks ago - related? Generally, I'm keen to remove complexity requirements and keep only length requirements, as I believe NIST recommends.)

jsandoval81: The tests are just for the generic functionality of the low-level InputPassword component, so we know we can use them later if needed.

You're correct that the app implementation in this ChangePassword component has also been updated to use the latest NIST recommendation. That was the decision on the last standup call - I went ahead and added that comment to the ticket just now.

jlkravitz: Got it- thank you for the context!

jlkravitz (approved):
@isabeleliassen This is good to go.

Issue #186, opened 2024-09-04
Add event records for license renewals, deactivations


jusdino:
Insurers and others will want to know, for a given day in the past, whether an individual was licensed to practice in a given jurisdiction. To do that we will need to present historical information on an individuals license status over time in the detailed view. We need to capture important license events such as the dates of renewals and any deactivation initiated by a board, combined with license the expiry date associated with each event, so we can have a complete day-by-day accounting of their licensure.

PR #434, opened 2025-01-10
Feat/license event records

jusdino:
### Description List
- Added `licenseUpdate` and `privilegeUpdate` record types
- Added change detection and update record creation on license ingest and privilege purchase processes
- Updated provider detail response to present change history in license and privilege objects

Closes #
#186


backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
```diff
@@ -223,54 +225,208 @@ def create_provider_privileges(
...
+                        'Key': {
+                            'pk': {'S': f'{compact_name}#PROVIDER#{provider_id}'},
+                            'sk': {'S': f'{compact_name}#PROVIDER'},
+                        },
+                        'UpdateExpression': 'ADD #privilegeJurisdictions :newJurisdictions',
+                        'ExpressionAttributeNames': {'#privilegeJurisdictions': 'privilegeJurisdictions'},
+                        'ExpressionAttributeValues': {':newJurisdictions': {'SS': jurisdiction_postal_abbreviations}},
+                    }
+                }
+            )
```
landonshumway-ia, line 303: We should leave a comment here explicitly calling out why we append the provider privilegeJurisdictions update last, so no one ever moves it to an earlier step in the algorithm.

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
```diff
@@ -223,54 +225,208 @@ def create_provider_privileges(
...
+                    transactions = transactions[batch_size:]
+                    if transactions:
+                        logger.info(
+                            'Breaking privilege updates into multiple transactions',
+                            provider_id=provider_id,
+                            privlige_jurisdictions=jurisdiction_postal_abbreviations,
+                            compact_transaction_id=compact_transaction_id,
+                        )
+                except ClientError as e:
+                    # Roll back any successful transactions
```
landonshumway-ia, line 331: missing a error log message here that would be useful for context

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
```diff
@@ -223,54 +225,208 @@ def create_provider_privileges(
...
+                                'Put': {
+                                    'TableName': self.config.provider_table_name,
+                                    'Item': TypeSerializer().serialize(
+                                        privilege_record_schema.dump(original_privilege)
+                                    )['M'],
+                                }
+                            }
+                        )
+                    else:
+                        # If it was a new creation, delete it
```
landonshumway-ia, line 397: it may be helpful to have a log message here noting that we are deleting something, so we at least have some paper trail of what record would have shown up in the system

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
```diff
@@ -223,54 +225,208 @@ def create_provider_privileges(
...
+                                    'sk': {'S': item['sk']},
+                                },
+                            }
+                        }
+                    )
+                elif item.get('type') == 'privilege':
+                    # For privilege records, check if it was an update or new creation
+                    original_privilege = existing_privileges_by_jurisdiction.get(item['jurisdiction'])
+                    if original_privilege:
+                        # If it was an update, restore the original record
```
landonshumway-ia, line 385: We should also have a log statement showing that we are restoring the record and what the change would have been. Given that we don't expect this code to get exercised very often, we should be fine to log once per privilege in the list

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/common.py
```diff
@@ -57,3 +61,67 @@ def ensure_value_is_datetime(value: str):
...
+            # "homeAddressStreet": "123"
+            # Could produce the same hash without a separator indicating where key ends and value begins.
+            change_hash.update(cls._b64encode_str(key))
+            # Between keys and values, we'll hash a slash ('/')
+            change_hash.update(b'/')
+            change_hash.update(cls._b64encode_str(value))
+            # After each item, we'll hash a dash ('-')
+            change_hash.update(b'-')
+        # After each dict, we'll hash a hash ('#')
+        change_hash.update(b'#')
```
landonshumway-ia, line 114: This is a pretty critical piece, as whatever pattern we choose we will essentially be locked into forever to determine the update records by hash. I am trying to walk through any drawbacks with this current approach, though honestly nothing comes to mind. I don't foresee the need to grab individual updates for a license or privilege, as the SK is designed to be able to get all records associated with a particular license/privilege, so this approach should meet our needs for the foreseeable future. This is the type of core logic that results in data migrations/bumping up major versions of an API if anything needs to be changed. So we want to be absolutely sure that this approach works for us, which is why I'm calling this out, even though I don't see anything that needs to be changed about this.

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/common.py
```diff
@@ -57,3 +61,67 @@ def ensure_value_is_datetime(value: str):
...
+        """
+        Sort the keys, values in the dictionary so they are hashed in a predictable order
+        """
+        sorted_keys = sorted(in_dict.keys())
+        return tuple((str(key), str(in_dict[key])) for key in sorted_keys)
+
+    @staticmethod
+    def _b64encode_str(value: str) -> bytes:
+        # We control the 62nd and 63rd b64 chars with `altchars` to prevent clashing with our separators
+        return b64encode(value.encode('utf-8'), altchars=b'+_')
```
landonshumway-ia, line 127: This right here makes me wonder if we should consider choosing other separators if we need to have logic like this... Are there not other options we could use that wouldn't require this?

backend/compact-connect/lambdas/python/provider-data-v1/handlers/__init__.py
```diff
@@ -30,10 +30,18 @@ def get_provider_information(compact: str, provider_id: str) -> dict:
...
+                logger.debug('Identified license update record', provider_id=provider_id)
+                licenses[record['jurisdiction']]['history'].append(record)
             case 'privilege':
                 logger.debug('Identified privilege record', provider_id=provider_id)
-                privileges.append(record)
+                privileges[record['jurisdiction']] = record
+                privileges[record['jurisdiction']].setdefault('history', [])
+            case 'privilegeUpdate':
+                logger.debug('Identified privilege update record', provider_id=provider_id)
+                privileges[record['jurisdiction']]['history'].append(record)
```
landonshumway-ia, line 44: Just need to check my understanding here: with the way this is structured, it suggests that the base license and privilege records will ALWAYS come first in the array before any update records for the corresponding license/privilege record. Is my analysis accurate? Are there any edge cases where this would not be true and this could break?

backend/compact-connect/lambdas/python/provider-data-v1/handlers/ingest.py
```diff
@@ -97,6 +108,65 @@ def populate_provider_record(*, provider_id: str, license_post: dict, privilege_
     )['M']
 
 
+def process_license_update(*, existing_license: dict, new_license: dict, dynamo_transactions: list):
```
landonshumway-ia, line 111: nit: should start with underscore if we don't intend to call it outside of this file: `_process_license_update`

backend/compact-connect/lambdas/python/provider-data-v1/handlers/ingest.py
```diff
@@ -97,6 +108,65 @@ def populate_provider_record(*, provider_id: str, license_post: dict, privilege_
...
+    # dateOfUpdate won't show up as a change because the field isn't in new_license, yet
+    updated_values = {key: value for key, value in new_license.items() if value != existing_license[key]}
+    if not updated_values:
+        return
+    # Categorize the update
+    update_record = populate_update_record(existing_license=existing_license, updated_values=updated_values)
+    dynamo_transactions.append({'Put': {'TableName': config.provider_table_name, 'Item': update_record}})
+
+
+def populate_update_record(*, existing_license: dict, updated_values: dict) -> dict:
```
landonshumway-ia, line 128: same here: `_populate_update_record`

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_providers.py
```diff
@@ -260,6 +267,12 @@ def _when_testing_get_provider_response_based_on_read_access(self, scopes: str,
         provider_data = json.loads(resp['body'])
         self.assertEqual(expected_provider, provider_data)
 
+        # The sk for a license-update record is sensitive so we'll do an extra, pretty broad, check just to make sure
+        # we guard against future changes that might accidentally send the key out via the API. See discussion on
+        # key generation in the LicenseUpdateRecordSchema for details.
+        sensitive_hash = self._get_sensitive_hash()
+        self.assertNotIn(sensitive_hash, resp['body'])
```
landonshumway-ia, line 274: Great check 👍

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/common.py
jusdino: Yeah, good call out. We can do a team design review to mull this over before committing to it.

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/common.py
jusdino: Yeah, we can talk through separator and b64 alphabet character choice when we do a design view. Ultimately, it doesn't matter a lot what we choose for separators vs 62nd, 63rd b64 characters, as long as they don't clash with our separators. I see it as mostly just an aesthetic choice. The [RFC for base64 encoding](https://datatracker.ietf.org/doc/html/rfc4648#section-4) provides two default base64 'alphabets', one of which includes a slash `/`, the other includes a dash `-`. We could adopt one of those more official alphabet choices if we swap out the clashing separator for some other available character.

backend/compact-connect/lambdas/python/provider-data-v1/handlers/__init__.py
jusdino: Yes, your understanding is correct. I don't see any cases where that would not be true for our data model but, now that you point it out, I don't think I want this limitation hiding in the code here, to catch us out if we decide we need to reverse the index scan or something. I'll revise to make it more flexible.

landonshumway-ia (approved):
Looks Great

landonshumway-ia:
This is ready for your review @jlkravitz

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/license/record.py
```diff
@@ -0,0 +1,128 @@
...
+    updatedValues = Nested(LicenseUpdateRecordPreviousSchema(partial=True), required=True, allow_none=False)
+
+    @post_dump  # Must be _post_ dump so we have values that are more easily hashed
+    def generate_pk_sk(self, in_data, **kwargs):  # noqa: ARG001 unused-argument
+        """
+        NOTE: Because the 'sk' field in this record type contains a hash that is generated based on the values of the
+        record itself and because, in some cases, the values could be guessed and verified by the hash with relative
+        ease, regardless of the strength of the hash, we need to treat the 'sk' field as if it were just as sensitive as
+        the most sensitive field in the record. More to the point, we need to be sure that this internal field is never
+        served out via API.
```
jlkravitz, line 108: Probably misunderstanding, but if the raw data is being served via API, what is the risk of the hash being served out via API? What additional data does that expose?

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
```diff
@@ -210,73 +211,239 @@ def create_provider_privileges(
...
+                        'ExpressionAttributeNames': {'#privilegeJurisdictions': 'privilegeJurisdictions'},
+                        'ExpressionAttributeValues': {':newJurisdictions': {'SS': jurisdiction_postal_abbreviations}},
+                    }
+                }
+            )
+
+            # Unfortunately, we can't guarantee that the number of transactions is below the 100 action limit
+            # for extremely large purchases. To handle those large purchases, we will have to break our transactions
+            # up and handle a multi-transaction roll-back on failure.
+            # We'll collect data for sizes, just so we can keep an eye on them and understand user behavior
```
jlkravitz, line 316: Helpful comment, thank you!

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
```diff
@@ -210,73 +211,239 @@ def create_provider_privileges(
...
+
+        # Execute rollback in batches of 100
+        batch_size = 100
+        while rollback_batch := rollback_transactions[:batch_size]:
+            try:
+                logger.info('Submitting rollback transaction', provider_id=provider_id, compact=compact)
+                self.config.dynamodb_client.transact_write_items(TransactItems=rollback_batch)
+                rollback_transactions = rollback_transactions[batch_size:]
+            except ClientError as e:
+                logger.error('Failed to roll back privilege transactions', error=str(e))
```
jlkravitz, line 443: Roll back the roll back? (Just kidding.)

Is this something we should have an alert on? Seems unlikely to get here, but potentially very bad?

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_ingest.py
```diff
@@ -187,21 +164,303 @@ def test_newer_active_license(self):
...
+        from handlers.ingest import ingest_license_message
+
+        provider_id = self._with_ingested_license()
+
+        with open('../common/tests/resources/ingest/message.json') as f:
+            message = json.load(f)
+
+        message['detail'].update({'familyName': 'VonSmitherton'})
+
+        # What happens if their license is renewed in a subsequent upload?
```
jlkravitz, line 354: old comment?

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/license/api.py
```diff
@@ -0,0 +1,145 @@
...
+    phoneNumber = ITUTE164PhoneNumber(required=False, allow_none=False)
+
+    @validates_schema
+    def validate_license_type(self, data, **kwargs):  # noqa: ARG001 unused-argument
+        license_types = config.license_types_for_compact(data['compact'])
+        if data['licenseType'] not in license_types:
+            raise ValidationError({'licenseType': [f'Must be one of: {', '.join(license_types)}.']})
+
+
+class LicenseUpdatePreviousGeneralResponseSchema(ForgivingSchema):
```
jlkravitz, line 65: I assume the explicitness was intentional, but any reason not to share fields with `LicenseGeneralResponseSchema`? Same question for the privilege schemas.

jlkravitz (commented):
good stuff! love the code reorg. few nits/comments/questions.

backend/compact-connect/lambdas/python/provider-data-v1/handlers/__init__.py
```diff
@@ -32,20 +33,24 @@ def get_provider_information(compact: str, provider_id: str) -> dict:
...
+    # Process update records after all base records have been identified
+    for record in provider_data['items']:
+        match record['type']:
+            case 'licenseUpdate':
+                logger.debug('Identified license update record', provider_id=provider_id)
+                licenses[record['jurisdiction']]['history'].append(record)
+            case 'privilegeUpdate':
+                logger.debug('Identified privilege update record', provider_id=provider_id)
+                privileges[record['jurisdiction']]['history'].append(record)
+
```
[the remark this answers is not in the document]
jusdino: This is not about the order of the update records going out via API. It's about the order that the records come out of the database. We process `licenseUpdate` and `privilegeUpdate` records after `provider`, `license`, and `privilege` records, so that we know that the `license` record will be packed into the API return object before we try to insert a `licenseUpdate` record into that license's history.

The way the records are designed, this should always happen anyway, by the way the sort keys of all the records are defined. Splitting out the processing into two separate loops is essentially just about future-proofing this method, since it doesn't _need_ to depend on the query always returning `license` before `licenseUpdate` and `privilege` before `privilegeUpdate`.

backend/compact-connect/docs/design/users-arch-diagram.pdf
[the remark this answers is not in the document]
jusdino: Yeah, pretty much. The scopes listed on the resource servers are the ones we've been referring to as 'coarse grained' scopes, which control access to API endpoints entirely, as opposed to the 'fine grained' scopes, which are enforced within the runtime logic itself. `readPrivate` is specifically a fine grained scope.

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/license/record.py
```diff
@@ -0,0 +1,128 @@
...
+        """
+        in_data['pk'] = f'{in_data['compact']}#PROVIDER#{in_data['providerId']}'
+        # This needs to include a POSIX timestamp (seconds) and a hash of the changes
+        # to the record. We'll use the current time and the hash of the updatedValues
+        # field for this. We'll use the same hash for both the pk and sk, since we
+        # don't need to query on the sk for this record type.
```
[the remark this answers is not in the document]
jusdino: Hm. I can't explain that. I'll remove the offending sentence 🤔

backend/compact-connect/lambdas/python/common/tests/function/test_data_client.py
```diff
@@ -0,0 +1,322 @@
...
+
+        # Store original provider record
+        original_provider = {
+            'pk': f'aslp#PROVIDER#{provider_uuid}',
+            'sk': 'aslp#PROVIDER',
+            'providerId': provider_uuid,
+            'compact': 'aslp',
+            'privilegeJurisdictions': set(jurisdictions),
+        }
+        self._provider_table.put_item(Item=original_provider)
```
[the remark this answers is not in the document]
jusdino: We're passing a more complete-looking provider record into `create_provider_privileges()`. That argument is required by the call signature but is only actually used in the case where an exception is raised and the transactions are rolled back. We need to be more thorough in this case because we have to be more diligent in simulating realistic 'before' data to have a realistic reference state. We can then assert that we have returned to that 'before' state in this test.

backend/compact-connect/lambdas/python/common/tests/unit/test_data_model/test_data_client.py
```diff
@@ -68,29 +68,3 @@ def test_get_provider_not_found(self):
         # Verify it raises CCNotFoundException
         with self.assertRaises(CCNotFoundException):
             self.client.get_provider(compact='aslp', provider_id='test_id', detail=True, consistent_read=False)
-
-    @patch('cc_common.config._Config.current_standard_datetime', datetime.fromisoformat('2024-11-08T23:59:59+00:00'))
-    def test_data_client_deletes_records_if_exception_during_create_privilege_records(self):
```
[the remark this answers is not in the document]
jusdino: Yeah, this case is now tested in the `function/test_data_client.py` module.

backend/compact-connect/lambdas/python/provider-data-v1/handlers/ingest.py
```diff
@@ -65,11 +68,19 @@ def ingest_license_message(message: dict):
     # If at least one active: last issued active license
     # If all inactive: last issued inactive license
     # Set (or replace) the posted license for its jurisdiction
+    existing_license = licenses.get(license_post['jurisdiction'])
+    if existing_license is not None:
+        _process_license_update(
+            existing_license=existing_license,
+            new_license=license_post,
+            dynamo_transactions=dynamo_transactions,
+        )
```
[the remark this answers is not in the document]
jusdino: Previously, we only overwrote the license record with the latest data. Now we add supplementary records that record change history over time. The corresponding API change for licenses only adds a new field to the license object (`history`) so could be called backwards compatible. However, the way privileges are represented in the same response is not backwards compatible. Currently, multiple issuances/renewals of a privilege in the same jurisdiction are represented as separate objects in the provider's `privileges` array. After this change, like licenses, there will only be one privilege object for each state, then there will be a `history` field with change history for that privilege.

backend/compact-connect/stacks/api_stack/v1_api/api_model.py
```diff
@@ -854,13 +854,72 @@ def _provider_detail_response_schema(self):
...
+                                        ),
+                                        'dateOfUpdate': JsonSchema(
+                                            type=JsonSchemaType.STRING, format='date', pattern=cc_api.YMD_FORMAT
+                                        ),
+                                        'previous': JsonSchema(
+                                            type=JsonSchemaType.OBJECT, properties=self._common_license_properties
+                                        ),
+                                        'updatedValues': JsonSchema(
+                                            type=JsonSchemaType.OBJECT, properties=self._common_license_properties
+                                        ),
```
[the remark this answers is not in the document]
jusdino: Yep. Good catch!

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
jusdino: All of this logic is part of an API call and this error would result in a 500 response status being returned. We already have alarms defined that fire for any 5XX responses from the API. We could instrument a dedicated alarm, though, since getting all the way to a failed rollback is an especially bad thing.

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/license/api.py
jusdino: Yeah, we were talking about the schema refactor here and decided that, for Response Schema in particular, we should move to a much more explicit style to reduce coupling across endpoints, and make API future changes much more deliberate. This way, it's a lot harder for a dev to accidentally change the API in a place they didn't intend because a common Schema was used across places they didn't see.

backend/compact-connect/lambdas/python/provider-data-v1/handlers/__init__.py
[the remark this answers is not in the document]
jlkravitz: Ah I got it. Thanks!

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/license/record.py
jlkravitz: Got it. That makes a lot of sense. Thanks for the thorough explanation.

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
jlkravitz: Got it. I leave that up to your discretion!

backend/compact-connect/lambdas/python/provider-data-v1/handlers/ingest.py
```diff
@@ -97,7 +108,66 @@ def populate_provider_record(*, provider_id: str, license_post: dict, privilege_
...
+def _process_license_update(*, existing_license: dict, new_license: dict, dynamo_transactions: list):
+    """
+    Examine the differences between existing_license and new_license, categorize the change, and add
+    a licenseUpdate record to the transaction if appropriate.
+    :param dict existing_license: The existing license record
+    :param dict new_license: The newly-uploaded license record
+    :param list dynamo_transactions: The dynamodb transaction array to append records to
+    """
+    # dateOfUpdate won't show up as a change because the field isn't in new_license, yet
+    updated_values = {key: value for key, value in new_license.items() if value != existing_license[key]}
```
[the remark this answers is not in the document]
jusdino: I did not handle the case where a key is _removed_ in a new license in the first pass. I've updated this logic (and the API) to include a `removedValues` field to represent fields removed in an update.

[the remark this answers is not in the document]
jlkravitz: @jusdino Looks good- do we need to handle the case where a field is _added_? Naively, I wonder if it would work to just do:

```python
    updated_values = {key: value for key, value in new_license.items() if key not in existing_license or value != existing_license[key]}
```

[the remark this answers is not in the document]
jusdino: Oh, of course. I'll switch to a `.get()`.

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
jusdino: Ran it by @isabeleliassen - we created a ticket for this: #449

jusdino:
@jlkravitz , last revision is in!

jlkravitz (approved):
@isabeleliassen This is good to go!

isabeleliassen (approved):

PR #438, opened 2025-01-14
Feat/weekly reports

### LINKED ISSUE #316: Generate weekly transaction reports
## User story: compact commission admins and state admins will want to know how many compact privileges were purchased in a week, who purchased them, which states they were purchased for, and the cost totals so that they can track costs and new practitioners in their state.
 
- For V1, we will only generate reports weekly.
- We will only report on transactions that were settled
- For V1 longer running times is fine, and we do not expect a high volume of transactions initially, but we will want to have alerting in place to notify us when the number of transactions within a week hits a certain threshold (~100) so we can re-evaluate for scalability.

- the following information would be needed for state reporting: transaction date, privilege(s) purchased, which practitioner account they are added to
- needed for financial reporting: transaction date, which practitioner account purchased the privileges, commission fee, state fee

This ticket involves creating a scheduled process that will grab all of the settled transactions for the previous week and generate a report for the privileges that were purchased.

Questions:
1. Where and how are these reports being sent? Email or a reporting page in the system?
2. Who is the intended audience of these reports?

landonshumway-ia:
compact commission admins and state admins will want to know how many compact privileges were purchased in a week, who purchased them, which states they were purchased for, and the cost totals so that they can track payments and new practitioners in their state.

This adds the needed infrastructure/logic to get the list of transactions for the previous week and generate reports for both compacts and all jurisdictions within those compacts.

### Requirements List
- If any compact or jurisdiction in the environment does not have any email addresses listed in their respective reporting list, the transaction reporting lambda will raise an exception, which will fire an alert once a week. This is by design as support teams will need to be notified if reports are not being sent to any recipients.

### Description List
- Added transaction reporting system which will send out transaction reports to all jurisdictions as well as the compact
- Updated email notification service lambda to handle file attachments

### Testing List
- Added tests for python and lambda logic
- Verified functionality in sandbox account
- Code review

Closes #316

### DIFF
--- a/backend/compact-connect/bin/generate_mock_transactions.py
+++ b/backend/compact-connect/bin/generate_mock_transactions.py
@@ -1,0 +1,186 @@
+# ruff: noqa: T201 we use print statements for local scripts
+#!/usr/bin/env python3
+# Script to generate mock transaction data for test environments
+# it spreads the transactions across a range of dates and providers to simulate real-world data.
+#
+# Run from 'backend/compact-connect' like:
+# bin/generate_mock_transactions.py --compact aslp --start_date 01/01/2024 --end_date 03/01/2024 --count 100
+
+import argparse
+import asyncio
+import json
+import os
+import random
+import sys
+from datetime import UTC, datetime
+
+import boto3
+from botocore.config import Config
+
+# Add the provider data lambda runtime to our pythonpath
+provider_data_path = os.path.join('lambdas', 'python', 'common')
+sys.path.append(provider_data_path)
+
+with open('cdk.json') as context_file:
+    _context = json.load(context_file)['context']
+COMPACTS = _context['compacts']
+
+
+def parse_date(date_str: str) -> datetime:
+    """Parse date string in mm/dd/yyyy format to datetime object."""
+    return datetime.strptime(date_str, '%m/%d/%Y').replace(tzinfo=UTC)
+
+
+def get_random_timestamp(start_date: datetime, end_date: datetime) -> datetime:
+    """Generate a random timestamp between start and end dates."""
+    time_diff = end_date.timestamp() - start_date.timestamp()
+    random_time = start_date.timestamp() + random.random() * time_diff
+    return datetime.fromtimestamp(random_time, tz=UTC)
+
+
+def get_table_by_pattern(tables: list[dict], pattern: str) -> str:
+    """Find table name that contains the given pattern."""
+    matching_tables = [t['TableName'] for t in tables if pattern in t['TableName']]
+    if not matching_tables:
+        raise ValueError(f'No table found containing pattern: {pattern}')
+    return matching_tables[0]
+
+
+async def get_provider_ids(provider_table, compact: str) -> set:
+    """Get all provider IDs for the given compact."""
+    provider_ids = set()
+    scan_kwargs = {'FilterExpression': boto3.dynamodb.conditions.Key('sk').eq(f'{compact}#PROVIDER')}
+
+    done = False
+    start_key = None
+    while not done:
+        if start_key:
+            scan_kwargs['ExclusiveStartKey'] = start_key
+        response = provider_table.scan(**scan_kwargs)
+        for item in response.get('Items', []):
+            if 'providerId' in item:
+                provider_ids.add(item['providerId'])
+        start_key = response.get('LastEvaluatedKey', None)
+        done = start_key is None
+
+    print(f'Found {len(provider_ids)} providers for compact {compact}')
+    return provider_ids
+
+
+def generate_transaction(compact: str, start_date: datetime, end_date: datetime, provider_id: str) -> dict:
+    """Generate a single mock transaction."""
+    settlement_time = get_random_timestamp(start_date, end_date)
+    submit_time = get_random_timestamp(start_date, settlement_time)
+    transaction_id = str(random.randint(100000000000, 999999999999))
+    batch_id = '15867123'
+    epoch_timestamp = int(settlement_time.timestamp())
+    month_key = settlement_time.strftime('%Y-%m')
+
+    return {
+        'pk': f'COMPACT#{compact}#TRANSACTIONS#MONTH#{month_key}',
+        'sk': f'COMPACT#{compact}#TIME#{epoch_timestamp}#BATCH#{batch_id}#TX#{transaction_id}',
+        'batch': {
+            'batchId': batch_id,
+            'settlementState': 'settledSuccessfully',
+            'settlementTimeLocal': settlement_time.strftime('%Y-%m-%dT%H:%M:%S'),
+            'settlementTimeUTC': settlement_time.strftime('%Y-%m-%dT%H:%M:%S.000Z'),
+        },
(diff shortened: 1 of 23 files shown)


backend/compact-connect/lambdas/python/common/cc_common/data_model/transaction_client.py
```diff
@@ -39,3 +39,86 @@ def store_transactions(self, compact: str, transactions: list[dict]) -> None:
...
+        :return: List of transactions for the month
+        """
+        all_matching_transactions = []
+        last_evaluated_key = None
+        while True:
+            if last_evaluated_key:
+                query_params['ExclusiveStartKey'] = last_evaluated_key
+
+            response = self.config.transaction_history_table.query(
+                KeyConditionExpression=('pk = :pk AND sk BETWEEN :start_sk AND :end_sk'),
```
jusdino, line 109: I think we can afford to be a bit more conservative here with our time window logic. BETWEEN is inclusive of both its beginning and end values, which makes it very easy to over- or under-report transactions, when there happen to be some right on the edge of our time window. Your logic here works as you intend, so long as `23:59:59.000` is really the latest timestamp possible in the window, which I think is true in this case, since I think your timestamps are all whole-second values. If that ever changes, though, we could have a very sneaky bug here leading to some accounting errors. Instead, we can set our end time to be `1:00:00`, just _after_ the time window we want to fully capture, then use `sk >= start_sk AND sk < end_sk`, which will capture all transactions in the window, regardless of time granularity.

A second minor thought: Since you're using the boto3 `Table` here, you can use `boto3.dynamodb.conditions.Key` to simplify your key condition expression: `Key('pk').eq(pk) & Key('sk').gte(start_sk) & Key('sk').lt(end_sk)`, then you don't have to even provide `ExpressionAttributeValues` - that's all calculated for you. [example](https://github.com/csg-org/CompactConnect/blob/b6edd2719d1243c8e937f18de794d1722f2eae5a/backend/compact-connect/lambdas/python/common/cc_common/data_model/client.py#L86)

jusdino (changes requested):
Really good! A few changes - biggest thing is probably making sure we load test this one lambda.

jusdino (approved):
Good to go! @jlkravitz , ready for you.

backend/compact-connect/lambdas/python/purchases/handlers/transaction_reporting.py
```diff
@@ -0,0 +1,332 @@
...
+            logger.error(
+                'Failed to send jurisdiction report email',
+                compact=compact,
+                jurisdiction=jurisdiction,
+                error=jurisdiction_response.get('FunctionError'),
+            )
+            lambda_error_messages.append(jurisdiction_response.get('FunctionError'))
+
+    if lambda_error_messages:
+        raise CCInternalException(f'Failed to send one or more reports. Errors: {lambda_error_messages}')
```
jlkravitz, line 137: Is this necessarily true for some of the errors, e.g. "Some providers were not found in the database."?

backend/compact-connect/lambdas/python/purchases/tests/function/test_handlers/test_transaction_reporting.py
```diff
@@ -0,0 +1,512 @@
...
+        'Payload': '{"message": "Email message sent"}',
+        'ExecutedVersion': '1',
+    }
+
+
+@mock_aws
+class TestGenerateTransactionReports(TstFunction):
+    """Test the process_settled_transactions Lambda function."""
+
+    def add_mock_provider_to_db(self, licensee_id, first_name, last_name) -> dict:
```
jlkravitz, line 117: Should this be prefixed with an underscore since it's a private method?

backend/compact-connect/lambdas/python/purchases/tests/function/test_handlers/test_transaction_reporting.py
```diff
@@ -0,0 +1,512 @@
...
+            record = json.load(f, object_hook=privilege_jurisdictions_to_set, parse_float=Decimal)
+            record['providerId'] = licensee_id
+            record['pk'] = f'{TEST_COMPACT}#PROVIDER#{licensee_id}'
+            record['givenName'] = first_name
+            record['familyName'] = last_name
+            self._provider_table.put_item(Item=record)
+
+        return record
+
+    def add_mock_transaction_to_db(
```
jlkravitz, line 133: Should this be prefixed with an underscore since it's a private method?

backend/compact-connect/lambdas/python/purchases/tests/function/test_handlers/test_transaction_reporting.py
```diff
@@ -0,0 +1,512 @@
...
+        # Verify each jurisdiction report shows the correct privilege
+        for jurisdiction in ['ky', 'ne', 'oh']:
+            jurisdiction_call = next(
+                call for call in calls_args[1:] if json.loads(call[1]['Payload'])['jurisdiction'] == jurisdiction
+            )
+            jurisdiction_payload = json.loads(jurisdiction_call[1]['Payload'])
+            self.assertIn(
+                f'{mock_user['givenName']},{mock_user['familyName']},{mock_user['providerId']},03-30-2025,100,{jurisdiction.upper()},10.50,{MOCK_TRANSACTION_ID}',
+                jurisdiction_payload['templateVariables']['jurisdictionTransactionReportCSV'],
+            )
```
jlkravitz, line 359: Should we also assert that transactions from other jurisdictions are _not_ in the report specific to this jurisdiction?

backend/compact-connect/lambdas/python/purchases/tests/function/test_handlers/test_transaction_reporting.py
```diff
@@ -0,0 +1,512 @@
...
+
+        generate_transaction_reports(generate_mock_event(), self.mock_context)
+
+        calls_args = mock_lambda_client.invoke.call_args_list
+        compact_call_payload = json.loads(calls_args[0][1]['Payload'])
+
+        # Verify summary totals
+        self.assertEqual(
+            'Total Transactions,600\n'
+            'Total Compact Fees,$6300.00\n'  # $10.50 x 600
```
jlkravitz, line 398: Thanks for adding these math comment/footnotes!

jlkravitz (commented):
Few questions!

backend/compact-connect/lambdas/python/purchases/handlers/transaction_reporting.py
landonshumway-ia: Good point, I've updated the message.

backend/compact-connect/bin/generate_mock_transactions.py
```diff
@@ -0,0 +1,186 @@
+# ruff: noqa: T201 we use print statements for local scripts
```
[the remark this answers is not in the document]
landonshumway-ia: I've renamed generate_mock_data to be more explicit in what it's doing.

jlkravitz (approved):

isabeleliassen (approved):

Issue #354, opened 2024-11-21
Add Past Privileges to LicenseeDashboard


ChiefStief:
Add Past Privileges section to LicenseeDashboard

Recent Privileges are defined as the most recent privilege for each state.

Past privileges are others associated with user

PR #448, opened 2025-01-16
Feat/license detail page

### LINKED ISSUE #402: frontend update to licensee detail view
## user story: compact and state admins want to view detailed information about practitioners, such as past privileges, DOB, license number, and current privileges so that they can easily find information they need about a practitioner. 
-  title but no dummy variable for license number, delete NPI
- birth/month limited to those who have admin account in relevant jurisdiction
- SSN field will not return any information

ChiefStief:
### Requirements List
- A practitioner account
- A read private account
- A non read private staff account
- All these accounts must be in the same compact
- To create a privilege history for the past privileges section:
1) Log in as practitioner user and purchasing a privilege
3) Edit the DB record to back-date it a year
4) Purchase the same privilege again

### Description List
- Modified `LicenseeDetail` page to match design
- Modified get licensee actions and mutations to work with current API
- Modified `LicenseeDashboard` page to match spec
- Modified `License` Model to reflect updated API where renewals do not create new licenses or privileges but entries in the history array
- Created `LicenseHistoryItem` model
- Created / updated all tests as needed
- Updated mockAPi

### Testing List
- Code review
- Look at the `LicenseeDashboard` page and `LicenseeDetail` page in all screen-sizes and confirm they look good
- For the `LicenseeDetail` page look at the page with a read private account and confirm the DOB  exists and is populated
- For the `LicenseeDetail` page look at the page with a non read private account and confirm the Birth Month-Day  exists and is populated

Closes #402, #354


webroot/src/models/License/License.model.ts
```diff
@@ -60,6 +61,7 @@ export class License implements InterfaceLicense {
     public renewalDate? = null;
     public expireDate? = null;
     public occupation? = null;
+    public history? = null;
```
jsandoval81, line 64: Let's also add this prop to the `InterfaceLicense` as `Array <LicenseHistoryItem>`. Here we can instantiate the default as `[]`. Avoiding null for array types is generally easier to work with - just counting on it always being an array.

webroot/src/models/LicenseHistoryItem/LicenseHistoryItem.model.ts
```diff
@@ -0,0 +1,62 @@
...
+import deleteUndefinedProperties from '@models/_helpers';
+
+// ========================================================
+// =                       Interface                      =
+// ========================================================
+export interface InterfaceLicenseHistoryItem {
+    type?: string | null;
+    updateType?: string | null;
+    previousValues?: object | null;
+    updatedValues?: object | null;
```
jsandoval81, line 17: We can probably type these 2 props in more detail. They probably don't need separate models yet. Just something like:

```
previousValues?: {
    compactTransactionId?: string | null,
    dateOfExpiration?: string | null,
    dateOfIssuance?: string | null,
    dateOfRenewal?: string | null,
    dateOfUpdate?: string | null,
};
updatedValues?: {
    compactTransactionId?: string | null,
    dateOfExpiration?: string | null,
    dateOfIssuance?: string | null,
    dateOfRenewal?: string | null,
};
```

I'm not sure if the `null`s are necessary - you can make that call as needed.

If this works, then you can also make the default values in the class `{}`. That seems easier to always expect it to be a real object rather than handling null cases.

webroot/src/models/User/User.model.spec.ts
```diff
@@ -61,7 +61,7 @@ describe('User model', () => {
                     ],
                 },
             ],
-            accountStatus: 'active',
+            accountStatus: 'active'
```
jsandoval81, line 64: This doesn't seem super necessary for this PR. I'm curious if this is something your editor yells about or does automatically?

webroot/src/pages/LicenseeDashboard/LicenseeDashboard.ts
```diff
@@ -172,6 +125,29 @@ export default class LicenseeDashboard extends Vue {
         return this.$t('licensing.licenseExpiredMessage');
     }
 
+    get pastPrivilegeList(): Array<License> {
+        const privilegeList: Array<License> = [];
+
+        this.licenseePrivileges.forEach((privilege) => {
+            if (privilege.history) {
+                (privilege.history as Array<any>).forEach((historyItem) => {
```
jsandoval81, line 133: This casting can probably go away if the updates to the model noted above are made. You may also be able to eliminate the `if` block by using `privilege.history?.forEach(` - or maybe with the updated typing the `?` may not even be needed.

webroot/src/pages/LicenseeDashboard/LicenseeDashboard.vue
```diff
@@ -55,16 +55,44 @@
         </div>
         <div class="privilege-section">
             <div class="privilege-section-title-row">
-                <div class="privilege-logo-container">
-                    <img class="home-state-img" src="@assets/images/black-ellipse.svg" alt="Privilege List Logo" />
+                <div class="title-info">
+                    <div class="privilege-logo-container">
+                        <img class="home-state-img" src="@assets/images/black-ellipse.svg" alt="Privilege List Logo" />
```
jsandoval81, line 60: `alt` text should ideally be translated. Looks like the language files would just need 1 new entry for ~"Privilege section".

webroot/src/pages/LicenseeDashboard/LicenseeDashboard.vue
```diff
@@ -55,16 +55,44 @@
...
+                    :privilege="privilege"
+                    class="no-touch-item"
+                />
+            </div>
+        </div>
+         <div class="privilege-section bottom">
+            <div class="privilege-section-title-row">
+                <div class="title-info">
+                    <div class="privilege-logo-container">
+                        <img class="home-state-img" src="@assets/images/black-ellipse.svg" alt="Privilege List Logo" />
```
jsandoval81, line 83: Same as previous comment.

webroot/src/pages/LicensingDetail/LicensingDetail.less
```diff
@@ -7,16 +7,23 @@
...
+        background-color: @white;
+
+        @media @tabletWidth {
+            margin-top: -9rem;
+            padding: 9rem @spacingTablet @spacingTablet @spacingTablet;
+        }
 
         @media @desktopWidth {
-            padding: @spacingMobile @spacingDesktop;
+            margin-top: -9rem;
```
jsandoval81, line 25: Negative margins are really dicey. They generally come back to bite when making updates. Is there an alternative here?

webroot/src/pages/LicensingDetail/LicensingDetail.less
```diff
@@ -35,122 +42,248 @@
...
+                    font-size: 1.6rem;
+                }
+
+                .tag-text {
+                    margin-left: 0.5rem;
+                }
+
+                .tag-icon-container {
+                    width: 1.4rem;
+                    margin-bottom: -0.2rem;
```
jsandoval81, line 72: Same as previous comment.

webroot/src/pages/LicensingDetail/LicensingDetail.ts
```diff
@@ -69,6 +89,161 @@ export default class LicensingDetail extends Vue {
...
+    get addressLine1(): string {
+        return this.licensee?.address?.street1 || '';
+    }
+
+    get addressLine2(): string {
+        return this.licensee?.address?.street2 || '';
+    }
+
+    get addressLine3(): string {
+        return `${this.licensee?.address?.city}, ${this.licensee?.address?.state?.abbrev} ${this.licensee?.address?.zip}`;
```
jsandoval81, line 127: If there's not a city, this looks like it will include the comma & space anyway. Is there a quick conditional we can perform to keep it cleaner? Maybe we also don't have to t deeply traverse `this.licensee.address` 3 times. I think there are cases where CSG stresses 1-liners, but this version may have actual performance & display drawbacks.

E.g.
```
const { address = {}} = this.licensee || {};
const { city = '', state: state?.abbrev = '', zip = '' } = address;
const delim = (city && state) ? ', ' : '';

return `${city}${delim}${state} ${zip}`.trim();
```

webroot/src/pages/LicensingDetail/LicensingDetail.ts
```diff
@@ -69,6 +89,161 @@ export default class LicensingDetail extends Vue {
...
+
+    get licenseExpiredMessage(): string {
+        return this.$t('licensing.licenseExpired');
+    }
+
+    get statusTitleText(): string {
+        return this.$t('licensing.status').toUpperCase();
+    }
+
+    get status(): string {
```
jsandoval81, line 154: Does the outer computed name maybe make more sense as `militaryStatus` and the inner function body can use `status`?

webroot/src/pages/LicensingDetail/LicensingDetail.ts
```diff
@@ -69,6 +89,161 @@ export default class LicensingDetail extends Vue {
...
+
+        return militaryStatus;
+    }
+
+    get affiliationTypeTitle(): string {
+        return this.$t('military.affiliationType').toUpperCase();
+    }
+
+    get affiliationType(): string {
+        let militaryStatus = '';
```
jsandoval81, line 169: The computed is `affiliationType` but the inner return is `militaryStatus`. Is that intentional?

webroot/src/pages/LicensingDetail/LicensingDetail.ts
```diff
@@ -69,6 +89,161 @@ export default class LicensingDetail extends Vue {
...
+        return this.$t('licensing.militaryAffilitionDocs').toUpperCase();
+    }
+
+    get militaryDocumentHeader(): any {
+        return { name: this.$t('military.fileName'), date: this.$t('military.dateUploaded') };
+    }
+
+    get listId(): string {
+        return 'military-affiliations';
+    }
```
jsandoval81, line 197: Is this something that will expand later? Or can it just be a string literal at the call site?

webroot/src/pages/LicensingDetail/LicensingDetail.vue
```diff
@@ -18,83 +18,212 @@
...
-                <div class="license-row-cell residence-location">
-                    <span v-if="$matches.phone.only" class="cell-title">{{ $t('licensing.residenceLocation') }}: </span>
+        <div class="personal-information-section">
+            <div class="title-row">
+                <div class="title-info">
+                    <div class="license-logo-container">
+                        <img
+                            class="home-state-img"
+                            src="@assets/images/black-ellipse.svg"
+                            alt="Personal Informational List Logo"
```
jsandoval81, line 59: Same as previous comment about `alt` text. Seems fine to use the same ~"License section" text as before for all the cases on this page.

webroot/src/pages/LicensingDetail/LicensingDetail.vue
```diff
@@ -18,83 +18,212 @@
...
                     </div>
                 </div>
+                <CollapseCaretButton
+                    @toggleCollapse="toggleRecentPrivsCollapsed"
+                />
+            </div>
+            <div v-if="!isRecentPrivsCollapsed" class="privilege-card-list-container">
+                <PrivilegeCard
+                    v-for="(privilege, index) in licenseePrivileges"
+                    :key="'privilege'+index"
```
jsandoval81, line 200: This is interesting. Was Vue complaining without the static prefix here?

webroot/src/pages/LicensingDetail/LicensingDetail.vue
```diff
@@ -18,83 +18,212 @@
...
                     </div>
                 </div>
+                <CollapseCaretButton
+                    @toggleCollapse="togglePastPrivsCollapsed"
+                />
+            </div>
+            <div v-if="!isPastPrivsCollapsed" class="privilege-card-list-container">
+                <PrivilegeCard
+                    v-for="(privilege, index) in pastPrivilegeList"
+                    :key="'privilege'+index"
```
jsandoval81, line 223: Same question as previous.

webroot/src/store/license/license.state.ts
```diff
@@ -12,6 +12,7 @@ export interface State {
     prevLastKey: string | null;
     lastKey: string | null;
     isLoading: boolean;
+    isLoadingOne: boolean;
```
jsandoval81, line 15: Just for my reference, how is `isLoadingOne` used differently than `isLoading` throughout the app?

jsandoval81 (changes requested):
Looking solid 👍 Nice clear naming of things.

webroot/src/store/license/license.state.ts
ChiefStief: isLoadingOne triggers the loading overlay, isLoading does not.

When the License Search page is loading the overlay doesn't pop up, the list itself has a loading spinner. I didn't want to affect that but did want to have a loading overlay for the licensee detail page.

Because some of the data is already present, but not all of it due the difference between get one and get all responses, the initial view was confusing because it was half populated while waiting for the rest of the data to come back.

 I could have implemented a few different component level loading spinners like the list view but it was simpler to just do the overlay.

webroot/src/models/User/User.model.spec.ts
ChiefStief: No, I think I accidentally added a prop here deleted it and was cleaning up after myself. Later, looking at it I didn't feel like reverting the change was necessary as not having the trailing comma is more correct in some ways. IM generally ambivalent here as one argument for always having trailing commas is that line diffs more accurately reflect the content of the changes. 

However, some times the linter does complain about trailing commas, off the top of my head I dont remember which instances it does and which it doesnt.

webroot/src/pages/LicenseeDashboard/LicenseeDashboard.ts
ChiefStief: I can get rid of the if block with the changes, but getting rid of the casting causes alot of other issues.

webroot/src/pages/LicensingDetail/LicensingDetail.ts
ChiefStief: No, I think I copy pasted code and didnt catch the misalignment

webroot/src/pages/LicensingDetail/LicensingDetail.ts
ChiefStief: String literal would be fine

webroot/src/pages/LicensingDetail/LicensingDetail.vue
ChiefStief: I slacked you about this

webroot/src/pages/LicenseeDashboard/LicenseeDashboard.ts
jsandoval81: Does moving from casting to type declaration make any difference? E.g.:

```
this.licenseePrivileges.forEach((privilege: License) => {
    ...
    privilege.history?.forEach((historyItem: LicenseHistoryItem) => {
        ...
```

I've been preferring type-declaration over type-casting up to this point. I feel like even if (for some reason) that last inner forEach argument had to be declared as `(historyItem: any) =>` that at might at least be more consistent and less verbose than the `(privilege.history as Array<any>)`. Feel free to move this over to slack if that's easier.

jsandoval81 (changes requested):
Looks really good. Tests well across browsers incl. the mobile simulator. 🚀 

Just a few follow-on questions below about 2 items before the renewal updates.

webroot/src/store/license/license.state.ts
jsandoval81: Another option is to just control loading UI from the page component itself. Even if it's not a custom spinner component like the Lists have, you could just have these licensee detail pages just render either the `PageLoadingMask` or the `LoadingSpinner` if `licenseStore.isLoading === true`. This actually feels easier.

jsandoval81 (approved):

jsandoval81:

webroot/src/pages/LicensingDetail/LicensingDetail.ts
```diff
@@ -77,4 +257,40 @@ export default class LicensingDetail extends Vue {
...
+
+    checkIfLicenseActive(license: License) {
+        let isLicenseActive = false;
+
+        if (license && license.statusState === LicenseStatus.ACTIVE) {
+            isLicenseActive = true;
+        }
+
+        return isLicenseActive;
+    }
```
jlkravitz, line 269: nit, but I like to prefix boolean functions with is, has, etc. I also think this is simple enough to make into a one-liner, e.g.

```suggestion
    isLicenseActive(license: License) {
        return license?.statusState === LicenseStatus.ACTIVE);
    }
```

webroot/src/store/license/license.mutations.ts
```diff
@@ -65,7 +65,7 @@ export default {
         if (licensee.id) { // Don't put objects with NULL IDs in the store
             if (state.model && state.model.length) {
                 const licenseeToUpdateIndex = state.model
-                    .findIndex((p: any) => Number(p.id) === Number(licensee.id));
+                    .findIndex((p: any) => p.id === licensee.id);
```
jlkravitz, line 68: What changed here?

jlkravitz (commented):
Looks good! Small nit/question.

ChiefStief:
Ready for re-review @jlkravitz

jlkravitz (dismissed):

isabeleliassen (approved):

isabeleliassen (approved):

jlkravitz (approved):

PR #461, opened 2025-01-21
Set API allowed origin headers

### LINKED ISSUE #165: Add CORS non-wildcard allowed-origin headers across API
We can improve the security of our browser app by restricting browser access to the domains we control (or potentially a list of known domains, if states want to integrate via their own browser app for some reason).

AC:
- All API endpoints to return a specific origin in the `Access-Control-Allow-Origin`
- wildcards or `null` are never returned by the API
- The API allowed-origin header configuration can support one or more origins, including localhost, if local ui support is enabled

jusdino:
### Description List
For `Access-Control-Allowed-Origin` headers:
- Added specific/multiple origin support in OPTIONS preflight methods
- Added specific/multiple origin support for headers in all Lambda API responses
- Added specific/ *single* origin support for headers in custom Gateway Responses

Note that [*Gateway Responses*](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-gatewayResponse-definition.html) are API responses that API Gateway itself returns, when it short-circuits a request before delivering it to our Lambda back-end. These responses are for requests that:
- Fail input validation for a POST body or query params,
- Are missing a valid authentication token, or
- Are missing required scopes for the endpoint.

These are specific 400, 401, and 403 responses in specific cases. In those cases, API Gateway has limited support for us customizing the content of the response. That customization does _not_ include calculating a dynamic value based on the request `Origin` header, like we are doing for the other endpoints/responses. Because of this limitation, in the case of these Gateway Responses, we can only support a single, static, value for the allowed origin header or, if we want to support both a proper domain name and `http://localhost:3018` as origins (like in the case of IA's test environment, where we want a common deployed test UI and also for devs to be able to connect their local UI to the shared test environment back-end), we have to still return a `'*'` as allowed origin to support both at once. This limitation does not constrain the production environment or any environments that will have a single allowed origin, since we _can_ set the single origin in the header as a static value.

### Testing List
- Code review
- Test browser behavior in IA's *test* environment

Closes #
Closes #165


landonshumway-ia (approved):

backend/compact-connect/lambdas/python/common/tests/unit/test_api_handler.py
```diff
@@ -95,3 +98,48 @@ def lambda_handler(event: dict, context: LambdaContext):  # noqa: ARG001 unused-
...
+    def test_disallowed_origin(self):
+        from cc_common.utils import api_handler
+
+        @api_handler
+        def lambda_handler(event: dict, context: LambdaContext):  # noqa: ARG001 unused-argument
+            return {'message': 'OK'}
+
+        with open('tests/resources/api-event.json') as f:
+            event = json.load(f)
+        event['headers']['Origin'] = 'https://example.com'  # not in ALLOWED_ORIGINS
```
jlkravitz, line 126: Isn't `ALLOWED_ORIGINS` set to `["https://example.org", "http://localhost:3018"]` in `backend/compact-connect/lambdas/python/common/tests/__init__.py`?

backend/compact-connect/lambdas/python/common/tests/unit/test_api_handler.py
```diff
@@ -95,3 +98,48 @@ def lambda_handler(event: dict, context: LambdaContext):  # noqa: ARG001 unused-
...
+
+        resp = lambda_handler(event, self.mock_context)
+        self.assertEqual(200, resp['statusCode'])
```
jlkravitz, line 129: Wouldn't we not expect a 200 response for an unallowed origin?

backend/compact-connect/lambdas/python/common/tests/unit/test_api_handler.py
```diff
@@ -95,3 +98,48 @@ def lambda_handler(event: dict, context: LambdaContext):  # noqa: ARG001 unused-
...
+        @api_handler
+        def lambda_handler(event: dict, context: LambdaContext):  # noqa: ARG001 unused-argument
+            return {'message': 'OK'}
+
+        with open('tests/resources/api-event.json') as f:
+            event = json.load(f)
+        event['headers']['Origin'] = 'http://localhost:3018'
+
+        resp = lambda_handler(event, self.mock_context)
+        self.assertEqual('http://localhost:3018', resp['headers']['Access-Control-Allow-Origin'])
```
jlkravitz, line 115: Can we use `local_ui_port` here instead of hardcoding 3018, or does that fall under the restrictions you described in the PR description?

jlkravitz, line 115: Do we intentionally not check for status 200 here like we do in the other tests?

backend/compact-connect/stacks/api_stack/cc_api.py
```diff
@@ -169,24 +169,28 @@ def __init__(
         self.log_groups = [access_log_group, self.web_acl.log_group]
         self._configure_alarms()
 
+        # These canned Gateway Response headers do not support dynamic values, so we have to set a static value for the
+        # Access-Control-Allow-Origin header. If we need to support multiple origins, we will have to just set
+        # allow origin '*'.
+        gateway_response_origin = stack.allowed_origins[0] if len(stack.allowed_origins) == 1 else '*'
```
jlkravitz, line 175: Perhaps a silly question, but at this point in the code, is `allowed_origins` not static? Why can't we directly dump the list of origins into the gateway response?

backend/compact-connect/lambdas/python/common/cc_common/utils.py
```diff
@@ -56,9 +56,17 @@ def api_handler(fn: Callable):
...
+        headers = event.get('headers') or {}
+        headers.pop('Authorization', None)
         (event.get('multiValueHeaders') or {}).pop('Authorization', None)
 
+        # Determine the appropriate CORS origin header value
+        origin = headers.get('Origin')
+        if origin in config.allowed_origins:
+            cors_origin = origin
+        else:
+            cors_origin = config.allowed_origins[0]
```
jlkravitz, line 68: Does the fact that this takes the first origin matter? (vs. the 2nd, 3rd, etc.)

jlkravitz (commented):
Few questions!

backend/compact-connect/lambdas/python/common/tests/unit/test_api_handler.py
jusdino: That's right. Note the `.org` vs `.com`, though. We're sticking to [reserved example domain names](https://datatracker.ietf.org/doc/html/rfc6761#section-6.5) for test data, so you have to keep a sharp eye out for those differences.

backend/compact-connect/lambdas/python/common/tests/unit/test_api_handler.py
jusdino: I'm not sure how `local_ui_port` is relevant here. We're testing that the example local origin provided via environment variable is returned in the header when it matches the origin in the request header. I could change the example to `http://localhost:1234` if that makes it less confusing?

backend/compact-connect/stacks/api_stack/cc_api.py
jusdino: We cannot put multiple origins in the response, so this line is picking one.

backend/compact-connect/lambdas/python/common/cc_common/utils.py
jusdino: Yes. The first origin is the 'default', where any subsequent origins will only be returned if it matches the request origin.

backend/compact-connect/lambdas/python/common/tests/unit/test_api_handler.py
jlkravitz: Ah got it. Yeah- let's change to 1234 to make it explicit that it's distinct from the standard local UI port.

backend/compact-connect/lambdas/python/common/tests/unit/test_api_handler.py
jlkravitz: Got it. I was misunderstanding...thanks for the explanation!

jusdino:
@jlkravitz , all set.

jlkravitz (approved):

isabeleliassen (approved):

PR #462, opened 2025-01-21
fix/make file extension matching case-insensitive

landonshumway-ia:
File extension matching should not fail if extension is a valid type by uses different casing


jlkravitz (approved):

isabeleliassen (approved):

Issue #302, opened 2024-11-05
connect screens of privilege purchase workflow


isabeleliassen:
- make sure to add routeguards to disallow user to skip ahead via manual url navigation
- progress bar
- make after all user workflow screens exist

ChiefStief:
make sure to add routeguards to disallow user to skip ahead via manual url navigation

jsandoval81:
Consider adding a top level route at `/Privileges`. This has a few advantages:

- Gives the main app nav router-link a parent path so the nav item can stay highlighted throughout the workflow
- Acts as a "home" page for purchasing that handles kicking off the workflow, or any error checks / redirects

ChiefStief:
- Clean up file names to be PrivilegesPurchase*
- Make sure route paths are consistent
- progress bar
- Audit back, next, cancel etc buttons
- PrivilegesPurchaseState storage

ChiefStief:
Make sure attestations are being send up on purchase

atringo:
Verified all requirements detailed in ticket that can be tested on frontend; workflow completed in test on Chrome desktop and mobile (Android Pixel 6a).

Was not able to complete purchase, separate ticket #590

PR #478, opened 2025-01-22
frontend/attestations page

### LINKED ISSUE #335: attestation page frontend
user story: As a state administrator, I want to make sure practitioners who buy privileges are making legally binding agreements so that they can be disciplined in the future if they lied on their application.
 
- dependent on #304 
- frontend will dynamically fetch the attestation from the backend for each purchase
- include scope of practice attestation if possible
- include personal information attestations
- can we only call the military attestation to show if the person has confirmed military affiliation?

jsandoval81:
### Requirements List
- _None_

### Description List
- Updated InputCheckbox to allow a validation message
- Added PrivilegeAttestation model & store / network updates
- Update PrivilegePurchaseAttestation page to match design
- Added `purchase` prop to User store to hold purchase data; will be expanded / completed in #302 

### Testing List
- Code review
- As a licensee, begin the purchase workflow & get to the attestations page
    - All attestations should require a selection before moving on

Closes #335


ChiefStief (commented):
This all looks good except that we are not saving which option the user selects (under investigation vs not).

It could be argued that this not in the scope of the ticket but I somewhat expected it to be.

jsandoval81:
@ChiefStief I pushed a commit based on a response from the backend team regarding all of the radio-type attestation IDs being required:

> We would only post the one that the user selected. You can disregard the 'required' field in this case, as the backend understands those two are a radio-button selection

ChiefStief (approved):

jsandoval81:

webroot/src/models/PrivilegeAttestation/PrivilegeAttestation.model.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+
+        // Test methods
+        expect(privilegeAttestation.dateCreatedDisplay()).to.equal(
+            moment(data.dateCreated, serverDateFormat).format(displayDateFormat)
+        );
+        expect(privilegeAttestation.dateUpdatedDisplay()).to.equal(
+            moment(data.dateOfUpdate, serverDateFormat).format(displayDateFormat)
+        );
+    });
+    it('should prepare a Staff User for server request through serializer', () => {
```
jlkravitz, line 125: should this be...

```suggestion
    it('should prepare a PrivilegeAttestation for server request through serializer', () => {
```

webroot/src/pages/PrivilegePurchaseAttestation/PrivilegePurchaseAttestation.ts
```diff
@@ -89,11 +199,73 @@ export default class PrivilegePurchaseAttestation extends mixins(MixinForm) {
...
+                this.isFormSuccessful = true;
+                this.$store.dispatch('user/setAttestationsAccepted', true);
+                this.endFormLoading();
+                this.$router.push({
+                    name: 'FinalizePrivilegePurchase',
+                    params: { compact: this.currentCompactType }
+                });
+            }
+
+            this.endFormLoading();
```
jlkravitz, line 221: Is this called twice if the above conditional block runs? (And is that a problem?)

webroot/src/store/user/user.state.ts
```diff
@@ -26,6 +26,7 @@ export interface State {
     currentCompact: Compact | null;
     selectedPrivilegesToPurchase: Array<string> | null;
     arePurchaseAttestationsAccepted: boolean;
+    purchase: any;
```
jlkravitz, line 29: Any reason this can't be typed more narrowly?

jlkravitz (commented):
Looks great! Couple small questions.

webroot/src/models/PrivilegeAttestation/PrivilegeAttestation.model.spec.ts
jsandoval81: Good catch. Fixed.

webroot/src/pages/PrivilegePurchaseAttestation/PrivilegePurchaseAttestation.ts
jsandoval81: If the conditional block runs, the `router.push` will likely unload this component before this line runs. Either way, `endFormLoading` is safe to run again if it does so before unload.

webroot/src/store/user/user.state.ts
jsandoval81: I meant to add a comment here. @ChiefStief & I discussed consolidating the purchase state. This is a step in that direction to avoid collision with another in-progress attestation PR. And the consolidation will happen in #302. I've added a comment to this line.

jlkravitz (approved):
@isabeleliassen Ready to merge.

isabeleliassen (approved):

PR #483, opened 2025-01-22
Simplify password constraints to match server

jsandoval81:
### Requirements List
- _None_

### Description List
- Joi password plugin was not string matching all cases as expected, so we just removed that constraint since it wasn't server enforced anyway.

### Testing List
- Code review
- New password should now just enforce character length only, which aligns exactly with the backend currently.

Hotfix for sprint merge.


jsandoval81:
@jlkravitz This hotfix is ready for your review.

jlkravitz (approved):

Issue #475, opened 2025-01-22
Add a staff-user delete endpoint


jusdino:
Add a DELETE method to the staff-user API. This DELETE method should remove the user's permission for the requested compact. It should do nothing to the Cognito user itself.

PR #486, opened 2025-01-22
Feat/delete staff user

jusdino:
### Description List
- Added DELETE method to /v1/compacts/:compact/staff-users/:userId to delete staff-user's compact permission record
- Identified and fixed a bug in the PATCH /v1/compacts/:compact/staff-users/:userId endpoint, if the user does not exist.
- Removed API change summary from README since it seemed low value to developers and was not being maintained
- Removed OPTIONS methods from openapi spec doc, since they were cluttering up the doc and making it harder to parse.

Closes #
#475

### DIFF
--- a/backend/compact-connect/lambdas/python/staff-users/handlers/users.py
+++ b/backend/compact-connect/lambdas/python/staff-users/handlers/users.py
@@ -2,7 +2,7 @@ import json
 
 from aws_lambda_powertools.utilities.typing import LambdaContext
 from cc_common.config import config, logger
-from cc_common.exceptions import CCNotFoundException
+from cc_common.exceptions import CCAccessDeniedException, CCNotFoundException
 from cc_common.utils import (
     api_handler,
     authorize_compact,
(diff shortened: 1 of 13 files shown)


backend/compact-connect/lambdas/python/staff-users/handlers/users.py
```diff
@@ -125,3 +125,39 @@ def post_user(event: dict, context: LambdaContext):  # noqa: ARG001 unused-argum
...
+    compact = event['pathParameters']['compact']
+    user_id = event['pathParameters']['userId']
+    # If the caller is a jurisdiction admin, they can only delete the user if the user only has permissions in the
+    # caller's jurisdiction. If the user has permissions in another jurisdiction, an admin in that other jurisdiction
+    # must remove those permissions first before this operation will succeed. A compact admin can delete any user in
+    # their compact, regardless of where they have permissions.
+    allowed_jurisdictions = get_allowed_jurisdictions(compact=compact, scopes=get_event_scopes(event))
+
+    # None means they are a compact admin - no jurisdiction restrictions at all
+    if allowed_jurisdictions is not None:
```
landonshumway-ia, line 142: Forgive my terrible memory, but is there a possibility a compact admin user could also have permissions within a jurisdiction for some reason, and this list would not be none? Is the intent here to always let a compact admin delete a user?

landonshumway-ia (commented):
This looks great. Just one question:

backend/compact-connect/lambdas/python/staff-users/handlers/users.py
jusdino: Yes, a compact admin can have jurisidiction permissions. In that case, the value is still `None`. Yes, the intent is that anyone with compact level admin permission will be unrestricted by jurisdiction-specifics.

landonshumway-ia (approved):

landonshumway-ia:
@jlkravitz This is ready for you to review

backend/compact-connect/lambdas/python/staff-users/handlers/users.py
```diff
@@ -125,3 +125,39 @@ def post_user(event: dict, context: LambdaContext):  # noqa: ARG001 unused-argum
...
+        common_jurisdictions = allowed_jurisdictions.intersection(user_jurisdictions)
+
+        # We won't show that the user even exists, if they have no common jurisdictions
+        if not common_jurisdictions:
+            raise CCNotFoundException('User not found')
+
+        # If they have permissions elsewhere, we return an error
+        if disallowed_jurisdictions:
+            raise CCAccessDeniedException(
+                f'User has permissions in other jurisdictions ({', '.join(disallowed_jurisdictions)}). Those must be'
```
jlkravitz, line 156: Two questions:

1. What if a jursidiction admin just wants to remove the user's jurisdiction privileges? Can they do that? Why not just let "delete a user" mean "remove the user's jurisdiction privileges" for this type of user?
2. Any concern re: information leakage? Is there any risk of an OH jurisdiction admin knowing that another OH user also has KY jurisdiction privileges? I don't know exactly how this is used, but perhaps if someone is fired? It might not be in the fired individual's interest that their OH boss knows they also work for KY. (Again, making this up. I leave it up to you to decide if this is a legitimate concern weighed with the value of including this addl info.)

jlkravitz (commented):
one q!

backend/compact-connect/lambdas/python/staff-users/handlers/users.py
jusdino: 1) We have a separate API method for permission addition/removal - PATCH. Deletion is intentionally a step beyond just removing permissions nodes from an existing user.
2) Yeah, we talked about the different access limitations a while back and decided that, if a staff user has permissions in a particular state, we were OK with an admin in that state being able to see that user's permissions for the whole compact. This disallowed jurisdictions pattern is consistent with the conclusions we came to back then.

jlkravitz: Got it- thanks!

jlkravitz (approved):

isabeleliassen (approved):

PR #495, opened 2025-01-24
Feat/reinvite staff user

### LINKED ISSUE #476: Reinvite staff user
Add a new API endpoint for staff users to trigger re-sending of an invitation to an existing user. The endpoint logic will need to check the status of the user and, if necessary, force them into a FORCE_PASSWORD_RESET state, so that they are eligible to receive a resent invite.

jusdino:
### Description List
- Added a `POST staff-user/:userId/reinvite` endpoint that will resend an invite email, resetting the user's password if necessary

### Testing List
- Code review

Closes #
closes #476


jusdino:
We'll need to merge #475 and rebase this branch, before we can merge this one.

backend/compact-connect/lambdas/python/common/cc_common/data_model/user_client.py
```diff
@@ -343,3 +348,45 @@ def delete_user(self, *, compact: str, user_id: str) -> None:
...
+            # If they're in CONFIRMED state, we need to reset their password first
+            if user_data['UserStatus'] == 'CONFIRMED':
+                self.config.cognito_client.admin_set_user_password(
+                    UserPoolId=self.config.user_pool_id,
+                    Username=email,
+                    # We need to reset their password, but they will never use this password, so we
+                    # just need to set it to something random. Note that this value should not be referenced
+                    # outside of this function, as it is a real password and we want it to be cleaned up
+                    # by the garbage collector, as soon as possible.
+                    Password=token_hex(48),
```
landonshumway-ia, line 373: I assume this should always satisfy all of the password constraints defined by the cognito user pool, otherwise it will fail right?

backend/compact-connect/lambdas/python/common/cc_common/data_model/user_client.py
```diff
@@ -343,3 +348,45 @@ def delete_user(self, *, compact: str, user_id: str) -> None:
...
+        """
+        try:
+            # Check their current status
+            user_data = self.config.cognito_client.admin_get_user(
+                UserPoolId=self.config.user_pool_id,
+                Username=email,
+            )
+
+            # If they're in CONFIRMED state, we need to reset their password first
+            if user_data['UserStatus'] == 'CONFIRMED':
```
landonshumway-ia, line 365: So this method is only accounting for 'CONFIRMED' or 'FORCE_RESET_PASSWORD' status values, but there are other potential values right? Shouldn't we be checking for those as well? If it is any other status the resend api call will fail

backend/compact-connect/lambdas/python/common/tests/function/__init__.py
```diff
@@ -51,6 +51,15 @@ def build_resources(self):
...
             UsernameAttributes=['email'],
+            Policies={
+                'PasswordPolicy': {
+                    'MinimumLength': 12,
+                    'RequireUppercase': False,
+                    'RequireLowercase': False,
+                    'RequireNumbers': False,
+                    'RequireSymbols': False,
+                },
+            },
```
landonshumway-ia, line 62: Does this policy align with what is defined in the CDK?

landonshumway-ia (changes requested):
Just one small gap in checking the Cognito Status values

backend/compact-connect/lambdas/python/common/cc_common/data_model/user_client.py
jusdino: Yes. The user pool is following the OWASP ASVS 4.X recommendations here and does not set character-type requirements, so a hex string of 48 bytes will satisfy the password requirements.

backend/compact-connect/lambdas/python/common/tests/function/__init__.py
jusdino: Yes, it does.

backend/compact-connect/lambdas/python/common/cc_common/data_model/user_client.py
jusdino: @landonshumway-ia , I've revised the logic in this method to be more explicit about checking user status:
- If user has `CONFIRMED`, reset their password before reinviting;
- If user has `FORCE_CHANGE_PASSWORD` or `RESET_REQUIRED` state, resend invite;
- If user has any other state, raise an exception as other states are unexpected.

landonshumway-ia: @jusdino I'm not sure about the `REST_REQUIRED` state. As I recall you can only resend the invite using the AdminCreateUser when the cognito user is in the `FORCE_CHANGE_PASSWORD` state, so I believe the API call will fail if the user is in a `REST_REQUIRED` state. BTW, the only way the user is put in this state is if they the user's password is deactivated through the console or using the 'AdminResetUserPassword' api, so I think we can assume they should never end in that state either, and we should just raise an exception.

landonshumway-ia: jk, looks like it works 😆

landonshumway-ia (approved):

backend/compact-connect/lambdas/python/common/cc_common/data_model/user_client.py
jusdino: No, you were right. My test was bad. Tweaking again.

landonshumway-ia (approved):

jusdino:
@jlkravitz , this one is ready for you!

backend/compact-connect/lambdas/python/staff-users/handlers/users.py
```diff
@@ -161,3 +161,27 @@ def delete_user(event: dict, context: LambdaContext):  # noqa: ARG001 unused-arg
...
+    user = config.user_client.get_user_in_compact(compact=compact, user_id=user_id)
+    if allowed_jurisdictions is not None:
+        allowed_jurisdictions = set(allowed_jurisdictions)
+        user = config.user_client.get_user_in_compact(compact=compact, user_id=user_id)
+        user_jurisdictions = user['permissions']['jurisdictions'].keys()
+        common_jurisdictions = allowed_jurisdictions.intersection(user_jurisdictions)
+
+        # We won't show that the user even exists, if they have no common jurisdictions
+        if not common_jurisdictions:
+            raise CCNotFoundException('User not found')
```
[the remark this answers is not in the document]
jusdino: This will return a 404 in the event that a user calls this endpoint with an id for a user they don't have permission to see (i.e. a board admin in Ohio tries to reinvite a user who has no permissions in Ohio).

jusdino:
@jlkravitz all resolved, I think

jlkravitz (approved):

isabeleliassen (approved):

PR #498, opened 2025-01-24
frontend/registration UI

jsandoval81:
### Requirements List
- New local `.env` variable:
    - `VUE_APP_RECAPTCHA_KEY`
        - Use the public key value that corresponds to your environment. If you are using the IA test environment, you can get a key from @jsandoval81 

### Description List
- Updated frontend deployment pipelines with recaptcha keys
- Updated CSP Lambda to allow calls to Google recpatcha
- Updated InputDate component to return empty string instead of null
- Added new API call to license API layer & store actions
- Updated license API layer to include status code in response
- Added registration link to disambiguation Login UI
- Added RegisterLicensee page
- Updated the license occupation lists

### Testing List
- Code review
- Use the mock populate to test that you receive a successful response
    - This won't result in an email received because it doesn't match any licenses.
- You should also be able to make ~4 requests within 15 minutes from the same IP to produce a rate-limit type of error
    - The error should clear after ~15 minutes
- If you're testing against a non-sandbox server environment, you can attempt to use valid info that would successfully locate a license and send an email
    - If you attempt this, it must be for a license that has not already been registered / claimed
    - Also note that the email will be the login email if successful, so you may want to use a testing variant of your real email to make things easier


backend/compact-connect/lambdas/nodejs/cloudfront-csp/index.js
```diff
@@ -278,6 +284,7 @@ const setCspHeader = (requestDomain, headers = {}) => {
                 domains.cognitoStaff,
                 domains.cognitoProvider,
                 cognitoIdpUrl,
+                'https://www.google.com/recaptcha',
```
ChiefStief, line 287: Just for my edification: is there a reason you did not have the trailing slash here?

webroot/src/network/licenseApi/data.api.ts
```diff
@@ -168,6 +168,19 @@ export class LicenseDataApi implements DataApiInterface {
...
     }
 
+    /**
+     * POST Create Licensee Account
+     * @param  {string}        compact A compact type.
+     * @param  {object}        data    The user request data.
+     * @return {Promise<any>}          The server response.
+     */
+    public async createAccount(compact: string, data: object) {
+        const requestData = { ...data, compact };
```
ChiefStief, line 178: I might want to name this `createLicenseeAccount`.

I initially wanted to suggest putting this in the userAPI, are you putting it here thinking that the logic may be used by staff users at some point?

ChiefStief:
On mobile the recaptcha icon is somewhat present and overlapping the inputs. Its not terrible though.

Do we want to do the recaptcha text instead?
Screenshot 2025-01-24 at 5 25 40 PM

ChiefStief:
Selecting a license type throws an error because

`<InputSelect :formInput="formData.licenseType" @input="setCompactFromLicenseType()" />`

`setCompactFromLicenseType()` does not exist. Do you still need the `@input="setCompactFromLicenseType()`"?

ChiefStief (changes requested):
See comments:

Edit: It forced me to write something here

backend/compact-connect/lambdas/nodejs/cloudfront-csp/index.js
jsandoval81: I actually grabbed these from a previous working config. Based on experience, they will all be treated the same by browsers.

webroot/src/network/licenseApi/data.api.ts
jsandoval81: In the outer `/network/data.api.ts` (to which store actions and the like are wired up) I've called it `createLicenseeAccount`. Here in the scoped License API I've simply called it `createAccount`. 

I've included it in the License API since it was sending actual license info (license type, home state, etc); with the objective of matching with a license in the backend. Just the initial reasoning. If later we find this isn't a good home we can totally move it. That goes for anything in these API configs 👍

backend/compact-connect/lambdas/nodejs/cloudfront-csp/index.js
jlkravitz: nit but can we add the slash here for consistency with the other URLs? I think this being the sole URL without a slash begs the question re: why it's missing, so adding it in might reduce that cognitive overhead.

backend/compact-connect/lambdas/nodejs/cloudfront-csp/test/index.test.js
```diff
@@ -125,6 +131,7 @@ const buildCspHeaders = (environment) => {
         cognitoStaffUrl,
         cognitoProviderUrl,
         cognitoIdpUrl,
+        'https://www.google.com/recaptcha',
```
jlkravitz, line 134: Same question here.

webroot/README.md
```diff
@@ -99,6 +99,13 @@
             - Prod: `4mnd3u2rp30ssgnm7dk81jcqsc`
         - _Local_ :arrow_heading_down:
             - `15mh24ea4af3of8jcnv8h2ic10`
+    - **`VUE_APP_RECAPTCHA_KEY`**
+        - _Server_ :arrow_heading_up:
+            - Dev(JCC-Test): `TODO`
+            - Test(CSG-Test): `TODO`
+            - Prod: `TODO`
```
jlkravitz, line 106: Are these actually todo? We have test and prod codes, no?

webroot/src/pages/RegisterLicensee/RegisterLicensee.ts
```diff
@@ -0,0 +1,367 @@
...
+            jurisdiction: licenseState,
+            licenseType,
+        };
+    }
+
+    async handleRecaptcha(data): Promise<void> {
+        const { recaptchaKey, isUsingMockApi } = this.$envConfig;
+
+        if (!isUsingMockApi) {
+            const { grecaptcha } = window as any;
```
jlkravitz, line 300: What's `grecaptcha`? Since this is from `window`, can we add a brief comment describing what this is and where it comes from?

webroot/src/pages/RegisterLicensee/RegisterLicensee.ts
```diff
@@ -0,0 +1,367 @@
...
+    }
+
+    initExtraFields(): void {
+        const passwordInput: HTMLElement = this.$refs.password as HTMLElement;
+
+        passwordInput.style.display = 'inline-block';
+        passwordInput.style.height = '1px';
+        passwordInput.style.width = '1px';
+        passwordInput.style.overflow = 'hidden';
+    }
```
jlkravitz, line 190: Why can't we style this in the Vue HTML template?

webroot/README.md
```diff
@@ -99,6 +99,13 @@
             - Prod: `4mnd3u2rp30ssgnm7dk81jcqsc`
```
jlkravitz, line 99: Per a discussion with @landonshumway-ia last week, I believe we need to update the backend README documentation to describe how to add the frontend recaptcha token to the AWS Code Pipeline. Is that something we can add in this PR?

jlkravitz (commented):
Looks good! Few nits/questions.

backend/compact-connect/lambdas/nodejs/cloudfront-csp/index.js
jsandoval81: Done

backend/compact-connect/lambdas/nodejs/cloudfront-csp/test/index.test.js
jsandoval81: Done

webroot/README.md
jsandoval81: Good catch. I updated these.

webroot/src/pages/RegisterLicensee/RegisterLicensee.ts
jsandoval81: It's google's SDK for fetching a token. I added a comment on that line referencing where the SDK was loaded.

webroot/src/pages/RegisterLicensee/RegisterLicensee.ts
jsandoval81: This is actually a honeypot field. We try to avoid calling it as such in code to not advertise to slightly sophisticated scrapers / attackers. We also have lint rules that disallow inline CSS. Implementing this way is sort of our way of calling it out as distinct to ourselves. If there are ever additional fields (TBD), we'd factor these honey style props so they at least wouldn't be repeated.

webroot/README.md
jsandoval81: Good call. I made a slight update to the backend README to add the distinction for the pipeline files.

webroot/src/pages/RegisterLicensee/RegisterLicensee.ts
jlkravitz: Ah got it. That makes a lot of sense. Is this something we can (safely) document – if not in the code itself, perhaps in a README with a reference to the README here? If you think any of these suggestions might facilitate a scraper, then we can leave as-is.

jsandoval81: That sounds good. I updated the frontend README and added a reference to that in the code.

isabeleliassen (approved):

PR #508, opened 2025-01-30
Feat/monthly reports

### LINKED ISSUE #493: Automated Monthly CSV Reports
This ticket involves updating the reporting system to send out monthly reports. It involves the following changes:

- Compress file attachments in email to reduce size of email message
- Send csv information from transaction reporter lambda to email lambda using S3
- Add event bridge rule for each compact to send out these reports the first day of each month
- ensure that transaction history collector uses fixed hour when collecting transaction history

landonshumway-ia:
CSG has requested that we implement automated monthly transaction reporting. This required several updates to increase the number of transactions that could be handled within a single report. Previously we were passing in the full csv report string to the email-service lambda payload, which was prone to a 6MB file size limit, now we are storing these reports in a compressed zip format in S3, and passing in the S3 key to the email service so it can pull down the file and attach the zip file to the email directly. This adds the needed infrastructure to support monthly reporting in a scaleable manner.

### Requirements List
- No new requirements, this change is backwards compatible with the current system

### Description List
- Added event bridge rule to trigger transaction reporter lambda on the first day of every month
- Added transaction reporting S3 bucket to store compiled csv report files in a compressed format
- Update email-notification-service node lambda to pull down report files from S3
- Fix formatting for reporting email body text to support markdown

### Testing List
-  Python tests added for monthly reporting code verification
- Verification of monthly report generation for over 100k transactions in sandbox environment.

Closes #493

### DIFF
--- a/backend/compact-connect/lambdas/nodejs/tests/email-notification-service-lambda.test.ts
+++ b/backend/compact-connect/lambdas/nodejs/tests/email-notification-service-lambda.test.ts
@@ -239,8 +239,9 @@ describe('EmailNotificationServiceLambda', () => {
             expect(rawEmailString).toContain('Content-Type: application/zip');
             expect(rawEmailString).toContain('Content-Disposition: attachment; filename=aslp-transaction-report.zip');
             expect(rawEmailString).toContain('Weekly Report for Compact ASLP');
-            expect(rawEmailString).toContain('Please find attached the weekly transaction reports for your compact');
-            expect(rawEmailString).toContain('for the period 2024-03-01 to 2024-03-07')
+            expect(rawEmailString).toContain('Please find attached the weekly settled')
+            expect(rawEmailString).toContain('transaction reports for your compact for the period 2024-03-01 to')
+            expect(rawEmailString).toContain('2024-03-07:</p>')
             expect(rawEmailString).toContain('To: summary@example.com');
         });
 
@@ -332,8 +333,9 @@ describe('EmailNotificationServiceLambda', () => {
             expect(rawEmailString).toContain('Content-Type: application/zip');
             expect(rawEmailString).toContain('Content-Disposition: attachment; filename=oh-transaction-report.zip');
             expect(rawEmailString).toContain('Ohio Weekly Report for Compact ASLP');
-            expect(rawEmailString).toContain('Please find attached the weekly transaction report for your');
-            expect(rawEmailString).toContain('jurisdiction for the period 2024-03-01 to 2024-03-07');
+            expect(rawEmailString).toContain('Please find attached the weekly settled');
+            expect(rawEmailString).toContain('transaction report for your jurisdiction for the period 2024-03-01 to');
+            expect(rawEmailString).toContain('2024-03-07.</div>');
             expect(rawEmailString).toContain('To: ohio@example.com');
         });
 
--- a/backend/compact-connect/lambdas/nodejs/tests/email-notification-service-lambda.test.ts
+++ b/backend/compact-connect/lambdas/nodejs/tests/email-notification-service-lambda.test.ts
@@ -189,7 +189,7 @@ describe('EmailNotificationServiceLambda', () => {
     });
 
     describe('Compact Transaction Report', () => {
-        const SAMPLE_TRANSACTION_REPORT_EVENT: EmailNotificationEvent = {
+        const SAMPLE_COMPACT_TRANSACTION_REPORT_EVENT: EmailNotificationEvent = {
             template: 'CompactTransactionReporting',
             recipientType: 'COMPACT_SUMMARY_REPORT',
             compact: 'aslp',
@@ -202,7 +202,7 @@ describe('EmailNotificationServiceLambda', () => {
         };
 
         it('should successfully send compact transaction report email', async () => {
-            const response = await lambda.handler(SAMPLE_TRANSACTION_REPORT_EVENT, {} as any);
+            const response = await lambda.handler(SAMPLE_COMPACT_TRANSACTION_REPORT_EVENT, {} as any);
 
             expect(response).toEqual({
                 message: 'Email message sent'
@@ -254,14 +254,14 @@ describe('EmailNotificationServiceLambda', () => {
                 }
             });
 
-            await expect(lambda.handler(SAMPLE_TRANSACTION_REPORT_EVENT, {} as any))
+            await expect(lambda.handler(SAMPLE_COMPACT_TRANSACTION_REPORT_EVENT, {} as any))
                 .rejects
                 .toThrow('No recipients found for compact aslp with recipient type COMPACT_SUMMARY_REPORT');
         });
 
         it('should throw error when required template variables are missing', async () => {
             const eventWithMissingVariables: EmailNotificationEvent = {
-                ...SAMPLE_TRANSACTION_REPORT_EVENT,
+                ...SAMPLE_COMPACT_TRANSACTION_REPORT_EVENT,
                 templateVariables: {}
             };
 
@@ -270,19 +270,19 @@ describe('EmailNotificationServiceLambda', () => {
                 .toThrow('Missing required template variables for CompactTransactionReporting template');
         });
 
-        it('should throw error when S3 fails to return report', async () => {
+        it('should throw error when S3 fails to return compact report', async () => {
             mockS3Client.on(GetObjectCommand).resolves({
                 Body: undefined
             });
 
-            await expect(lambda.handler(SAMPLE_TRANSACTION_REPORT_EVENT, {} as any))
+            await expect(lambda.handler(SAMPLE_COMPACT_TRANSACTION_REPORT_EVENT, {} as any))
                 .rejects
                 .toThrow('Failed to retrieve report from S3: compact/aslp/reports/test-report.zip');
         });
     });
 
     describe('Jurisdiction Transaction Report', () => {
-        const SAMPLE_TRANSACTION_REPORT_EVENT: EmailNotificationEvent = {
+        const SAMPLE_JURISDICTION_TRANSACTION_REPORT_EVENT: EmailNotificationEvent = {
             template: 'JurisdictionTransactionReporting',
             recipientType: 'JURISDICTION_SUMMARY_REPORT',
             compact: 'aslp',
@@ -296,7 +296,7 @@ describe('EmailNotificationServiceLambda', () => {
         };
 
         it('should successfully send jurisdiction transaction report email', async () => {
-            const response = await lambda.handler(SAMPLE_TRANSACTION_REPORT_EVENT, {} as any);
+            const response = await lambda.handler(SAMPLE_JURISDICTION_TRANSACTION_REPORT_EVENT, {} as any);
 
             expect(response).toEqual({
                 message: 'Email message sent'
@@ -348,14 +348,14 @@ describe('EmailNotificationServiceLambda', () => {
                 }
             });
 
-            await expect(lambda.handler(SAMPLE_TRANSACTION_REPORT_EVENT, {} as any))
+            await expect(lambda.handler(SAMPLE_JURISDICTION_TRANSACTION_REPORT_EVENT, {} as any))
                 .rejects
                 .toThrow('No recipients found for jurisdiction oh in compact aslp');
         });
 
         it('should throw error when required template variables are missing', async () => {
             const eventWithMissingVariables: EmailNotificationEvent = {
-                ...SAMPLE_TRANSACTION_REPORT_EVENT,
+                ...SAMPLE_JURISDICTION_TRANSACTION_REPORT_EVENT,
                 templateVariables: {}
             };
 
@@ -366,7 +366,7 @@ describe('EmailNotificationServiceLambda', () => {
 
         it('should throw error when jurisdiction is missing', async () => {
             const eventWithMissingJurisdiction: EmailNotificationEvent = {
-                ...SAMPLE_TRANSACTION_REPORT_EVENT,
+                ...SAMPLE_JURISDICTION_TRANSACTION_REPORT_EVENT,
                 jurisdiction: undefined
             };
 
@@ -375,12 +375,12 @@ describe('EmailNotificationServiceLambda', () => {
                 .toThrow('Missing required jurisdiction field for JurisdictionTransactionReporting template');
         });
 
-        it('should throw error when S3 fails to return report', async () => {
+        it('should throw error when S3 fails to return jurisdiction report', async () => {
             mockS3Client.on(GetObjectCommand).resolves({
                 Body: undefined
             });
 
-            await expect(lambda.handler(SAMPLE_TRANSACTION_REPORT_EVENT, {} as any))
+            await expect(lambda.handler(SAMPLE_JURISDICTION_TRANSACTION_REPORT_EVENT, {} as any))
                 .rejects
                 .toThrow('Failed to retrieve report from S3: compact/aslp/reports/jurisdiction-transactions/jurisdiction/oh/reporting-cycle/weekly/2024/03/07/transaction-report.zip');
         });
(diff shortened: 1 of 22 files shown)


backend/compact-connect/stacks/persistent_stack/transaction_reports_bucket.py
```diff
@@ -0,0 +1,48 @@
...
+        **kwargs,
+    ):
+        super().__init__(
+            scope,
+            construct_id,
+            encryption=BucketEncryption.KMS,
+            encryption_key=encryption_key,
+            server_access_logs_bucket=access_logs_bucket,
+            versioned=True,
+            cors=[
```
jusdino, line 30: Isn't this bucket only used internally? Why CORS?

backend/compact-connect/lambdas/python/purchases/handlers/transaction_reporting.py
```diff
@@ -11,16 +15,187 @@
...
+    gzip_buffer = BytesIO()
+    with gzip.GzipFile(fileobj=gzip_buffer, mode='wb') as gz:
+        gz.write(transaction_detail.encode('utf-8'))
+    s3_client.put_object(Bucket=bucket_name, Key=paths['transaction_detail_gz'], Body=gzip_buffer.getvalue())
+
+    # Create and store zip with uncompressed CSV
+    zip_buffer = BytesIO()
+    with ZipFile(zip_buffer, 'w', compression=ZIP_DEFLATED) as zip_file:
+        zip_file.writestr(f'{jurisdiction}-transaction-detail-{date_range}.csv', transaction_detail.encode('utf-8'))
+    s3_client.put_object(Bucket=bucket_name, Key=paths['report_zip'], Body=zip_buffer.getvalue())
```
jusdino, line 182: Are we planning to store these records forever, currently?

jusdino (changes requested):
Looks really good! A few questions:

backend/compact-connect/lambdas/python/purchases/handlers/transaction_reporting.py
```diff
@@ -11,16 +15,154 @@
...
+    :return: Dictionary of file types to their S3 paths
+    """
+    date_range = f"{start_time.strftime('%Y-%m-%d')}--{end_time.strftime('%Y-%m-%d')}"
+    base_path = (
+        f"compact/{compact}/reports/compact-transactions/reporting-cycle/{reporting_cycle}/"
+        f"{end_time.strftime('%Y/%m/%d')}"
+    )
+
+    # Define paths for all report files
+    paths = {
```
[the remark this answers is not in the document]
landonshumway-ia: Good call out, the gz versions are stored for future use, as @isabeleliassen mentioned there is a possibility that admins will want to be able to download previously generated reports. I'll add a comment to capture this.

backend/compact-connect/lambdas/python/purchases/handlers/transaction_reporting.py
landonshumway-ia: Until we have pick up the ticket for creating an archive for all the system data, it makes sense to just hold onto it for now, but I'm open to discuss adding a lifecycle policy.

jusdino: I'm good with leaving it if you think it will be useful - we'll just want to flag it somewhere as a data store that will only ever grow, so we have a plan to manage it.

landonshumway-ia: I'll add a TODO to the CDK construct that we will add a lifecycle policy when we get into archive work

backend/compact-connect/lambdas/nodejs/tests/email-notification-service-lambda.test.ts
```diff
@@ -322,5 +374,15 @@ describe('EmailNotificationServiceLambda', () => {
                 .rejects
                 .toThrow('Missing required jurisdiction field for JurisdictionTransactionReporting template');
         });
+
+        it('should throw error when S3 fails to return report', async () => {
```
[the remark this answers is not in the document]
landonshumway-ia: not quite, one tests the error through the compact transaction reporting method, the other through the jurisdiction reporting method.

jusdino (changes requested):
Nearly there...

backend/compact-connect/lambdas/python/purchases/tests/function/test_handlers/test_transaction_reporting.py
```diff
@@ -520,3 +806,230 @@ def test_generate_report_handles_unknown_jurisdiction(self, mock_lambda_client):
...
+        self._add_mock_transaction_to_db(
+            jurisdictions=['oh'],
+            licensee_id=mock_user['providerId'],
+            month_iso_string='2024-03',
+            transaction_settlement_time_utc=datetime.fromisoformat('2024-03-01T00:00:00+00:00'),
+        )
+
+        # Calculate expected date range
+        # the end time should be the last day of the month
+        end_time = datetime.fromisoformat('2024-02-29T23:59:59:9999+00:00')
```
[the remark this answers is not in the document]
landonshumway-ia: That value was just for display and filename purposes. I refactored the test variables to make this easier to read.

jusdino (approved):
Great! @jlkravitz , this is ready for you.

backend/compact-connect/lambdas/nodejs/tests/email-notification-service-lambda.test.ts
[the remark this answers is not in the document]
jlkravitz: I'm a bit confused on this one- the description and code for both tests look identical (except for the `toThrow` check). What distinguishes them? Can we clarify?

backend/compact-connect/lambdas/python/purchases/handlers/transaction_history.py
```diff
@@ -32,8 +32,8 @@ def process_settled_transactions(event: dict, context: LambdaContext) -> dict:
     current_batch_id = event.get('currentBatchId')
     processed_batch_ids = event.get('processedBatchIds', [])
 
-    # Calculate time range for the last 24 hours
-    end_time = config.current_standard_datetime
+    # This lambda is triggered at noon UTC-4, so we calculate the time range for the last 24 hours
+    end_time = config.current_standard_datetime.replace(hour=16, minute=0, second=0, microsecond=0)
```
jlkravitz, line 36: Any reason to tie this so strongly to the lambda runtime? Can't we just subtract 24 hours?

If it needs to stay as-is, mind adding a quick note in parens like "at noon UTC-4 (i.e., 16:00 UTC)" to make the math a little clearer?

backend/compact-connect/lambdas/python/purchases/handlers/transaction_reporting.py
```diff
@@ -11,16 +15,185 @@
...
+def _get_display_date_range(reporting_cycle: str) -> tuple[datetime, datetime]:
+    """Get the display date range for reports.
+
+    These dates are used for report filenames and email notifications.
+
+    :param reporting_cycle: Either 'weekly' or 'monthly'
+    :return: Tuple of (start_time, end_time) in UTC for display purposes
+    """
+    if reporting_cycle == 'weekly':
+        # Reports run on Friday 10:00 PM UTC
```
jlkravitz, line 27: This code seems agnostic to when the reports run. If that's the case, I think we can remove this comment.

backend/compact-connect/lambdas/python/purchases/handlers/transaction_reporting.py
```diff
@@ -11,16 +15,185 @@
...
+        end_time = config.current_standard_datetime
+        # Go back 7 days to capture the full week
+        start_time = end_time - timedelta(days=7)
+        return start_time, end_time
+    if reporting_cycle == 'monthly':
+        # Reports run shortly after midnight on the first day of the month
```
jlkravitz, line 33: This code seems agnostic to when the reports run, as long as it is in the month _after_ the target report month. If that's the case, I think we can update this comment to focus on this precondition.

backend/compact-connect/lambdas/python/purchases/handlers/transaction_reporting.py
```diff
@@ -11,16 +15,185 @@
...
+        return start_time, end_time
+    raise ValueError(f'Invalid reporting cycle: {reporting_cycle}')
+
+
+def _get_query_date_range(reporting_cycle: str) -> tuple[datetime, datetime]:
+    """Get the query date range for DynamoDB queries.
+
+    Our Sort Key format for transactions includes additional components after the timestamp
+    (COMPACT#name#TIME#timestamp#BATCH#id#TX#id), So the DynamoDB BETWEEN condition is INCLUSIVE for the beginning
+    range and EXCLUSIVE at the end range. We need to adjust our timestamps accordingly to ensure we capture all
```
jlkravitz, line 49: I'm not entirely sure how there being addl components after the timestamp in the SK implies that the BETWEEN condition is INCLUSIVE/EXCLUSIVE in the manner you describe. Can you clarify?

backend/compact-connect/lambdas/python/purchases/handlers/transaction_reporting.py
```diff
@@ -11,16 +15,185 @@
...
+    settled transactions exactly once.
+
+    :param reporting_cycle: Either 'weekly' or 'monthly'
+    :return: Tuple of (start_time, end_time) in UTC for DynamoDB queries
+    """
+    if reporting_cycle == 'weekly':
+        # Reports run on Friday 10:00 PM UTC
+        end_time = config.current_standard_datetime.replace(hour=22, minute=0, second=0, microsecond=0)
```
jlkravitz, line 57: Ah, I guess in this case, to ensure the collection of weekly reports is 'complete', you _do_ want there to be a strong dependency on the runtime of the report. True?

backend/compact-connect/lambdas/python/purchases/handlers/transaction_reporting.py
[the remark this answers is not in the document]
jlkravitz: Couldn't admins still download the zip files? What's the intent behind storing the files twice?

backend/compact-connect/lambdas/python/purchases/handlers/transaction_reporting.py
```diff
@@ -11,16 +15,154 @@
...
+    :return: Dictionary of file types to their S3 paths
+    """
+    date_range = f"{start_time.strftime('%Y-%m-%d')}--{end_time.strftime('%Y-%m-%d')}"
+    base_path = (
+        f"compact/{compact}/reports/jurisdiction-transactions/jurisdiction/{jurisdiction}/"
+        f"reporting-cycle/{reporting_cycle}/{end_time.strftime('%Y/%m/%d')}"
+    )
+
+    # Define paths for all report files
+    paths = {
```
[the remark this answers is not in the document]
jlkravitz: Same question as above.

backend/compact-connect/stacks/reporting_stack.py
```diff
@@ -175,11 +178,26 @@ def _add_transaction_reporting_chain(self, persistent_stack: ps.PersistentStack)
             Rule(
                 self,
                 f'{compact.capitalize()}-WeeklyTransactionReportRule',
-                schedule=Schedule.cron(week_day='7', hour='1', minute='0', month='*', year='*'),
+                # Send weekly reports every Friday at 10:00 PM UTC
+                schedule=Schedule.cron(week_day='FRI', hour='22', minute='0', month='*', year='*'),
```
jlkravitz, line 182: Why was the choice made to send the report at 16 UTC rather than shortly after midnight (as we do for the weekly reports)? Is this something we can document here and/or in a README?

jlkravitz (commented):
Generally looks good. A couple questions/comments!

backend/compact-connect/lambdas/python/purchases/handlers/transaction_history.py
landonshumway-ia: We need a fixed time frame, so that we ensure that the same time block is always collected and none of the batches slip through the cracks. If we just go purely off of what the exact time is minus 24 hours, you run the risk of a gap where a batch settled within a small time window that would not be included by the collector (ie event bridge performance is degraded and it takes a couple minutes to trigger the lambda, so you have a couple minute gap where a batch might have been settled)

I will update the comment though to make this clearer.

backend/compact-connect/lambdas/python/purchases/handlers/transaction_reporting.py
landonshumway-ia: I will update the comment to explain why this is the case

backend/compact-connect/lambdas/python/purchases/handlers/transaction_reporting.py
landonshumway-ia: It is not, since again we want a fixed time window that we report on to make sure that no transactions slip through any gaps. I will update the comment to make this clearer.

landonshumway-ia: Actually, you're correct on this line, since this particular method is just for getting the display date. It is the query time range which is dependent on when the lambda is triggered. I will remove the comment for this line here.

backend/compact-connect/lambdas/python/purchases/handlers/transaction_reporting.py
landonshumway-ia: I've updated the comment to clarify this.

backend/compact-connect/lambdas/python/purchases/handlers/transaction_reporting.py
landonshumway-ia: Yes, for when we calculate the query range, we want to make the the time window is fixed so there are no gaps.

backend/compact-connect/stacks/reporting_stack.py
landonshumway-ia: I'll answer both this and your question above about the zip files, since I can't comment on those items for some reason. For the .gz files, the intent was that compact admins may want to download just one report but not both (in the case of jurisdiction admins, there is only one report, so I would agree it makes sense to just have the one zip file, I'll make that change)

In terms of why we send the weekly report at this time, this would be Friday afternoon when admins would receive these reports at the end of the work week, so it made sense to send them around that time, rather than after midnight. However, I will clarify with @isabeleliassen if there is a specific time we should be sending both the weekly and monthly reports.

landonshumway-ia: As I walk through this though, it may be best to just keep this logic simple and only store the zip files with both reports inside, and if a compact admin wants a report for that period they will get all reports in the zip. I will simplify this aspect of the system.

backend/compact-connect/lambdas/python/purchases/handlers/transaction_reporting.py
jlkravitz: Thanks! I guess my question still stands - is it not worth replacing the day to 1, since this function doesn't _need_ to run on the first day of the month?

```python
end_time = config.current_standard_datetime.replace(day=1, hour=0, minute=0, second=0, microsecond=0) - timedelta(
            days=1
        )
```

landonshumway-ia: Sure, that makes sense. I'll make that update.

jlkravitz (approved):
Looks great! Thanks for clarifying the tests. @isabeleliassen Good to merge.

isabeleliassen:
@jlkravitz still some unresolved conversations I think!

landonshumway-ia:
@isabeleliassen those have now been resolved. This should be ready to go. Thanks.

isabeleliassen (approved):

PR #510, opened 2025-01-30
Feat/privilege numbers

### LINKED ISSUE #436: reduce license types for counseling to say just licensed professional counselor

### LINKED ISSUE #466: privilege number assigned to each privilege

jusdino:
### Description List
- Assign a new privilege id on privilege purchase
- Reuse privilege id for re-issue of existing privilege
- Update API docs
- Leverage new logger context manager feature

### Breaking change
With this PR, privileges will all be expected to include `privilegeId` in their schema, so any existing privileges (I think there are a total of 0 in CSG environments) will need updating to include one or just be deleted.

### Testing List
- Code review

Closes #466 
Closes #436


backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
```diff
@@ -613,3 +630,27 @@ def batch_get_providers_by_id(self, compact: str, provider_ids: list[str]) -> li
...
+                '#count': 'privilegeCount',
+            },
+            ExpressionAttributeValues={
+                ':increment': 1,
+            },
+            ReturnValues='UPDATED_NEW',
+        )
+        privilege_count = resp['Attributes']['privilegeCount']
+        logger.info('Claimed privilege id', compact=compact, privilege_count=privilege_count)
+        return privilege_count
```
landonshumway-ia: clever, I like it 👍

backend/compact-connect/lambdas/python/common/tests/__init__.py
```diff
@@ -30,7 +30,12 @@ def setUpClass(cls):
                 'COMPACTS': '["aslp", "octp", "coun"]',
                 'JURISDICTIONS': '["ne", "oh", "ky"]',
                 'LICENSE_TYPES': json.dumps(
-                    {'aslp': ['audiologist', 'speech-language pathologist', 'speech and language pathologist']},
+                    {
+                        'aslp': [
+                            {'name': 'audiologist', 'abbreviation': 'AUD'},
+                            {'name': 'speech-language pathologist', 'abbreviation': 'SLP'},
```
landonshumway-ia, line 36: This doesn't quite match up with the cdk.json, which has the abbreviations lower-cased.

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
```diff
@@ -30,45 +31,48 @@ def __init__(self, config: _Config):
...
+            try:
+                resp = self.config.ssn_table.get_item(
+                    Key={'pk': f'{compact}#SSN#{ssn}', 'sk': f'{compact}#SSN#{ssn}'},
+                    ConsistentRead=True,
+                )['Item']
+            except KeyError as e:
+                logger.info('Provider not found by SSN', exc_info=e)
+                raise CCNotFoundException('No licensee found by that identifier') from e
+
+            return resp['providerId']
```
landonshumway-ia: This is interesting that we need to use a context manager for this, given that the value of 'compact' is generally consistent for the lifecycle of the lambda invocation that would be calling these methods (ie you can't query for multiple compacts in a single call) I'm ok with it if gives us clear logging. I'm just wondering if there is another way to handle this with the context syntax overhead, like some way to set the marker that persists for the lifetime of the logger.

landonshumway-ia (changes requested):
Looks really good, just one small thing I noticed in the test data setup and one question.

Also, this is a breaking change right? We will want to note that in the PR description so @jlkravitz is aware

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
jusdino: I added a little flexibility to the decorator, so we can reduce the cases where we have to indent a whole block. What do you think?

landonshumway-ia: It's good. I like how we can explicitly call out what args we want to log, without having to change the formatting of the underlying method code. 👍

landonshumway-ia (approved):

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
```diff
@@ -249,20 +263,51 @@ def _generate_privilege_record(
...
+        else:
+            date_of_issuance = current_datetime
+            # Claim a privilege number for this jurisdiction
+            # Note that this number claim is not rolled back on failure, which can result in gaps
+            # in the privilege numbers. Having gaps in the privilege numbers was deemed acceptable
+            # for exceptional circumstances like errors in this flow.
+            privilege_number = self.claim_privilege_number(compact=compact)
+            logger.info('Claimed a new privilege number', privilege_number=privilege_number)
+            privilege_id = '-'.join(
+                (license_type_abbreviation.upper(), jurisdiction_postal_abbreviation.upper(), str(privilege_number))
```
jlkravitz, line 295: I assume this was discussed, but my understanding is that the privilege ID is visible to users. Normally, I'd veer away from sequential IDs to avoid giving users any sense of the scale of data in the database and / or to avoid easy lookups. The latter – I think – is irrelevant since we do not have an API for looking up a privilege by ID. The former may or may not be a practical issue.

jlkravitz (commented):
Looks great! Love the new decorator- very clean. Thanks for the thorough comments.

I had one question below on the privilege ID.

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
jusdino: Yep, we talked about it. This feature is directly from a request from compact EDs. They want privilege ids that have a sequential number plus the license type and jurisdiction. We are likely to also add a query-by-priv-number feature later, as well.

jlkravitz: Got it- And fair to say that someone more easily being able to figure out the number of privileges purchased for a specific license type in a specific jurisdiction is not a significant risk?

jusdino: No, I think that's ok. In this case, it's actually the number of privileges across a whole compact, though. The counter is not type/jurisdiction specific.

jlkravitz (approved):

isabeleliassen (approved):

PR #525, opened 2025-02-06
Feat/support licensee transaction fees

### LINKED ISSUE #502: Support passing transaction fees to practitioner when purchasing privileges
The ASLP and OT compacts want practitioners to pay for the transaction fees charged by the Compact's Merchant Service Provider (Authorize.net's MSP charges 2.9% + 30 cents per transaction). We need to support getting this transaction fee rate to the frontend so it can display it to the user, and the backend needs to add this as a line item when processing transactions in Authorize.net.

This ticket involves updating the backend to support this feature through our compact configuration, as well as our privilege purchasing endpoint to start storing these extra fees.

landonshumway-ia:
Several compacts have determined to charge credit card transaction fees to licensees, in order to absorb the costs of the fees charged by their Merchant Service Providers. This adds the needed compact configuration fields which compacts will need to specify when they onboard into the compact connect system.

### Requirements List
- This change is backwards compatible, as these are optional fields that a compact may or may not choose to specify.

### Description List
- Added a `transactionFeeConfiguration` field to the compact configuration schema, which includes config for licensee transaction fees.
- Returning these values in the API when the frontend gets the list of compact configuration. 
- Updated the privilege purchase flow to include these transaction fee charges in a separate line item.

### Testing List
- uint tests to verify licensee transaction fees are added
- Code review

Closes #502

### DIFF
--- a/backend/compact-connect/lambdas/python/purchases/tests/unit/test_purchase_client.py
+++ b/backend/compact-connect/lambdas/python/purchases/tests/unit/test_purchase_client.py
@@ -100,17 +100,17 @@ def _generate_aslp_compact_configuration(include_licensee_charges: bool = False)
         return Compact(compact)
 
 
-def _generate_selected_jurisdictions(jursidction_items: list[dict] = None):
+def _generate_selected_jurisdictions(jurisdiction_items: list[dict] = None):
     from cc_common.data_model.schema.jurisdiction import Jurisdiction
 
-    if jursidction_items is None:
-        jursidction_items = [
+    if jurisdiction_items is None:
+        jurisdiction_items = [
             {'postalCode': 'oh', 'jurisdictionName': 'ohio', 'jurisdictionFee': 100.00},
         ]
 
     jurisdiction_configurations = []
 
-    for jurisdiction_test_item in jursidction_items:
+    for jurisdiction_test_item in jurisdiction_items:
         with open('../common/tests/resources/dynamo/jurisdiction.json') as f:
             jurisdiction = json.load(f)
             jurisdiction['jurisdictionFee'] = Decimal(jurisdiction_test_item['jurisdictionFee'])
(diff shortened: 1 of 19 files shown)


backend/compact-connect/lambdas/python/purchases/tests/unit/test_purchase_client.py
```diff
@@ -303,6 +322,79 @@ def test_purchase_client_sends_expected_line_items_when_purchasing_privileges_wi
...
+            user_active_military=False,
+        )
+
+        call_args = mock_create_transaction_controller.call_args.args
+        api_contract_v1_obj = call_args[0]
+
+        # we check every line item of the object to ensure that the correct values are being set
+        self.assertEqual(4, len(api_contract_v1_obj.transactionRequest.lineItems.lineItem))
+        # first line item is the jurisdiction fee
+        self.assertEqual('aslp-oh', api_contract_v1_obj.transactionRequest.lineItems.lineItem[0].itemId)
```
jusdino, line 357: With this many assertions on the same structure, should we make it more readable and just assert on the whole dict shape (minus a field or two, if needed)?

jusdino (approved):
Looks great! Only one very minor nit, that I won't hold this up for:

jusdino:
@jlkravitz, this is ready for you!

backend/compact-connect/lambdas/python/purchases/tests/unit/test_purchase_client.py
landonshumway-ia: Unfortunately, this transactionRequest object is not a dict, it is a type unique to the Authorize.net SDK, which is incredibly nasty to work with since under the hood it is converting this object to XML. We have to access each attribute of this object using the . operator, we can't dump it to a json object.

landonshumway-ia: If you feel it is worth the investment, we might be able to find a way to convert it into a JSON object and make assertions on it. I've already added test setup code to convert a JSON object into a magic mock, but it seems like going from authorize.net's object to JSON might be a little more tricky.

jusdino: No, that seems like more work than value.

jlkravitz (commented):
One nit but otherwise looks great!

jlkravitz (approved):

jlkravitz (approved):
@isabeleliassen Good to merge when tests pass.

isabeleliassen (approved):

Issue #425, opened 2025-01-06
Make 'attestations' field required when purchasing privileges


landonshumway-ia:
In an effort to keep the application able to be demo'd at all times, we previously implemented a feature to the privilege purchasing endpoint to validate a list of attestations, but left the 'attestations' field in the post body as optional. This ticket includes ensuring all the known attestations are in the DB (i.e. military attestations, jurisprudence-confirmation, etc.) as well as enforcing that the 'attestation' field is present with all of the required attestations when calling the privilege purchase endpoint.

landonshumway-ia:
This ticket can be completed after #335

PR #534, opened 2025-02-07
Feat/tie purchase flow

### LINKED ISSUE #302: connect screens of privilege purchase workflow

ChiefStief:
### Requirements List
- Ideally test against an API that requires the attestations be passed up to confirm that we are meeting criteria there. My sandbox is set up to do this or we can get https://github.com/csg-org/CompactConnect/issues/425 merged into whatever environment we need

### Description List
- Refactored purchase flow to be in one page with subcomponents to support privilege purchase nav highlighting and most easily support total control of purchase flow and dynamic progress calculation
- Renamed purchase flow pages to be more easily found: `PrivilegePurchase*`
- Abstracted out purchase flow into `PurchaseFlowState`, `PurchaseFlowStep` and step level logic to better support step handling (find next step, clean state appropriately, redirect appropriately etc)
- Added `AcceptedAttestationToSend` model
- Made military affiliation attestation only show and send if user is military affiliated
- Added sending of attestations accepted to purchase API call and relevant serializer
- Improved test coverage
- Added `ProgressBar` Component
- Fixed safari display bug on SelectPrivileges screen

### Testing List
- Code review
- Try going through the privilege purchase flow process and confirm that the happy path works
- Now try navigating around the site, and with the back and cancel buttons and confirm that the form flow works as expected:

1. Back should always return to the previous flow step
2. Cancel should clear the flow entirely and bring the user to `LicenseeDashboard` screen
3. Navigating out of the flow while still in the app will retain the users progress (including privilege purchase successful), coming back to the flow should bring them back to the screen they left off at
4. Refreshing the page at any point or navigating with the url bar will clear the flow
5. Clicking finish on the Successful screen will clear the privilege flow and re-fetch the users updated privileges

Closes #302 

Note: The behavior above is what I decided to implement to avoid complexity surrounding auto filling potentially changed form fields due to the server attestations being updated on the server, complexity surrounding form race conditions and caching user responses vs updated attestations.

I also opted not to use localStorage to have this data persist across sessions. What I did do was set up the form flow so that any of these alterations can be made more easily if it turns out they are desired.


webroot/src/components/ProgressBar/ProgressBar.ts
```diff
@@ -0,0 +1,40 @@
...
+    // Lifecycle
+    //
+
+    //
+    // Computed
+    //
+
+    //
+    // Methods
+    //
```
jsandoval81, line 35: Minor: remove

webroot/src/components/ProgressBar/ProgressBar.less
```diff
@@ -0,0 +1,21 @@
...
+//
+.progress-bar {
+    display: flex;
+    flex-direction: row;
+    width: 100%;
+
+    .finished-portion {
+        height: 10px;
+        background-color: @primaryColor;
+    }
```
jsandoval81, line 15: How would you feel about adding something like ~`transition: width 1s ease-in-out;`?

webroot/src/models/AcceptedAttestationToSend/AcceptedAttestationToSend.model.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+    it('should create a PrivilegePurchaseOption with expected defaults', () => {
+        const acceptedAttestationToSend = new AcceptedAttestationToSend();
+
+        expect(acceptedAttestationToSend).to.be.an.instanceof(AcceptedAttestationToSend);
+        expect(acceptedAttestationToSend.attestationId).to.equal(null);
+        expect(acceptedAttestationToSend.version).to.equal(null);
+    });
+    it('should create a AcceptedAttestationToSend with specific values', () => {
+        const data = {
+            attestationId: 'attest-attaestation',
```
jsandoval81, line 26: Minor: typo

webroot/src/pages/PrivilegePurchase/PrivilegePurchase.ts
```diff
@@ -0,0 +1,131 @@
...
+        return Math.round(((this.currentFlowStep + 1) / this.numFlowSteps) * 100);
+    }
+
+    //
+    // Methods
+    //
+    handlePurchaseFlowState() {
+        const { $store, $router, currentFlowStep } = this;
+
+        const nextStepNeeded = $store.getters['user/getNextNeededPurchaseFlowStep']();
```
jsandoval81, line 113: Minor: remove empty line

webroot/src/store/user/user.state.ts
```diff
@@ -37,10 +36,10 @@ export const state: State = {
...
     isLoadingAccount: false,
     isLoadingPrivilegePurchaseOptions: false,
-    arePurchaseAttestationsAccepted: false,
-    selectedPrivilegesToPurchase: null,
     refreshTokenTimeoutId: null,
     currentCompact: null,
-    purchase: {},
+    purchase: new PurchaseFlowState({
+        steps: []
+    }),
```
jsandoval81, line 43: Minor: Can we rely on the default `steps` value, rather then setting it here?

webroot/src/components/Page/PageMainNav/PageMainNav.ts
```diff
@@ -175,7 +175,7 @@ class PageMainNav extends Vue {
                 isExactActive: false,
             },
             {
-                to: 'PrivilegePurchaseInformationConfirmation',
+                to: 'PrivilegePurchase',
```
jsandoval81, line 178: Should this now have a conditional to be disabled (from view or from click) if purchase is not available? The same concept that controls the click-ability of the obtain-privileges button on the dashboard page.

webroot/src/components/PrivilegePurchaseAttestation/PrivilegePurchaseAttestation.ts
```diff
@@ -150,18 +176,22 @@ export default class PrivilegePurchaseAttestation extends mixins(MixinForm) {
...
-                value: false,
-            }),
             submit: new FormInput({
                 isSubmitInput: true,
                 id: 'submit',
             }),
         });
+
+        if (this.isMilitaryAffiliated) {
+            this.formData.militaryAffiliation = new FormInput({
```
jsandoval81, line 186: Not that this is conditional, the mock-populate function will also need to same conditional. Currently mock-populate errors if the military attestation is not needed.

On the subject of mock-populate: how would you feel about adding the to each step component where it's missing?

webroot/src/components/PrivilegePurchaseAttestation/PrivilegePurchaseAttestation.vue
```diff
@@ -9,24 +9,27 @@
...
-                </div>
-                <div v-else>
-                    <MockPopulate :isEnabled="isMockPopulateEnabled" @selected="mockPopulate" />
-                    <InputRadioGroup :formInput="formData.investigations" />
-                    <div class="form-section-title">{{ $t('licensing.discipline') }} *</div>
-                    <InputCheckbox :formInput="formData.disciplineCurrent" />
-                    <InputCheckbox :formInput="formData.disciplinePrior" />
-                    <div class="form-section-title form-section-gap">
-                        {{ $t('licensing.provisionOfInformation') }} *
+                <ProgressBar :progressPercent="progressPercent" />
```
jsandoval81, line 12: Would it make more sense to have the `ProgressBar` component just exist on the `PrivilegePurchase` page template?

I'm not sure if that would make it easier or harder to manage. It might be easier since I notice the bar currently has a little oddity the further along in the flow it gets (e.g. progress looks further at attestations than at payment - images below). Since each step can have a different width, options might be:

1. The parent page has a container with a width that the child components just `width: 100%` inherit, _or_
2. The progress bar could be more de-coupled from the step components, king of floating above them centered or something, _or_
3. Keep the progress bar as part of each component but just fix the finished width to more closely represent the overall progress.
4. Other ideas?

Screenshot 2025-02-10 at 10 38 26 AM
Screenshot 2025-02-10 at 10 38 45 AM

jsandoval81 (changes requested):
This is excellent 🚀 I really like the purchase-flow-state mechanism.

webroot/src/components/PrivilegePurchaseAttestation/PrivilegePurchaseAttestation.vue
ChiefStief: I considered having it as part of the top level page due to the fact that it is a function of that top level data and could be abstracted out of each page's logic.

I ultimately chose not to because it was more difficult to abstract out the display / layout of the progress bar:
1) The progress bar does not go backwards in that page it's max width is just only the white part because the blue on blue would not really work. I actually deviated from the designs here, it seems like this is confusing and maybe I should have followed the designs exactly.
2) Besides that screen, the form input containers have different amounts of padding. It could certainly be argued that these padding values should be standardized (I believe they actually are in the designs) and combined with the above change we could have the ProgressBar coming from the higher level page

ChiefStief: What's your opinion of the blue on blue progress bar? If you feel like thats fine, reverting to that and standardizing the padding + moving the ProgressBar up into the top level page probably makes the most sense

jsandoval81: If design is good with the blue-on-blue and it's been approved, then I say that's the probably best. That at least would solve the issue from the screenshots.

If there is some value in cleaning up the padding + moving the ProgressBar, then that sounds good too. Whatever options leave us with meeting-the-spec + easiest-to-maintain.

webroot/src/components/Page/PageMainNav/PageMainNav.ts
jsandoval81: For the general "should we allow purchase" criteria, we can confirm that with the product team. For this specific nav item, it should just (eventually) match whatever the Obtain Privileges button is doing. 

It's fine if that's in the cleanup ticket and not in this PR, as long as we've got it documented over there.

jsandoval81 (approved):

jsandoval81:

webroot/src/store/user/user.mutations.ts
```diff
@@ -204,4 +182,13 @@ export default {
         state.isLoadingAccount = false;
         state.error = error;
     },
+    [MutationTypes.CLEAN_PURCHASE_FLOW_STATE]: (state: any, flowStepNum: number) => {
```
jlkravitz, line 185: nit: The name `CLEAN_PURCHASE_FLOW_STATE` confused me throughout the codebase, as I would've expected it to reset the entire flow. That made the unnamed parameter to the mutation confusing where it's called in other files (`flowStepNum`).

Might a name like 'RESET_TO_PURCHASE_FLOW_STEP` (or similar) be clearer? If you agree, could we change these for the model, mutation, etc. (wherever the `CLEAN_PURCHASE_FLOW_STATE` pops up).

webroot/src/models/PurchaseFlowState/PurchaseFlowState.model.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+describe('PurchaseFlowState model', () => {
+    it('should create a PrivilegePurchaseOption with expected defaults', () => {
+        const purchaseFlowState = new PurchaseFlowState();
+
+        expect(purchaseFlowState).to.be.an.instanceof(PurchaseFlowState);
+        expect(purchaseFlowState.steps).to.matchPattern([]);
+    });
+    it('should create a PurchaseFlowState with specific values', () => {
+        const data = {
+            steps: [ new PurchaseFlowStep()]
```
jlkravitz, line 27: extraneous space

jlkravitz (commented):
Looks good! Few questions/comments.

jlkravitz (approved):
Looks great! @isabeleliassen Good to merge.

isabeleliassen (approved):

PR #535, opened 2025-02-10
Feat/get one ssn

### LINKED ISSUE #392: Add new get-one full SSN endpoint (backend)
- Add a new permission node that is specific to SSN access, similar to readPrivate with compact-wide access

jusdino:
### Description List
- Added a GET .../providers/:providerId/ssn endpoint to retrieve a provider's SSN
- Added a new readSSN permissions action for compact/jurisdiction-level access to SSNs
- Added basic metrics/Alarms for monitoring activity on this sensitive endpoint

Closes #392


backend/compact-connect/stacks/persistent_stack/__init__.py
```diff
@@ -194,6 +195,17 @@ def _add_data_resources(self, removal_policy: RemovalPolicy):
         )
 
         self.ssn_table = SSNTable(self, 'SSNTable', removal_policy=removal_policy)
+        # Run migration 392 to add a GSI pk field to the SSN records
+        DataMigration(
+            self,
+            '392SSNMigration',
+            migration_dir='392_ssn',
```
landonshumway-ia, line 211: Can we add a comment here that the 392 is associated with the ticket number? It took me some time to draw the connection.

backend/compact-connect/tests/app/test_api/test_query_providers_api.py
```diff
@@ -233,3 +234,41 @@ def test_synth_generates_get_provider_ssn_endpoint(self):
             'GET_PROVIDER_SSN_RESPONSE_SCHEMA',
             overwrite_snapshot=False,
         )
+
+    def test_synth_generates_get_provider_ssn_alarm(self):
```
landonshumway-ia, line 238: Great test to check for 👍

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
```diff
@@ -50,6 +50,46 @@ def get_provider_id(self, *, compact: str, ssn: str) -> str:
...
+    def get_ssn_by_provider_id(self, *, compact: str, provider_id: str) -> str:
+        logger.info('Getting ssn by provider id', compact=compact, provider_id=provider_id)
+        resp = self.config.ssn_table.query(
+            KeyConditionExpression=Key('providerIdGSIpk').eq(f'{compact}#PROVIDER#{provider_id}'),
+            IndexName=self.config.ssn_index_name,
+        )['Items']
+        if len(resp) == 0:
+            raise CCNotFoundException('Provider not found')
+        if len(resp) != 1:
+            raise CCInternalException(f'Expected 1 SSN index record, got {len(resp)}')
```
landonshumway-ia, line 90: We'll have to figure out how this works in the case where a provider's SSN was mistyped on a license and then later adjusted, but I would imagine we would just be making an in place update on the existing SSN record which matches this particular GSI... not a problem we need to solve right now, but just wanted to call that out here.

backend/compact-connect/lambdas/python/provider-data-v1/handlers/providers.py
```diff
@@ -116,8 +119,40 @@ def get_provider(event: dict, context: LambdaContext):  # noqa: ARG001 unused-ar
...
-    return sanitize_provider_data_based_on_caller_scopes(
-        compact=compact, provider=provider_information, scopes=get_event_scopes(event)
-    )
+        return sanitize_provider_data_based_on_caller_scopes(
+            compact=compact, provider=provider_information, scopes=get_event_scopes(event)
+        )
+
+
+@api_handler
+@authorize_compact(action='readSSN')
```
landonshumway-ia, line 131: nit - It would be nice if we could move 'readSSN' into an enum, rather than hardcoding it here.

backend/compact-connect/stacks/api_stack/v1_api/query_providers.py
```diff
@@ -162,23 +171,25 @@ def _query_providers_handler(
         ssn_table: SSNTable,
         lambda_environment: dict,
     ) -> PythonFunction:
-        handler = PythonFunction(
+        self.query_providers_handler = PythonFunction(
             self.resource,
             'QueryProvidersHandler',
             description='Query providers handler',
+            # We will not associate with this role, once we remove SSN queries
```
landonshumway-ia, line 186: Can we mark this as a TODO, so that it's harder to miss?

backend/compact-connect/stacks/api_stack/v1_api/query_providers.py
```diff
@@ -192,4 +203,76 @@ def _query_providers_handler(
...
+            description='Get provider SSN handler',
+            lambda_dir='provider-data-v1',
+            index=os.path.join('handlers', 'providers.py'),
+            handler='get_provider_ssn',
+            environment=lambda_environment,
+            alarm_topic=self.api.alarm_topic,
+        )
+        # The lambda needs to read providers from the provider table and the SSN from the ssn table
+        ssn_table.key.grant_decrypt(self.get_provider_ssn_handler)
+        ssn_table.grant_read_data(self.get_provider_ssn_handler)
```
landonshumway-ia: I feel like this is a case where we should explicitly list the API calls this lambda is allowed to make, rather than using the 'grant_read_data' method, which of course grants more than is needed. Even though you are restricting it on the resource policy, it's still nice to restrict it on this side as well.

backend/compact-connect/lambdas/python/common/cc_common/utils.py
```diff
@@ -468,10 +470,24 @@ def get_sub_from_user_attributes(attributes: list):
...
-    if f'{compact}/{compact}.readPrivate' in scopes:
+def _user_has_read_private_access_for_provider(compact: str, provider_information: dict, scopes: set[str]) -> bool:
+    return _user_has_permission_for_action_on_user(
+        action='readPrivate', compact=compact, provider_information=provider_information, scopes=scopes
+    )
+
+
+def user_has_read_ssn_access_for_provider(compact: str, provider_information: dict, scopes: set[str]) -> bool:
+    return _user_has_permission_for_action_on_user(
+        action='readSSN', compact=compact, provider_information=provider_information, scopes=scopes
```
landonshumway-ia, line 481: ditto on using an enum for the 'readSSN' permission, makes it much easier to track where we are using the permission

backend/compact-connect/lambdas/python/staff-user-pre-token/user_data.py
```diff
@@ -60,6 +60,10 @@ def _process_compact_permissions(self, compact_name, compact_permissions):
             # This action only has one level of authz, since there is no external scope for it
             self.scopes.add(f'{compact_name}/{compact_name}.readPrivate')
 
+        if 'readSSN' in compact_actions:
```
landonshumway-ia, line 63: Honestly, it would be good to have an enum for all of these permission actions now that the list keeps growing. But I won't hold up the PR if you just want to stick with all the hardcoded values for now.

backend/compact-connect/stacks/api_stack/v1_api/query_providers.py
```diff
@@ -247,32 +303,22 @@ def _get_provider_ssn_handler(
...
+            role=ssn_table.api_query_role,
             environment=lambda_environment,
             alarm_topic=self.api.alarm_topic,
         )
         # The lambda needs to read providers from the provider table and the SSN from the ssn table
-        ssn_table.key.grant_decrypt(self.get_provider_ssn_handler)
-        ssn_table.grant_read_data(self.get_provider_ssn_handler)
+        # Through, ssn table access is granted via resource policies on the table and key
+        # ssn_table.key.grant_decrypt(self.get_provider_ssn_handler)
+        # ssn_table.grant_read_data(self.get_provider_ssn_handler)
```
landonshumway-ia, line 320: any reason to leave this in a commented out state, rather than just removing them completely?

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
jusdino: Yeah, correcting that kind of error is definitely not something we currently support and will have to figure out eventually. 👍

backend/compact-connect/stacks/api_stack/v1_api/query_providers.py
jusdino: Yeah, I guess we could just remove these. I was kind of thinking that this was a way to point out that the lack of `.grant` calls here was not a mistake, but I can just make the comment more clear.

landonshumway-ia (approved):
Great, this looks much better with the CCPermissionsAction enum

landonshumway-ia:
@jlkravitz This should be ready for your review.

backend/compact-connect/common_constructs/data_migration.py
```diff
@@ -0,0 +1,113 @@
...
+        migration_dir: str,
+        lambda_environment: dict,
+        role: IRole = None,
+        custom_resource_properties: dict = None,
+    ):
+        """
+        This construct is used to run a data migration.
+        It will create a lambda function and a provider that will run the migration.
+
+        :param migration_dir: The directory containing the migration code. Name the directory after the associated
```
jlkravitz, line 30: Out of curiosity, why tie migrations to issues rather than PRs or git commits? It seems the latter would be easier to track in the scenario where I have my own local deployment, and I pull new changes from the remote working branch. Alternatively, why not suggest sequential directories (1,2,3,4,etc.), such that it's clear the order in which they were created and should be run?

Not worth holding up this PR for that, but just curious if you have thoughts.

backend/compact-connect/lambdas/python/common/tests/function/test_data_client.py
```diff
@@ -422,3 +567,50 @@ def test_claim_privilege_id_increments_existing_counter(self):
...
+                'pk': 'aslp#SSN#123-12-5678',
+                'sk': 'aslp#SSN#123-12-5678',
+                'providerIdGSIpk': 'aslp#PROVIDER#89a6377e-c3a5-40e5-bca5-317ec854c570',
+                'compact': 'aslp',
+                'ssn': '123-12-5678',
+                'providerId': '89a6377e-c3a5-40e5-bca5-317ec854c570',
+            }
+        )
+
+        # We didn't create the provider this time, so this won't exist
```
jlkravitz, line 614: Copy/pasted comment, I believe.

backend/compact-connect/lambdas/python/provider-data-v1/handlers/providers.py
```diff
@@ -116,8 +120,46 @@ def get_provider(event: dict, context: LambdaContext):  # noqa: ARG001 unused-ar
...
+        logger.info('Processing provider SSN request')
+        provider_information = get_provider_information(compact=compact, provider_id=provider_id)
+
+        # Inspect the caller's scopes to determine if they have readSSN permission for this provider
+        if not user_has_read_ssn_access_for_provider(
+            compact=compact,
+            provider_information=provider_information,
+            scopes=get_event_scopes(event),
+        ):
+            metrics.add_metric(name='unauthorized-ssn-access', value=1, unit='Count')
```
jlkravitz, line 152: Glad we're tracking this carefully! Does AWS automatically associate any IP or user information with these unauthorized accesses, and are they worth integrating?

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_providers.py
```diff
@@ -385,3 +385,82 @@ def test_get_provider_returns_expected_general_response_when_caller_does_not_hav
         self._when_testing_get_provider_response_based_on_read_access(
             scopes='openid email aslp/readGeneral', expected_provider=expected_provider
         )
+
+
+@mock_aws
+class TestGetProviderSSN(TstFunction):
```
jlkravitz, line 391: Great test coverage and well-written tests. Thanks!

backend/compact-connect/stacks/persistent_stack/__init__.py
```diff
@@ -203,6 +204,17 @@ def _add_data_resources(self, removal_policy: RemovalPolicy):
         )
 
         self.ssn_table = SSNTable(self, 'SSNTable', removal_policy=removal_policy)
+        # Run migration 392 (GH issue number) to add a GSI pk field to the SSN records
```
jlkravitz, line 207: Should this code stay uncommented after it runs once?

backend/compact-connect/stacks/persistent_stack/ssn_table.py
```diff
@@ -58,8 +58,9 @@ def __init__(
                             'dynamodb:BatchGetItem',
                             'dynamodb:BatchWriteItem',
                             'dynamodb:PartiQL*',
-                            'dynamodb:Query',
-                            'dynamodb:Scan',
+                            # We will allow Scan to open up the table for migration
+                            # TODO: Uncomment this after the migration is complete  # noqa: FIX002
```
jlkravitz, line 62: I'm confused- shouldn't this be commented _out_ after the migration is complete, at which point the Scan permission is no longer needed?

jlkravitz (commented):
Looks good! Appreciate the attention to detail when it comes to SSN access. A few comments/questions.

backend/compact-connect/lambdas/python/provider-data-v1/handlers/providers.py
jusdino: We have logs of request source IP and user id (if authenticated) for all calls via our API access logs, but we shouldn't include dynamic values like that into these metrics because they're not really designed to include dynamic metadata (_dimensions_ in CloudWatch metric jargon). It becomes prohibitively expensive if you start logging metrics with a lot of different dimension values.

backend/compact-connect/stacks/persistent_stack/__init__.py
jusdino: It will only run the migration once, so no need to comment it out until it has been deployed across all environments and we're ready to remove it. The CustomResource mechanism tracks the migration as if it were a physical resource we were 'provisioning', so once we 'Create' this, it will not be invoked again unless we 'Update' it by changing the properties we feed the custom resource (we're feeding it none), or we 'Delete' it by removing it. In the Delete case, our handler of the Delete event is simply a no-op.

backend/compact-connect/stacks/persistent_stack/ssn_table.py
jusdino: This is a `DENY` statement on the table policy, so once we're done with the migration, uncommenting the `dynamodb:Scan` line will block all Scan actions on this table.

backend/compact-connect/common_constructs/data_migration.py
jusdino: I didn't tie it to a PR number because we don't know what the number of the PR will be at the time we're developing the feature. I'm not sure about the git commit. How would you envision a git-commit associated migration id being named?

For sequential directories, we don't want you to be concerned with the order that migrations should be run. Generally, we'll be trying to make sure each of these migration scripts is pretty self-contained and not dependent on deployment order. If they do have a required deployment order, we'll handle that in the cdk code by marking one migration as dependent on the other, so CloudFormation manages the sequencing, rather than leaving you to do any fancy git footwork to make sure updates go in the right order.

jlkravitz: Got it- that makes sense. My mind was probably in web framework migration land, which often produce sequential migrations that can be run (or undone) in sequence from the initial starting state. This is obviously not that.

backend/compact-connect/lambdas/python/provider-data-v1/handlers/providers.py
jlkravitz: Makes sense!

backend/compact-connect/stacks/persistent_stack/__init__.py
jlkravitz: Ah, that makes sense. Thanks for explaining.

backend/compact-connect/stacks/persistent_stack/ssn_table.py
jlkravitz: Aha! Missed that.

jlkravitz (approved):

jlkravitz:
@isabeleliassen This is good to merge. The failed test is due to an out-of-date dependency that has been fixed in `development`, but not in this branch.

isabeleliassen:
@jlkravitz we'll need your approval review as well!

isabeleliassen (approved):

jlkravitz (approved):

PR #542, opened 2025-02-12
Feat/reporting updates

### LINKED ISSUE #492: Adjustments to reporting csv format
Based on feedback from CSG EDs, we need to make the following adjustments to the transaction csv reports

For the 'financial summary report':
- Remove the word 'Total' from compact fees
- Show the number of privileges purchased per state
- Show the credit card fees collected (this may not be possible, as it appears that information is dependent on the Merchant Service Provider (MSP) and is not available through the API, see [Batch and Settlement Process](https://support.authorize.net/knowledgebase/Knowledgearticle/?code=000001244#:~:text=Once%20a%20batch%20has%20settled,funds%20into%20your%20merchant%20account.)) Edit - we will just be estimating these based on the compact's transaction fee rate, and making it explicit that it is an estimation EDIT Feb 12: We will not be estimating these since not all compacts will not be using the default MSP, which complicates attempting to make estimates for this.
- Include date range in report file name

For the 'transaction detail reports':
- Include date range in report file name
- Add UTC to 'Transaction Date' label
- Include 'Transaction Status' field

Questions:
1. Authorize.net sends monthly statements about the transaction fees to owners of the Authorize.net account. For the MVP, do we simply not include transaction fees in the reporting, and have the Compact admins reference this monthly statement?
answer: We will only be providing an estimate for the transaction fees to be collected, with a notice that they need to check with their bank statement for the actual amount.

If we must include transaction fees in the reporting, the following questions must be answered:
2. Will the compacts split the transactions fees between themselves and the states, or will the compact cover the transaction fee costs just themselves? (the first choice is SIGNIFICANTLY more expensive to implement and prone to accounting discrepancies)
      a.  If states are responsible for paying a portion of the transaction fees, and a user purchases multiple privileges within a single transaction, how should those transaction fees be split up? (This could get really complicated)
      b. If states are responsible for paying a portion of the transaction fees, should we include the fee amount in their summary report?

Answer: In addition to the based administrative fee, the ASLP and OTP compacts will charge a flat rate per privilege to absorb the costs of the transaction fees. This flat fee will show up as a separate line item in the transaction.

3. Are all the compacts using authorize.net's Merchant Service Provider (Global Payments)?
Answer: No

5. Based on a phone conversation with an authorize.net representative, the credit card fees are not available in the authorize.net API, but they can be calculated by applying the 2.9% + 30 cents per transaction. According to the representative, these transaction fee rates will never change without a service level agreement between the compact and authorize.net. Is it acceptable to use this rate to calculate what the transaction fees will be based on this rate? We could include a configuration field for the compact where they must specify what their transaction fee (We will include a notice in the summary report that the shown fees are based on the compacts configured transaction fee rating, and the admins should confirm this aligns with the actual rate they are being charged) 
Answer: As part of compact onboarding, the compact will specify what the transaction fee rate is with their MSP.

landonshumway-ia:
Updating the CSV reports based on feedback from compact EDs, the biggest change being we are now including the associated privilege ids for each transaction, so that admins can make the correlation between charges in authorize.net and the compact connect system.

### Requirements List
- As part of this change, we are now updating the line item 'itemId' for every privilege purchase with the prefix `priv:` to more readily distinguish what is a privilege jurisdiction cost and what is compact related fees. This means that older transactions which do not include the `priv:` prefix will not show up in transaction reporting. Given that the system is not live and this change is only related to reporting, this will not impact any other functionality of the system.

### Description List
- Updated compact financial summary reports to include the following:
    - privilege purchase count by jurisdiction
    - total processed amount
    - change `Compact Fee` to `Administrative Fees`
- Updated transaction detail reports to include the following:
   - Collected transaction fee if compact charged licensee transaction fees
   - Specify settlement date is UTC
   - Associated privilege id for the transaction according to compact connect
   - Transaction Status according to authorize.net

- Added compactTransactionId global secondary index for mapping authorize.net transaction ids to compact connect privilege ids.

### Testing List
- load testing script added for processing large numbers of transactions
- local testing in sandbox environment
- added unit/functional tests for reporting lambdas
- Code review

Closes #492


backend/compact-connect/stacks/persistent_stack/provider_table.py
```diff
@@ -51,6 +52,21 @@ def __init__(
             sort_key=Attribute(name='licenseGSISK', type=AttributeType.STRING),
             projection_type=ProjectionType.ALL,
         )
+        self.add_global_secondary_index(
+            index_name=self.compact_transaction_gsi_name,
```
jusdino, line 56: Do we need a DataMigration to add this GSI field to existing transactions in the table?

backend/compact-connect/lambdas/python/common/cc_common/data_model/transaction_client.py
```diff
@@ -128,3 +119,67 @@ def _query_transactions_for_month(
...
+            response = self.config.provider_table.query(
+                IndexName=self.config.compact_transaction_id_gsi_name,
+                KeyConditionExpression=Key('compactTransactionIdGSIPK').eq(gsi_pk),
+            )
+
+            # Process each privilege record
+            for jurisdiction in jurisdictions_to_process:
+                item_id = f'priv:{compact}-{jurisdiction}'
+                # find the first privilege record for the jurisdiction
+                matching_privilege = next(
```
jusdino, line 157: Let's add `providerId` to the GSI projected attributes, so we can include a sanity check here to verify that the privilege we matched is actually associated with the individual we expect.

jusdino (changes requested):
Looks great! A few minor items for you:

jusdino (approved):
Great! @jlkravitz , this is ready for you!

backend/compact-connect/lambdas/python/purchases/tests/function/test_handlers/test_transaction_history.py
```diff
@@ -240,3 +327,120 @@ def test_process_settled_transactions_returns_batch_failure_status(self, mock_pu
...
+        )
+        self.assertEqual('UNKNOWN', stored_transactions['Items'][0]['lineItems'][0]['privilegeId'])
+
+    @patch('handlers.transaction_history.PurchaseClient')
+    def test_process_settled_transactions_maps_privilege_ids_to_line_items(self, mock_purchase_client_constructor):
+        """Test that transactions are stored in DynamoDB."""
+        from handlers.transaction_history import process_settled_transactions
+
+        # In this test, we simulate a user purchasing a privilege for ky,
+        # then making another purchase for oh, ky, and ne.
```
jlkravitz, line 359: L399-L446 seem to imply the reverse– first a purchase of oh, ky, and ne. Then, an update on KY. Am I missing something?

backend/compact-connect/stacks/persistent_stack/provider_table.py
```diff
@@ -51,6 +52,22 @@ def __init__(
...
+            partition_key=Attribute(name='compactTransactionIdGSIPK', type=AttributeType.STRING),
+            # in this case, we only need to include the privilegeId, updatedValues, type, and jurisdiction
+            # since this GSI is used to map compactTransactionIds to privileges
```
jlkravitz, line 59: Looks like other attributes are included too, though?

backend/compact-connect/lambdas/python/purchases/tests/function/test_handlers/test_transaction_reporting.py
```diff
@@ -1025,10 +1097,178 @@ def test_generate_weekly_report_includes_expected_settled_transactions_for_full_
...
+        from handlers.transaction_reporting import generate_transaction_reports
+
+        _set_default_lambda_client_behavior(mock_lambda_client)
+
+        self._add_compact_configuration_data(
+            jurisdictions=[OHIO_JURISDICTION, KENTUCKY_JURISDICTION, NEBRASKA_JURISDICTION]
+        )
+
+        mock_user = self._add_mock_provider_to_db('12345', 'John', 'Doe')
+        # Create a transaction with in which the licensee is charged transaction fees
```
jlkravitz, line 1126: within?

backend/compact-connect/lambdas/python/purchases/tests/function/test_handlers/test_transaction_reporting.py
```diff
@@ -1025,10 +1097,178 @@ def test_generate_weekly_report_includes_expected_settled_transactions_for_full_
...
+            generate_transaction_reports(generate_mock_event(), self.mock_context)
+
+        # check that the error message contains the expected line item id
+        self.assertEqual(
+            'One or more errors occurred while generating reports. '
+            "Errors: ['transaction line item id does not match any known pattern "
+            f'- transactionId={MOCK_TRANSACTION_ID} '
+            "- itemId=unknown-line-item-fee']",
+            exc_info.exception.message,
+        )
```
jlkravitz, line 1235: Do we want it to throw an error or just log a warning? What's the benefit of throwing an error if it still saves the (partial?) report?

backend/compact-connect/tests/smoke/load_test_privilege_purchasing.py
```diff
@@ -0,0 +1,201 @@
...
+    '4007000000027',  # Visa
+    '4012888818888',  # Visa
+    '4111111111111111',  # Visa
+    '5424000000000015',  # Mastercard
+    '2223000010309703',  # Mastercard
+    '2223000010309711',  # Mastercard
+    AMEX_CARD_NUMBER,  # American Express
+    '6011000000000012',  # Discover
+    '3088000000000017',  # JCB
+    # '38000000000006',    # Diners Club/ Carte Blanche
```
jlkravitz, line 41: Should this be commented out?

backend/compact-connect/tests/smoke/load_test_privilege_purchasing.py
```diff
@@ -0,0 +1,201 @@
...
+from smoke_common import (
+    SmokeTestFailureException,
+    call_provider_users_me_endpoint,
+    get_provider_user_auth_headers_cached,
+)
+
+# Load-Testing Script for populating authorize.net with transactions
+# This is a bit of a hack, because in order to purchase so many
+# privileges with a single user we have to delete the records from
+# the DB right after completing the transactions.
```
jlkravitz, line 18: Why?

jlkravitz (commented):
Looks good! A few comments/questions.

backend/compact-connect/lambdas/python/common/cc_common/data_model/transaction_client.py
```diff
@@ -128,3 +119,73 @@ def _query_transactions_for_month(
...
+            # Process each privilege record
+            for jurisdiction in jurisdictions_to_process:
+                item_id = f'priv:{compact}-{jurisdiction}'
+                # find the first privilege record for the jurisdiction that matches the provider ID
+                matching_privilege = next(
+                    (
+                        item
+                        for item in response.get('Items', [])
+                        if item['jurisdiction'].lower() == jurisdiction and item['providerId'] == licensee_id
+                    ),
```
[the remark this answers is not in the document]
landonshumway-ia: Not currently, as the system only supports one license type per jurisdiction. We have an upcoming ticket on the backlog for supporting multiple license types within the same jurisdiction, however, my understanding is that a user will only be able to purchase privileges for one specific license at a time, so the transaction id will still be unique to a particular privilege, so we can be confident that the privilege id we fetch from this query is specific to the transaction that occurred.

backend/compact-connect/lambdas/python/common/cc_common/data_model/transaction_client.py
```diff
@@ -128,3 +119,73 @@ def _query_transactions_for_month(
...
+                        privilege_id = matching_privilege['previous']['privilegeId']
+
+                    # Find and update the matching line item
+                    self._set_privilege_id_in_line_item(
+                        line_items=line_items, item_id=item_id, privilege_id=privilege_id
+                    )
+                else:
+                    logger.error(
+                        'No matching jurisdiction privilege record found for transaction. '
+                        'Cannot determine privilege id for this transaction',
```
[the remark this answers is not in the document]
landonshumway-ia: No, transactions are specific to a single compact, as each compact will have their own unique authorize.net account. We will never have a single transaction with line items for multiple compacts.

backend/compact-connect/lambdas/python/purchases/tests/function/test_handlers/test_transaction_reporting.py
landonshumway-ia: We throw the error in order to fire an alert for support to investigate the issue, as this situation should never happen in production and warrants an investigation into the cause. We still want to send the report in this case, so that admins have visibility into the generated report with the unknown charges. This way we have visibility for both admins and technical staff to triage the issue.

backend/compact-connect/lambdas/python/purchases/tests/function/test_handlers/test_transaction_history.py
landonshumway-ia: I split out the tests with more comments to hopefully simplify the distinction of what we are testing for here.

backend/compact-connect/stacks/persistent_stack/provider_table.py
landonshumway-ia: I've updated the comment

backend/compact-connect/lambdas/python/purchases/tests/function/test_handlers/test_transaction_reporting.py
landonshumway-ia: Fixed

backend/compact-connect/tests/smoke/load_test_privilege_purchasing.py
landonshumway-ia: I've removed it, it is an invalid card.

backend/compact-connect/tests/smoke/load_test_privilege_purchasing.py
landonshumway-ia: I've added clarifying comments

jlkravitz (commented):
One question!

Just to note for the next sprint– I'm not sure if this warrants updates to the README or architecture diagram or something else, but I'm realizing the number of assumptions the system is making is starting to grow. I'm admittedly not as deep into the code as you all are, but that growth has made it harder to understand some of the logic we've incorporated into the system (in this case, matching privileges with transactions). Perhaps what we need is some kind of privilege purchase flow diagram, showing how data is generated, how it's stored, and how different pieces relate to one another (privilege record, privilege update record, transaction record, etc.).

landonshumway-ia:
@jlkravitz I'm going to add some more supporting documentation concerning the entire transaction reporting flow with an updated diagram.

landonshumway-ia:
@jlkravitz As I was writing up the design walkthrough for this, I realized there is an error in my current transaction collection logic that needs to be addressed to account for partial batch settlement errors in authorize.net. I am working on that fix currently.

landonshumway-ia:
@jlkravitz [This commit](https://github.com/csg-org/CompactConnect/pull/542/commits/86135790057f82f0d17cae03f050e666cee17c04) and [this commit](https://github.com/csg-org/CompactConnect/pull/542/commits/d18715f7ee698bfea79c06002838da3d6d9f49e3) are the two changes I made since your last review and should be ready for another review. Thanks

backend/compact-connect/docs/design/README.md
```diff
@@ -262,3 +263,88 @@ When purchasing privileges, providers must accept all required attestations. The
...
+     - Settled: Successfully processed
+     - Settlement Error: Failed to settle (triggers alerts) see [Batch Settlement Failure Handling](#batch-settlement-failure-handling)
+
+3. **Transaction Collection**
+   - A Step Function workflow runs daily at 5:00 PM Pacific Time (1:00 AM UTC)
+   - The workflow:
+     - Queries Authorize.net for settled batches in the last 24 hours
+     - Processes transactions in groups of up to 500 (to avoid lambda timeouts)
+     - Fetches the associated privilege id of the privilege that was purchased for that transaction 
+      and injects it into the data that is stored in the Transaction History DynamoDB Table
```
jlkravitz, line 299: To clarify- the "Transaction Collection" step is basically how we track Authorize.net transactions in our own database? What do you mean by "Fetches" here? In my mind, since the privilege ID should be in the Authorize.net response (since we store it there), "fetching" is really just "copying" the ID to our DynamoDB table. Is that true?

backend/compact-connect/docs/design/README.md
```diff
@@ -262,3 +263,88 @@ When purchasing privileges, providers must accept all required attestations. The
...
+   - By default, Authorize.net batches settlements daily at 4:00 PM Pacific Time (this can be changed by the Authorize.net account owners)
+   - Each batch contains transactions since the last batch was settled
+   - Batches can be in one of several states:
+     - Settled: Successfully processed
+     - Settlement Error: Failed to settle (triggers alerts) see [Batch Settlement Failure Handling](#batch-settlement-failure-handling)
+
+3. **Transaction Collection**
+   - A Step Function workflow runs daily at 5:00 PM Pacific Time (1:00 AM UTC)
+   - The workflow:
+     - Queries Authorize.net for settled batches in the last 24 hours
```
jlkravitz, line 296: Is this still true given your changes? Don't we also store failed batches now, too?

jlkravitz (commented):
Really helpful updates to the README. Thank you! I read it as if I didn't have much knowledge of the code- just a couple of comments on items that might be worth clarifying.

backend/compact-connect/docs/design/README.md
landonshumway-ia: We do not record the privilege id in the authorize.net transaction, as this presents a chicken and egg problem (we need to authorize.net transaction id to create the privilege with a privilege id, so one has to be created before the other is known). Because of this, we map the privilege ids from the privileges into the transaction data we store into our transaction history table. I've updated the wording here to hopefully make it clear that we are getting the privilege id from the compact connect provider data table.

backend/compact-connect/docs/design/README.md
landonshumway-ia: I've updated the wording here to be more general

jlkravitz (approved):
Looks great. Thanks for the README updates. @isabeleliassen Good to merge this one.

isabeleliassen (approved):

PR #545, opened 2025-02-12
fix/attestation typo

jlkravitz:
Fix typo identified in Sprint 17 PR #543


landonshumway-ia:
@jlkravitz you will also need to update the compact configuration snapshot test to account for this change.

For the record, all of the attestation text came directly from the [Figma designs](https://www.figma.com/design/SYM0uWszsu8Sf0YfxAhIMY/JCC?node-id=6575-1111&t=CYobKg6lvrWIs6ic-0) as it was assumed this was the latest version of the attestations. If that was an incorrect assumption, we will need to make sure the attestation text matches what is actually expected.

carlsims (approved):

PR #561, opened 2025-02-17
Feat/machine-to-machine app client support

### LINKED ISSUE #463: API data upload enable for states

landonshumway-ia:
Reorganizes Cognito resource server configuration to better support machine-to-machine 
authentication for automated license data uploads. This change enables IT departments 
to securely automate their license data submissions through app clients.

Previously, we configured Cognito resource servers per compact with scopes in the 
format {compact}/{jurisdiction}.{action}. This led to complications when managing 
jurisdiction-specific scopes and pushed us over to the 100-scope limit per 
resource server imposed by Cognito, which would not work for machine-to-machine app clients.

To address this, we have inverted the scope model. In addition to our compact resource servers, there is now a 
resource server per jurisdiction, and inside of those resource servers we are defining scoped permissions for each compact. For example:

Before
```
aslp/ky.write
```

Updated
```
ky/aslp.write
```
This gets rid of the complex course-grain vs fine-grain scope model, and gives us plenty of room for scope growth within a jurisdiction resource server (supporting up to 10 compacts with 10 actions each) with the potential to create additional resource servers as needed (Cognito supports a max of 300 resource servers per user pool, see https://docs.aws.amazon.com/cognito/latest/developerguide/quotas.html#resource-quotas).

Initially, machine-to-machine app clients will be created manually by support technical staff. This change adds detailed instructions for this process along with an example yaml file showing how these app clients are to be documented in the project.

### Requirements List
- This change will break any active user sessions at the time of deployment, but any new logins after the deployment will work as is, since the scope setting is performed at login time. Because the Staff user permission schema in Dynamo is not actually being changed, this change is backwards compatible with the data model, and does not require any data migrations.

### Description List
- inverted our scope model in the python code to account for resource servers at the jurisdiction level
  - updated the pre-token generation to generate the scopes according to this updated pattern
  - updated the auth wrapper methods to check for scopes at both the compact and jurisdiction resource level
 - Updated Staff Users CDK definition to create resource servers per active jurisdiction in an environment 
 - Added documentation for creating app clients with example yaml file.

### Testing List
- Added CDK tests to verify expected resources servers are created
- Created test app client in sandbox environment and verified access tokens could authenticate with the system
- Functional/unit tests added for lambda code
- Code review

Closes #463


backend/compact-connect/stacks/persistent_stack/__init__.py
```diff
@@ -436,3 +437,58 @@ def _get_ui_base_path_url(self) -> str:
...
+        active for the environment.
+        """
+
+        active_jurisdictions = []
+
+        # Read all jurisdiction configuration YAML files from each active compact directory
+        for jurisdiction_config_file in os.listdir(os.path.join('compact-config', compact)):
+            if jurisdiction_config_file.endswith('.yml'):
+                with open(os.path.join('compact-config', compact, jurisdiction_config_file)) as f:
+                    # convert YAML to JSON
```
jusdino, line 485: Strictly speaking, JSON isn't involved here - we're just loading the YAML file into a python object.

backend/compact-connect/tests/app/test_pipeline.py
```diff
@@ -58,29 +58,58 @@ def test_synth_pipeline(self):
...
+            # remove dynamic user pool id
+            del resource_server_properties['UserPoolId']
+            jurisdiction_resource_server_config.append(resource_server_properties)
+
+        # sort the resource server list by jurisdiction for consistency
+        jurisdiction_resource_server_config.sort(key=lambda jurisdiction: jurisdiction['Identifier'])
+        # sort the scopes within the resource server by name for consistency
+        for resource_server in jurisdiction_resource_server_config:
+            resource_server['Scopes'].sort(key=lambda scope: scope['ScopeName'])
+        self.compare_snapshot(
```
jusdino, line 107: I think an explanation here would be helpful. Am I right that we have no `coun.*` scopes in this snapshot because, while the compact is active in the `test` environment, none of its jurisdictions are so the `ky/coun.*`, etc, scopes are, rightly, being skipped?

backend/compact-connect/tests/app/test_pipeline.py
```diff
@@ -58,29 +58,58 @@ def test_synth_pipeline(self):
...
         persistent_stack = self.app.pipeline_stack.test_stage.persistent_stack
         persistent_stack_template = Template.from_stack(persistent_stack)
 
         # Get the resource servers created in the persistent stack
-        resource_servers = persistent_stack.staff_users.resource_servers
+        resource_servers = persistent_stack.staff_users.compact_resource_servers
         # We must confirm that these scopes are being explicitly created for each compact
         # which are absolutely critical for the system to function as expected.
-        self.assertEqual(self.context['compacts'], list(resource_servers.keys()))
+        self.assertEqual(sorted(self.context['compacts']), sorted(list(resource_servers.keys())))
```
jusdino, line 69: Just checking on your intent here, since I'm not 100% certain from your comment: Are you deliberately enforcing that, for the `test` environment only, we now _must_ have the compact configuration be active for every compact we define in `cdk.json`?

backend/compact-connect/tests/unit/test_app_client_yaml.py
```diff
@@ -0,0 +1,103 @@
...
+                    for env in yaml_content['environments']:
+                        self.assertIn(env, VALID_ENVIRONMENTS, f'Invalid environment: {env}')
+
+                    # ensure ownerContact is an email address
+                    # at least one @ and .
+                    if not re.match(r'[^@]+@[^@]+\.[^@]+', yaml_content['ownerContact']):
+                        self.fail(f"Invalid ownerContact: {yaml_content['ownerContact']}")
+
+                # ensure scopes is a list, we don't care about the values
+                # since the
```
jusdino, line 101: `since the`?

jusdino (changes requested):
Really good stuff! I haven't done a full test-drive yet, but I've got a bunch of minor comments you could look at in the meantime, so here's a preliminary review:

backend/compact-connect/tests/app/test_pipeline.py
landonshumway-ia: I've updated the comment here to be more clear that the test simply expects all compacts that are marked as active to be present

backend/compact-connect/tests/unit/test_app_client_yaml.py
landonshumway-ia: That comment is really outdated. I updated it to match what it is actually checking for.

jusdino (approved):
Awesome! I poked at this all day in my sandbox and couldn't find any flaws.

jusdino:
@jlkravitz , this is ready for you!

backend/compact-connect/tests/app/test_pipeline.py
```diff
@@ -58,29 +58,92 @@ def test_synth_pipeline(self):
...
+    def test_synth_generates_jurisdiction_resource_servers_with_expected_scopes_for_staff_users_test_stage(self):
+        """
+        Test that the jurisdiction resource servers are created with the expected scopes
+        for the staff users in the test environment.
+        """
+        persistent_stack = self.app.pipeline_stack.test_stage.persistent_stack
+        self._when_testing_jurisdiction_resource_servers(
+            persistent_stack=persistent_stack,
+            snapshot_name='JURISDICTION_RESOURCE_SERVER_CONFIGURATION_TEST_ENV',
+            overwrite_snapshot=True
```
jlkravitz, line 132: Do we intend to overwrite the snapshot here? If yes, can we add a comment (since this is atypical in the codebase)?

backend/compact-connect/tests/app/test_pipeline.py
```diff
@@ -58,29 +58,92 @@ def test_synth_pipeline(self):
...
+    def test_synth_generates_jurisdiction_resource_servers_with_expected_scopes_for_staff_users_prod_stage(self):
+        """
+        Test that the jurisdiction resource servers are created with the expected scopes
+        for the staff users in the prod environment.
+        """
+        persistent_stack = self.app.pipeline_stack.prod_stage.persistent_stack
+        self._when_testing_jurisdiction_resource_servers(
+            persistent_stack=persistent_stack,
+            snapshot_name='JURISDICTION_RESOURCE_SERVER_CONFIGURATION_PROD_ENV',
+            overwrite_snapshot=True
```
jlkravitz, line 144: Do we intend to overwrite the snapshot here? If yes, can we add a comment (since this is atypical in the codebase)?

backend/compact-connect/tests/app/test_pipeline.py
```diff
@@ -58,29 +58,92 @@ def test_synth_pipeline(self):
...
+            del resource_server_properties['UserPoolId']
+            jurisdiction_resource_server_config.append(resource_server_properties)
+
+        # sort the resource server list by jurisdiction for consistency
+        jurisdiction_resource_server_config.sort(key=lambda jurisdiction: jurisdiction['Identifier'])
+        # sort the scopes within the resource server by name for consistency
+        for resource_server in jurisdiction_resource_server_config:
+            resource_server['Scopes'].sort(key=lambda scope: scope['ScopeName'])
+        # this will only include resource server scopes for compacts/jurisdictions that are marked as active
+        # for the test environment
```
jlkravitz, line 115: can't this be the prod environment, too? Should we change to–

```suggestion
        # for the environment
```

backend/compact-connect/app_clients/README.md
```diff
@@ -0,0 +1,106 @@
+# App Client Management for Staff Users
+
+## Overview
+This document outlines the process for managing Cognito app clients for machine-to-machine authentication in the Staff Users pool. All app clients must be documented in yaml files in the `/app_clients` directory for disaster recovery purposes. See the example app client for the format.
```
jlkravitz, line 4: Can you say more about what you mean when you say "disaster recovery purposes"? I think it's perfectly reasonable to document clients in this directory, but I'm not sure I understand this reasoning.

backend/compact-connect/app_clients/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Creating a New App Client
+
+### 1. Documentation Prerequisites
+Before creating a new app client, ensure you have:
+- Jurisdiction requirements documented
+- Contact information for the consuming team
+- Approval for requested scopes
```
jlkravitz, line 12:
```suggestion
- Approval to grant the app client with the requested scopes
```

backend/compact-connect/app_clients/README.md
```diff
@@ -0,0 +1,106 @@
...
+- Update access to the Staff Users user pool for the needed AWS accounts
```
jlkravitz, line 13: What does this mean?

backend/compact-connect/app_clients/README.md
```diff
@@ -0,0 +1,106 @@
...
+   --prevent-user-existence-errors ENABLED \
+   --generate-secret \
+   --token-validity-units AccessToken='minutes'\
+   --access-token-validity 15 \
+   --allowed-o-auth-flows-user-pool-client \
+   --allowed-o-auth-flows 'client_credentials' \
+   --allowed-o-auth-scopes '<compact>/readGeneral' '<jurisdiction>/<compact>.write'
+   ```
+
+   It is important to note that generated access tokens will expire after 15 minutes, so the consuming team will need to have application logic to generate a new access token before it expires (this time can be adjusted according to the needs of the consuming team by updating the access token validity in the cli command).
```
jlkravitz, line 60: It took me a while to understand what I think is true: This command generates a client, not an access token, that has the ability to _produce_ access tokens with 15 minute lifespans. Can you create a client that can generate tokens with _no_ expiration, or does AWS disallow that?

If that's right, perhaps we can clarify this sentence (just a draft below – feel free to rewrite).
```suggestion
   This command creates an app client with the ability to generate access tokens that expire after 15 minutes. AWS requires that this field is specified when creating an app client, but the value can be adjusted according to the needs of the consuming team. (Or, "Although AWS allows you to create an app client that can generate app tokens with no expiration date, we do not recommend this because...") Regardless, the consuming team will need to have application logic to generate new access tokens on a regular basis.
```

backend/compact-connect/app_clients/README.md
```diff
@@ -0,0 +1,106 @@
...
+        }
+   }
+   ```
+
+   These credentials should be securely transmitted to the consuming team via a encrypted channel. You will also need to provide them with the user pool domain for the Staff Users user pool, which can be determined using the following cli command.
+   ```
+   aws cognito-idp describe-user-pool --user-pool-id '<user pool id>' --query 'UserPool.Domain' --output text
+   ```
+
+   If the team has questions about generating access tokens, refer them to Amazon's [documentation](https://docs.aws.amazon.com/cognito/latest/developerguide/token-endpoint.html). You can also give them the following curl example for generating an access token (Instruct them to replace the `<user pool domain>`, `<client id>`, `<client secret>` with the values you send them. The `<jurisdiction>` and `<compact>` scope values should be the jurisdiction postal code and compact abbreviation they are requesting access to such as `ky` for Kentucky and `aslp` for the ASLP compact).
```
jlkravitz, line 82: Are state abbreviations considered "postal codes"?

backend/compact-connect/app_clients/README.md
```diff
@@ -0,0 +1,106 @@
...
+   ```
+    curl --location --request POST 'https://<user pool domain>.auth.us-east-1.amazoncognito.com/oauth2/token?grant_type=client_credentials&client_id=<client id>&client_secret=<client secret>&scope=<jurisdiction>%2F<compact>.write' \
+    --header 'Content-Type: application/x-www-form-urlencoded' \
+    --header 'Accept: application/json'
+   ```
+
+
+## Rotating App Client Credentials
+Unfortunately, AWS Cognito does not support rotating app client credentials for an existing app client. The only way to rotate credentials is to create a new app client with a new clientId and clientSecret and then delete the old one.
```
jlkravitz, line 91: Can you provide some guidance here as to why this might be done, how often it should be done, etc.?

backend/compact-connect/app_clients/README.md
```diff
@@ -0,0 +1,106 @@
+# App Client Management for Staff Users
+
+## Overview
+This document outlines the process for managing Cognito app clients for machine-to-machine authentication in the Staff Users pool. All app clients must be documented in yaml files in the `/app_clients` directory for disaster recovery purposes. See the example app client for the format.
+
```
jlkravitz, line 5: To clarify, I assume the audience of this document are IA or CSG staff? (Anyone with write permission to the codebase.)

jlkravitz (commented):
Looks good! Most of my comments are nits on the documentation, plus a few questions. As @jusdino said, it's nice how much this simplifies the scope logic.

One question– was machine-to-machine app client support not possible under the prior setup, or does this just make it simpler?

backend/compact-connect/tests/app/test_pipeline.py
landonshumway-ia: No not intentional, just got lost in the shuffle 😄 I'll fix those

backend/compact-connect/app_clients/README.md
landonshumway-ia: This is a great question which is challenging to put a specific timeframe on. Obviously the longer an app client secret is in existence the greater the chance that it can be compromised. Typically I've seen companies rotate credentials such as these anywhere from a quarterly basis to annual. For now, I can mention that it should be performed in the event of a compromised secret, which is the main use case for this section.

backend/compact-connect/app_clients/README.md
landonshumway-ia: I've updated this line to make it more clear the user needs AWS CLI access to perform the action.

backend/compact-connect/tests/app/test_pipeline.py
landonshumway-ia: nope, I've updated it to false

landonshumway-ia:
@jlkravitz To answer your initial question: machine-to-machine app client support was not possible under the old scope model, since we were granting implicit scopes not defined on the compact resource servers when users would log in using a pre-token generation trigger. There is no support for pre-token generation triggers for machine-to-machine app clients, which is why we had to redesign the model.

This should now be ready for another review. Thanks

backend/compact-connect/app_clients/README.md
jlkravitz: That sounds great.

jlkravitz (approved):
@landonshumway-ia Thanks for the updates, and the explanation. That makes sense.

isabeleliassen (approved):

Issue #521, opened 2025-02-04
frontend for deactivate privilege


isabeleliassen:
see #469

PR #562, opened 2025-02-17
Feat/display license type

### LINKED ISSUE #437: List license type (OT, OTA, etc) associated with each license and privilege
- depends on #505 
- add to licensee detail view and practitioner dashboard
- if someone is both an OT and an OTA, list both
- in licensee detail view, may have licensee who has not designated home state, so will need a version of the view with no home state
- frontend to check that the correct information is pulled from backend (based on what licensee says upon registration)
- display also shows privilege number

ChiefStief:
### Requirements List
- A licensee user in the same compact as a staff user

### Description List
- Adjusted `LicenseCard` and `PrivilegeCard` components to have license number and privilegeId, added icon when desired and added occupation abbreviation
- Created `LicenseIcon` component and use that rather than svg asset
- Added occupation abbreviations to translations
- Modified `License` model

### Testing List
- Code review
- Look at `LicenseeDashboard` and `LicenseeDetail` page and confirm that LicenseCard and PrivilegeCard match designs

Closes #437


webroot/src/components/Icons/LicenseIcon/LicenseIcon.ts
```diff
@@ -0,0 +1,24 @@
...
+    Vue,
+    toNative,
+    Prop
+} from 'vue-facing-decorator';
+
+@Component({
+    name: 'LicenseIcon',
+})
+class LicenseIcon extends Vue {
+   @Prop({ default: 'black' }) fill?: string;
```
jsandoval81, line 19: It would be preferred to just handle the `fill` in CSS. That's what we've done with other SVGs.

webroot/src/models/License/License.model.ts
```diff
@@ -112,6 +114,14 @@ export class License implements InterfaceLicense {
 
         return occupationName;
     }
+
+    public occupationAbbreviation(): string {
+        const occupations = this.$tm('licensing.occupations') || [];
+        const occupation = occupations.find((translate) => translate.key === this.occupation);
+        const occupationAbbrev = occupation?.abbrev || '';
```
jsandoval81, line 121: Is there reason to not just do the upper-casing here? We've already got the `key` & `compactKey` to pass around as an identifiers; so I'd suspect the `name` and `abbrev` are just display?

Or is the `abbrev` also a key for backend-related things?

jsandoval81 (changes requested):
Nice job. Just an FYI that we're going to have some merge conflicts between this and ticket #521 - they both touch that same area of the PrivilegeCard.

webroot/src/models/License/License.model.ts
ChiefStief: Id be fine uppercasing here. I was not just to make it more general in case we wanted it for some other use. 

At the moment, this is the only use to my knowledge. The backend takes in the occupation key.

webroot/src/locales/en.json
```diff
@@ -507,6 +507,8 @@
         "birthMonthDay": "Birth month-day",
         "pastPrivilegesTitle": "Past privileges",
         "licenseDetails": "License details",
+        "license#": "License #",
+        "privilege#": "Privilege #",
```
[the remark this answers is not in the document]
ChiefStief: Ok, we already have `licenseNumber` but I'll come up with something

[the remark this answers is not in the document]
jsandoval81: Sounds good. Seems like the intention is to be a shortened version to fit better in small UI spaced. Maybe `licenseNumberShort` or something?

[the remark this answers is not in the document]
ChiefStief: oh I went with *numSymbol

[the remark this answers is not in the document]
jsandoval81: Fair enough. Moving forward, we can treat these key names more like CSS class names, where it describes what the value is, not what it looks like.

jsandoval81 (approved):

jsandoval81:

webroot/src/components/PrivilegeCard/PrivilegeCard.ts
```diff
@@ -89,6 +91,16 @@ class PrivilegeCard extends Vue {
 
         return isPastDate;
     }
+
+    get occupationAbbrev(): string {
+        const bestHomeStateLicense = this.licensee?.bestHomeStateLicense() || new License();
+
+        return bestHomeStateLicense?.occupationAbbreviation();
```
jlkravitz, line 98: Is occupation the same as "license type", or is it something else? If the former, I wonder if it'd be worth staying consistent here on terminology.

jlkravitz (commented):
One question- otherwise looks good!

webroot/src/components/PrivilegeCard/PrivilegeCard.ts
jsandoval81: By this point, that change would probably mean touching components, models, translations, tests, etc. that would probably call for this to go back to the in-progress column on the board, and do some regression testing outside the scope of this ticket.

@ChiefStief Is this something reasonable to add to the current cleanup ticket? Or should it be a separate ticket?

ChiefStief: I can add it to the clean up ticket. @jlkravitz Does that work for you?

jlkravitz: Sounds good. I may also add the license vs. privilege distinction to that ticket - If I recall, the usage is a little different in the front and backends which can get a little confusing.

And thanks for the context.

jlkravitz: @ChiefStief Let me know if these cleanup items have been tracked, and I'll then close this conversation and approve the PR.

ChiefStief: @jlkravitz I have added those items as the last two bullet points in this ticket: https://github.com/csg-org/CompactConnect/issues/581

isabeleliassen (approved):

PR #563, opened 2025-02-18
Feat/deactivate privilege

### LINKED ISSUE #469: admins can deactivate a privilege backend
- does not contemplate how refunds would work
- initially will be only accessible for admin permission level
- state admins would only be able to cancel privileges for their state; compact admins might need to cancel privileges from all states
- should still appear in privilege history (still should be displayed in recent or past privileges as other privileges would be).
- privilege revocation permission needed
- design will be needed
- no multi-cancel functionality needed
- notification for practitioner and for privilege state. Notification is immediate
- who in a state would need to be notified? Potentially creates new notification category for states--for now, we will notify those in state who receive reports
- reinstate privileges is not contemplated under MVP
- any revoked or canceled privilege will be possible to purchase again (system will not prevent from re-purchasing a revoked privilege, except under adverse action)

Backend work:
- new endpoint
- only admins can call
- need ID of privilege and provider
- add deactivation to event history

jusdino:
### Description List
- Add deactivate provider API endpoint
- Fire data event bus event on deactivation of privilege or license

Closes #469


backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
```diff
@@ -889,3 +890,93 @@ def find_home_state_license(self, *, compact: str, provider_id: str, licenses: l
...
+            KeyConditionExpression=Key('pk').eq(f'{compact}#PROVIDER#{provider_id}')
+            & Key('sk').eq(f'{compact}#PROVIDER#privilege/{jurisdiction}#'),
+        )['Items']
+
+        # Find the main privilege record (not history records)
+        privilege_record_schema = PrivilegeRecordSchema()
+        privilege_record = next(
+            (privilege_record_schema.load(record) for record in privilege_records if record['type'] == 'privilege'),
+            None,
+        )
```
landonshumway-ia, line 916: You shouldn't need to iterate through the list since you are doing an `eq` on the sk right? Wouldn't it make more sense to check that only one privilege item item was returned? In fact, we could just use the getItem api call instead of query, since we know the exact pk and sk in this case.

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
```diff
@@ -889,3 +890,93 @@ def find_home_state_license(self, *, compact: str, provider_id: str, licenses: l
...
+
+        if privilege_record is None:
+            raise CCNotFoundException(f'Privilege not found for jurisdiction {jurisdiction}')
+
+        # If already inactive, do nothing
+        if privilege_record.get('persistedStatus', 'active') == 'inactive':
+            return
```
landonshumway-ia, line 923: we should probably log a message here to indicate that the privilege was already inactive.

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
```diff
@@ -889,3 +890,93 @@ def find_home_state_license(self, *, compact: str, provider_id: str, licenses: l
...
+                    # We're relying on the schema to trim out unneeded fields
+                    **privilege_record,
+                },
+                'updatedValues': {
+                    'persistedStatus': 'inactive',
+                },
+            }
+        )
+
+        # Update the privilege record and create history record
```
landonshumway-ia, line 945: we should add a log statement here with a message that we are setting the privilege to inactive.

backend/compact-connect/lambdas/python/provider-data-v1/handlers/privileges.py
```diff
@@ -0,0 +1,44 @@
...
+
+    :param event: Standard API Gateway event, API schema documented in the CDK ApiStack
+    :param LambdaContext context:
+    """
+    compact = event['pathParameters']['compact']
+    provider_id = event['pathParameters']['providerId']
+    jurisdiction = event['pathParameters']['jurisdiction']
+    # Note: We currently only support one license type per jurisdiction, so this is not used
+    # We require this in the API to avoid one breaking change when we move to multiple license types per jurisdiction
+    license_type = event['pathParameters']['licenseType']
```
landonshumway-ia, line 25: Awesome, I was going to ask about this. Glad we're including it in the api path 👍

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_privileges.py
```diff
@@ -0,0 +1,140 @@
...
+
+    @patch('cc_common.config._Config.current_standard_datetime', datetime.fromisoformat('2024-11-08T23:59:59+00:00'))
+    def test_compact_admin_can_deactivate_privilege(self):
+        """
+        If a compact admin has admin permission in the compact, they can deactivate a privilege
+        """
+        self._load_provider_data()
+
+        # The user has admin permission for aslp
+        resp = self._request_deactivation_with_scopes('openid email aslp/admin aslp/aslp.admin')
```
landonshumway-ia, line 90: I'll have to make sure to update these tests in my current PR to use the new scope pattern, assuming this gets merged first

backend/compact-connect/lambdas/python/provider-data-v1/handlers/privileges.py
```diff
@@ -0,0 +1,44 @@
...
+
+    # Get the user's scopes to check for jurisdiction-specific admin permission
+    scopes = get_event_scopes(event)
+    jurisdiction_admin_scope = f'{compact}/{jurisdiction}.{CCPermissionsAction.ADMIN}'
+    compact_admin_scope = f'{compact}/{compact}.{CCPermissionsAction.ADMIN}'
```
landonshumway-ia: this will also need to be updated to use the new scope pattern in this PR: https://github.com/csg-org/CompactConnect/pull/561

landonshumway-ia (changes requested):

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
landonshumway-ia: Or are we using a query in anticipation of supporting multiple license types? Either way given the `eq` condition on the sk this should only ever return one and only one item.

jusdino: Good find - I don't know why I went with query but get_item is definitely the better way.

landonshumway-ia (approved):

landonshumway-ia:
@jlkravitz This one is ready for your review

backend/compact-connect/docs/design/license-events-diagram.pdf
jlkravitz, line 1: image

Should the arrow from the event bus to the Ingest Stack include "license.deactivation" in addition to "license.ingest"? Mostly wondering where the "deactivate" path comes into play.

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
```diff
@@ -889,3 +890,94 @@ def find_home_state_license(self, *, compact: str, provider_id: str, licenses: l
...
+        privilege_update_record = PrivilegeUpdateRecordSchema().dump(
+            {
+                'type': 'privilegeUpdate',
+                'updateType': 'deactivation',
+                'providerId': provider_id,
+                'compact': compact,
+                'jurisdiction': jurisdiction,
+                'dateOfUpdate': self.config.current_standard_datetime.isoformat(),
+                'previous': {
+                    # We're relying on the schema to trim out unneeded fields
```
jlkravitz, line 936: Thanks for this context!

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
```diff
@@ -889,3 +890,94 @@ def find_home_state_license(self, *, compact: str, provider_id: str, licenses: l
...
+                        'Key': {
+                            'pk': {'S': f'{compact}#PROVIDER#{provider_id}'},
+                            'sk': {'S': f'{compact}#PROVIDER'},
+                        },
+                        'UpdateExpression': 'DELETE privilegeJurisdictions :jurisdiction',
+                        'ExpressionAttributeValues': {
+                            ':jurisdiction': {'SS': [jurisdiction]},
+                        },
+                    },
+                },
```
jlkravitz, line 981: What are we deleting here?

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
```diff
@@ -889,3 +890,94 @@ def find_home_state_license(self, *, compact: str, provider_id: str, licenses: l
...
+        self.config.dynamodb_client.transact_write_items(
+            TransactItems=[
+                {
+                    'Update': {
+                        'TableName': self.config.provider_table.name,
+                        'Key': {
+                            'pk': {'S': f'{compact}#PROVIDER#{provider_id}'},
+                            'sk': {'S': f'{compact}#PROVIDER#privilege/{jurisdiction}#'},
+                        },
+                        'UpdateExpression': 'SET persistedStatus = :status, dateOfUpdate = :dateOfUpdate',
```
jlkravitz, line 956: Is the idea that we're updating the `persistedStatus` of all privilege records - original upload and updates?

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/privilege/record.py
```diff
@@ -48,6 +50,13 @@ class PrivilegeRecordSchema(CalculatedStatusRecordSchema):
     attestations = List(Nested(AttestationVersionRecordSchema()), required=True, allow_none=False)
     # the human-friendly identifier for this privilege
     privilegeId = String(required=True, allow_none=False)
+    # the persisted status of the privilege, which can be manually set to inactive
+    persistedStatus = ActiveInactive(required=True, allow_none=False)
+
+    # This field is the actual status referenced by the system, which is determined by the expiration date
+    # in addition to the persistedStatus. This should never be written to the DB. It is calculated
+    # whenever the record is loaded.
```
jlkravitz, line 58: Helpful!

backend/compact-connect/lambdas/python/common/tests/function/test_data_client.py
```diff
@@ -613,3 +617,189 @@ def test_get_ssn_by_provider_id_raises_exception_multiple_records_found(self):
...
+            provider_id=provider_id,
+            jurisdiction='ne',
+        )
+
+        # Verify that the privilege record was updated
+        new_privilege = self._provider_table.query(
+            KeyConditionExpression=Key('pk').eq(f'aslp#PROVIDER#{provider_id}')
+            & Key('sk').begins_with('aslp#PROVIDER#privilege/ne#'),
+        )['Items']
+        self.assertEqual([original_privilege, original_history], new_privilege)
```
jlkravitz, line 799: Is this comment right? Aren't we verifying that `deactive_privilege` did nothing?

backend/compact-connect/lambdas/python/provider-data-v1/handlers/privileges.py
```diff
@@ -0,0 +1,63 @@
...
+                        {
+                            'eventTime': config.current_standard_datetime.isoformat(),
+                            'compact': compact,
+                            'jurisdiction': jurisdiction,
+                            'providerId': provider_id,
+                        }
+                    ),
+                    'EventBusName': config.event_bus_name,
+                }
+            )
```
jlkravitz, line 61: I keep losing track of the concept of a license vs. a privilege. How is a license deactivation different from a privilege deactivation?

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_privileges.py
```diff
@@ -0,0 +1,159 @@
...
+
+        # The user has read permission for aslp
+        resp = self._request_deactivation_with_scopes('openid email aslp/readGeneral aslp/ne.readPrivate')
+        self.assertEqual(403, resp['statusCode'])
+
+    @patch('cc_common.config._Config.current_standard_datetime', datetime.fromisoformat('2024-11-08T23:59:59+00:00'))
+    def test_deactivate_privilege_not_found(self):
+        """
+        If a privilege is not found, the response should be a 404
+        """
```
jlkravitz, line 157: nit: Can we add a brief comment here noting that we have left out `self._load_provider_data()` - took me a second to realize why the privilege would not be found.

jlkravitz (commented):
A few questions/comments, but generally looks great.

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
jlkravitz: Further reading of the code made me understand - can we just add a brief comment above this DynamoDB 'command' (and maybe all three)?

backend/compact-connect/docs/design/license-events-diagram.pdf
jusdino: There is no explicit 'deactivate a license' action in the API, so a deactivation cannot originate there. Instead, what happens is the ingest flow (stack) assesses existing licenses that are being ingested again for what changed and categorizes some changes, such as a deactivation. If a deactivation is detected, the ingest flow will trigger an additional event to announce that deactivation.

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
jusdino: We're updating the privilege record's privilegeStatus to `inactive`, creating a history record to reflect the change, and updating the provider record to remove the affected privilege jurisdiction.

backend/compact-connect/lambdas/python/provider-data-v1/handlers/privileges.py
jusdino: A **license** is issued to a practitioner by a practitioner's home state as a result of them meeting that state's requirements for licensure to practice a particular profession in that state. CompactConnect has no part in that process, other than storing **license** data for sharing across the compact, when a state uploads their own **license** data for eligible practitioners. A **privilege** is what a practitioner can buy through CompactConnect, which uses their home state **license** as a basis to justify their **privilege** to practice in other compact member states. CompactConnect _does_ administrate **privileges** on behalf of the compacts.

jusdino:
@jlkravitz , comments addressed!

jlkravitz (approved):
@isabeleliassen Good to go.

backend/compact-connect/lambdas/python/provider-data-v1/handlers/privileges.py
jlkravitz: Created a ticket #589

isabeleliassen (approved):

PR #568, opened 2025-02-19
Add dependency license report script

### LINKED ISSUE #564: compile license information

jusdino:
### Description List
- Add a script at a root `bin` folder to look for NodeJS and Python dependencies, then report on their licenses

### Testing List
- Run it!

Closes #564


bin/report_dependency_licenses.py
```diff
@@ -0,0 +1,354 @@
...
+    for dep in deps:
+        key = f'{dep.language}:{dep.name}'
+        if key not in unique_deps:
+            unique_deps[key] = dep
+        else:
+            existing = unique_deps[key]
+            # If the existing version is 'latest', prefer the specific version
+            if existing.version == 'latest' and dep.version != 'latest':
+                unique_deps[key] = dep
+            # If both have specific versions, keep the first one found
```
landonshumway-ia, line 293: not sure if this comment is accurate, since its referring to versions but line 294 is referring to location

bin/report_dependency_licenses.py
```diff
@@ -0,0 +1,354 @@
...
+    for dep in unique_deps:
+        dep.get_license_info()
+
+    # Write results to CSV
+    logger.info('Writing results to CSV')
+    with open('dependency_licenses.csv', 'w', newline='') as f:
+        writer = csv.writer(f)
+        writer.writerow(['Package Name', 'Version', 'Language', 'License', 'URL', 'Location'])
+
+        for dep in sorted(unique_deps, key=lambda x: (x.language, x.license or 'Unknown', x.name.lower(), x.version)):
```
landonshumway-ia, line 337: Is the intent for a human developer to go track down any `Unknown` license if the script is unable to find it? If so we should probably leave a comment somewhere in here stating that intent.

bin/report_dependency_licenses.py
```diff
@@ -0,0 +1,354 @@
...
+    # Get license information for each dependency
+    logger.info('Getting license information for each dependency')
+    for dep in unique_deps:
+        dep.get_license_info()
+
+    # Write results to CSV
+    logger.info('Writing results to CSV')
+    with open('dependency_licenses.csv', 'w', newline='') as f:
+        writer = csv.writer(f)
+        writer.writerow(['Package Name', 'Version', 'Language', 'License', 'URL', 'Location'])
```
landonshumway-ia, line 335: It would be really helpful to have links to both the GH url (if applicable) as well as the package repository (npm or pypi) since the dev can then go to those links and find the license if the script wasn't able to extract it.

landonshumway-ia (changes requested):
This is a good start, but I notice a handful of rows in the csv file have UNKNOWN or NOASSERTION for the license even though that information is available through GH and Pypi, for example the `publication` python module is using the MIT license, but the script was unable to find it. 

publication GH: https://github.com/glyph/publication?tab=MIT-1-ov-file
publication Pypi: https://pypi.org/project/publication/

cachecontrol (licensed under apache 2.0)
GH: https://github.com/psf/cachecontrol?tab=License-1-ov-file
Pypi: https://pypi.org/project/CacheControl/

Obviously, the script likely won't be able to find every single license, but if the intent is to have a human fill in the gaps, then we should check in the latest csv file with all the unknowns filled in so that the next time we need to run this, the human developer doesn't have to dig so much for all the unknown licenses.

bin/report_dependency_licenses.py
jusdino: Oh, ok. Yeah, this line is actually more referring to the lack of an overwrite, not really the line below. I guess it's more misleading than helpful. I'll remove it.

bin/report_dependency_licenses.py
jusdino: This should never really happen, since the `Dependency` subclasses each raise an exception if they fail to find their license. I just wanted a known value here to look for to check for bugs. I'll add a comment to that effect.

bin/report_dependency_licenses.py
```diff
@@ -1,4 +1,14 @@
 #!/usr/bin/env python3
+"""
+Script to search this project for Python and NodeJS dependencies, identify their licenses, and write the results to a
+CSV file.
+
+Note: Some dependencies don't make finding their license information very easy (shame, Python devs!), so this script does its best
+to mine PyPi and GitHub for license data. It does not find all licenses. The resulting CSV will have some NOASSERTION licenses,
+which will need manual search and entry.
```
landonshumway-ia, line 8: 😆

landonshumway-ia (approved):
Looks great! This script will come in handy as the project grows

bin/report_dependency_licenses.py
```diff
@@ -0,0 +1,385 @@
...
+        data = json.loads(result.stdout)['data']
+        license_name = data['license']
+
+        repo_url = None
+        repo_url = data.get('repository', {}).get('url')
+        # If they don't list a repository, we'll use the homepage as a fallback
+        if not repo_url:
+            repo_url = data.get('homepage', '').split('#')[0]
+
+        # As a last resort, point to npmjs.com https://www.npmjs.com/package/@usewaypoint/email-builder
```
jlkravitz, line 165: I think you mean –

```suggestion
        # As a last resort, point to npmjs.com (e.g., https://www.npmjs.com/package/@usewaypoint/email-builder)
```

bin/report_dependency_licenses.py
```diff
@@ -0,0 +1,385 @@
...
+                version = version_match.group(1) if version_match else 'latest'
+
+                deps.append(cls(parts, version, os.path.dirname(file_path)))
+
+        return deps
+
+    def get_license_info(self):
+        """Get license information for a Python package."""
+        logger.info('Getting license information for Python package %s', self.name)
+        result = subprocess.run(['pip', 'show', self.name], capture_output=True, text=True)  # noqa: S607 S603
```
jlkravitz, line 210: Does this need to include the version the same way we handle JS dependencies?

bin/report_dependency_licenses.py
```diff
@@ -0,0 +1,385 @@
...
+
+    for dep in deps:
+        key = f'{dep.language}:{dep.name}'
+        if key not in unique_deps:
+            unique_deps[key] = dep
+        else:
+            existing = unique_deps[key]
+            # If the existing version is 'latest', prefer the specific version
+            if existing.version == 'latest' and dep.version != 'latest':
+                unique_deps[key] = dep
```
jlkravitz, line 321: Similar question– If we're using both versions, it seems we'd want to track both licenses (if different).

jlkravitz (commented):
Looks good! I don't have much context on the use-case here (though I think it is a good thing to do), so not sure how accurate this needs to be. My two comments only really matter if accuracy is key (e.g., for legal compliance).

jlkravitz (approved):
@isabeleliassen Good to go.

isabeleliassen (approved):

PR #595, opened 2025-02-24
Feat/public lookup backend endpoints

### LINKED ISSUE #447: Public Lookup backend list view
- new endpoint for public lookup
- both "results" list view and detail view
- the public can look up a licensee by name, state, privilege number, or license number (if supported eventually) and view their active and past privileges, purchase and expiration dates, and if any of those licenses have discipline
- private information such as phone, email, address, DOB, SSN is not visible
- lookup separated by compact/profession
- discuss ways to prevent scraping? rate limiting? For now can use default
- need some idea of how much a real human user might look up in an hour
- anticipate using similar designs for search page and results
- no design yet for how to access the search page--need some kind of landing page ("search providers" or similar added to existing login page)
- try to keep design similar to existing designs]
- Brad recommends a URL for a provider's public page so they can link to privilege on their website (QR code on license doc could eventually link to this page)

landonshumway-ia:
We have a requirement to allow unauthenticated users to search for licensees within the system from a public lookup page. This adds two public endpoints, one for querying for a list of matching licensees, and the other for getting the public details of an individual licensee. 
```
POST /v1/public/compacts/{compact}/providers/query
GET /v1/public/compacts/{compact}/providers/{providerId}
```

These endpoints are very similar to the licensee lookup endpoints used by Staff Users, except they only return public license information (ie no personal address information). This adds public API schema classes which explicitly define which fields are allowed to be returned, and removes all others. 

### Requirements List
- This change should be fully backwards compatible with the current system, no migration changes needed

### Description List
- Added public query endpoint
- Added public get provider details endpoint
- Update api spec and postman collection to latest

### Testing List
- Code review

Closes #447


backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/provider/api.py
```diff
@@ -73,3 +76,48 @@ class ProviderGeneralResponseSchema(ForgivingSchema):
...
+    status = ActiveInactive(required=True, allow_none=False)
+
+    privilegeJurisdictions = Set(String, required=False, allow_none=False, load_default=set())
+    providerDateOfUpdate = Raw(required=False, allow_none=False)
+
+    # these records are present when getting provider information from the GET endpoint
+    # so we check for them here and sanitize them if they are present
+    licenses = List(Nested(LicensePublicResponseSchema(), required=False, allow_none=False))
+    privileges = List(Nested(PrivilegePublicResponseSchema(), required=False, allow_none=False))
+    homeJurisdictionSelection = Nested(
```
jusdino, line 121: Do we want to include this? We're explicitly telling the caller whether or not this user is already registered in the system if we provide this value.

backend/compact-connect/stacks/api_stack/v1_api/api_model.py
```diff
@@ -1242,6 +1232,276 @@ def get_provider_ssn_response_model(self) -> Model:
...
+                'jurisdictionStatus',
+                'givenName',
+                'familyName',
+                'dateOfIssuance',
+                'dateOfRenewal',
+                'dateOfExpiration',
+                'status',
+            ],
+            properties={
+                'type': JsonSchema(type=JsonSchemaType.STRING, enum=['license-home']),
```
jusdino, line 1379: I'm assuming that this was a deliberate choice but I'd be interested to hear your thoughts on duplicating all of the license properties here, as well as in the `_common_license_properties` as opposed to setting up a hierarchy of properties, like something along the lines of `public < common < private`. I'm finding that keeping these models up to date is a bit unwieldy.

jusdino (approved):
Looks great! Just a couple of questions:

backend/compact-connect/stacks/api_stack/v1_api/api_model.py
landonshumway-ia: I hesitate to combine these schemas as IMO it increases the risk of properties being set in certain schemas even if we are not returning them. The idea is if I update a specific endpoint, I just need to update that specific API model associated with it, whereas if I couple them and there are big change I now need to redesign how the hierarchy is sharing stuff. Both setups have their drawbacks, especially with schemas this large. I'm ok with either approach, but if we do try to set up a hierarchy I'll have to walk through and get everything lined out. This is the same reason I separated the public marshmallow API schemas into their own classes.

jusdino: Understood. I don't have a strong opinion here, just wanting to think things through 👍

landonshumway-ia:
@jlkravitz This is ready for your review. Thanks!

landonshumway-ia:
@jlkravitz The requirements for this ticket have just changed based on feedback from the Compact EDs, I am pulling this ticket back to in progress to implement those new requirements.

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
```diff
@@ -969,20 +988,6 @@ def deactivate_privilege(self, *, compact: str, provider_id: str, jurisdiction:
...
-                        'Key': {
-                            'pk': {'S': f'{compact}#PROVIDER#{provider_id}'},
-                            'sk': {'S': f'{compact}#PROVIDER'},
-                        },
-                        'UpdateExpression': 'DELETE privilegeJurisdictions :jurisdiction',
-                        'ExpressionAttributeValues': {
-                            ':jurisdiction': {'SS': [jurisdiction]},
-                        },
-                    },
-                },
```
jlkravitz, line 986: I assume this is the case, but confirming that we don't have any data in production that will make this a breaking change?

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/provider/api.py
```diff
@@ -73,3 +73,36 @@ class ProviderGeneralResponseSchema(ForgivingSchema):
...
+    npi = NationalProviderIdentifier(required=False, allow_none=False)
+    givenName = String(required=True, allow_none=False, validate=Length(1, 100))
+    middleName = String(required=False, allow_none=False, validate=Length(1, 100))
+    familyName = String(required=True, allow_none=False, validate=Length(1, 100))
+    suffix = String(required=False, allow_none=False, validate=Length(1, 100))
+    status = ActiveInactive(required=True, allow_none=False)
+
+    privilegeJurisdictions = Set(String, required=False, allow_none=False, load_default=set())
+    # We only return privileges when getting provider information from the public GET endpoint
+    # so we check for them here and sanitize them if they are present
```
jlkravitz, line 107: What does this mean?

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/__init__.py
```diff
@@ -155,7 +155,9 @@ def privilege_jurisdictions_to_set(obj: dict):
             else:
                 self._provider_table.put_item(Item=record)
 
-    def _generate_providers(self, *, home: str, privilege: str, start_serial: int, names: tuple[tuple[str, str]] = ()):
+    def _generate_providers(
+        self, *, home: str, privilege: str | None, start_serial: int, names: tuple[tuple[str, str]] = ()
```
jlkravitz, line 159: `privilege` is a little vague - perhaps `privilege_jurisdiction`?

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_public_lookup.py
```diff
@@ -0,0 +1,387 @@
...
+
+        body = json.loads(resp['body'])
+        self.assertEqual(10, len(body['providers']))
+        self.assertEqual({'providers', 'pagination', 'query', 'sorting'}, body.keys())
+        self.assertIsInstance(body['pagination']['lastKey'], str)
+        # Check we're actually sorted
+        dates_of_update = [provider['dateOfUpdate'] for provider in body['providers']]
+        self.assertListEqual(sorted(dates_of_update), dates_of_update)
+
+    def test_public_query_providers_updated_sorting_filters_providers_without_privileges(self):
```
jlkravitz, line 86: can we add a quick comment on this test that clarifies why testing "without privileges" is relevant?

e.g., 

```"""Tests that the public endpoint only returns providers with matching privileges, not licenses."""```

File: backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_public_lookup.py [hunk 7]
```diff
@@ -0,0 +1,387 @@
...
+            privilege='ne',
+            start_serial=9999,
+            names=(('Testerly', 'Tess'), ('Testerly', 'Ted')),
+        )
+
+        with open('../common/tests/resources/api-event.json') as f:
+            event = json.load(f)
+
+        # The user has read permission for aslp
+        event['requestContext']['authorizer']['claims']['scope'] = 'openid email aslp/readGeneral'
```
jlkravitz, line 173: Isn't this a public endpoint that doesn't need authorization? If this is intentional, can we update the test name and add a docstring comment to reflect that?

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_public_lookup.py
```diff
@@ -0,0 +1,387 @@
...
+            privilege=None,
+            start_serial=9999,
+            names=(('Testerly', 'Tess'), ('Testerly', 'Ted')),
+        )
+
+        with open('../common/tests/resources/api-event.json') as f:
+            event = json.load(f)
+
+        # The user has read permission for aslp
+        event['requestContext']['authorizer']['claims']['scope'] = 'openid email aslp/readGeneral'
```
jlkravitz, line 212: Same question here.

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_public_lookup.py
```diff
@@ -0,0 +1,387 @@
...
+
+    def test_public_get_provider_missing_provider_id(self):
+        from handlers.public_lookup import public_get_provider
+
+        with open('../common/tests/resources/api-event.json') as f:
+            event = json.load(f)
+
+        # public endpoint does not have authorizer
+        del event['requestContext']['authorizer']
+        # providerId _should_ be included in these pathParameters. We're leaving it out for this test.
```
jlkravitz, line 381: Good call-out.

jlkravitz (commented):
Looks good! Few questions.

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
landonshumway-ia: We shouldn't, I doubt anyone has even called this deactivation endpoint in production.

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_public_lookup.py
landonshumway-ia: Good catch, that is a copy/paste error from the other search tests which require authorization

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_public_lookup.py
landonshumway-ia: same, copy/paste error

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/provider/api.py
landonshumway-ia: I've reworded this to hopefully make more sense

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
landonshumway-ia: Even still, this would not be a breaking change, since we are not adding new restrictions, but rather loosening restrictions so that the `privilegeJurisdictions` set now includes both active and inactive privileges.

jlkravitz: Got it. Was just thinking about the case where this is used for search or elsewhere. But, regardless, sounds like it's a moot point.

jlkravitz (approved):
Looks great! Thanks for the comment rewrite- it reads really well!

isabeleliassen (approved):

PR #606, opened 2025-02-26
Frontend/Fix license list edge cases

### LINKED ISSUE #604: License list page advance after detail view
### Notes
After navigating into a detail view and then returning to the list view, the page advance button may not actually advance the page.

### Tasks
- [x] Fix workflow described above
- [x] Cross browser device testing
- [x] Confirm `yarn build` is error & warning free
- [x] Confirm `yarn serve` is error & warning free
- [x] Confirm `yarn test:unit:all` is error & warning free
- [x] Confirm `yarn test:coverage` is error & warning free
- [x] QA Notes added

### QA Notes
1. As Staff user, perform a license search that will result in multiple pages of licensees
1. Click on a result row to be navigated to the licensee detail page
1. Return to the list page
1. Attempt to advance to next pages
1. Regression: Make sure full license search / results / view flow still works stably

jsandoval81:
### Requirements List
- _None_

### Description List
- Fixed page-advance breaking after detail view
- Better handle edge case of cache getting cleared on detail view

### Testing List
- Code review
- License search
    - As Staff user, perform a license search that will result in multiple pages of licensees
    - Click on a result row to be navigated to the licensee detail page
    - Return to the list page
    - Attempt to advance to next pages
    - Regression: Make sure full license search / results / view flow still works stably

Closes #604

### DIFF
--- a/webroot/src/components/Licensee/LicenseeList/LicenseeList.ts
+++ b/webroot/src/components/Licensee/LicenseeList/LicenseeList.ts
@@ -46,15 +46,22 @@ class LicenseeList extends Vue {
     // Lifecycle
     //
     async created() {
-        if (this.licenseStore.model?.length) {
+        if (this.licenseStoreRecordCount) {
             this.hasSearched = true;
         }
     }
 
     async mounted() {
-        if (!this.licenseStore.model?.length) {
+        if (!this.licenseStoreRecordCount) {
+            // License store is empty - apply defaults
             await this.setDefaultSort();
             await this.setDefaultPaging();
+        } else if (this.licenseStoreRecordCount === 1 && !this.searchDisplayAll) {
+            // Edge case: Returning from a detail page that was refreshed / cache-cleared
+            this.shouldShowSearchModal = true;
+        } else {
+            // License store already has records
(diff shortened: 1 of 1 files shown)


ChiefStief (approved):

jsandoval81:

jlkravitz:
@jsandoval81 Looks good. One question– if I'm reading the code correctly, every time a user clicks on a detail page, then navigates back, they will be brought to the search page instead of the results page that they were navigating. True? Probably not in scope for this PR, but isn't that a bit unintuitive?

jsandoval81:
@jlkravitz That will be true, but only in cases where the user explicitly clears their cache while on the detail page.

Previously that edge case would land them back at a list view with only 1 row (the 1 record they had in their cache from the detail page). That was a lot more of an ambiguous state. In that edge case, this PR now just do what the list view does if the user refreshes, which is send them back to the search page. This felt more consistent, and also gives us a more stable point to progressively enhance the caching in the future (if ever needed).

jlkravitz (approved):
@isabeleliassen This is good to merge _after_ #596 is merged.

isabeleliassen (approved):

PR #623, opened 2025-03-06
Set explicit export of SSN key arn to address CDK cross-stack dep

landonshumway-ia:
We removed reference to this KMS key in the ingest stack, but due to CDK's handling of cross-stack deps we need to explicitly export the key arn until the change has gone through all envs of the system


jlkravitz (approved):

landonshumway-ia:
@jlkravitz Sorry, Justin ran this locally and found the S3 bucket export also needs to be set explicitly as a result of this change. If you wouldn't mind reviewing again. Thanks

jlkravitz (approved):

isabeleliassen (approved):

PR #626, opened 2025-03-07
Feat/ssn alarms

### LINKED ISSUE #393: Add monitoring/alerting controls to full SSN endpoint (backend)
Depends on #392 

- Produce a metric for each request, including staff user id as a dimension
- Produce a daily count metric for total endpoint usage across all users
- Configure anomaly detection for the daily count metric
- Configure a per-user daily request rate alarm with a constant limit
- Configure an alarm on anomaly detection for the daily count
- Configure alarms to notify operation support staff

landonshumway-ia:
The most sensitive endpoint in this system is the GET provider SSN endpoint, which returns full SSNs of providers and can only be accessed by admins which have been granted the readSSN permission. There is still potential risk for abuse of this endpoint if an admin's credentials are compromised, and we need proactive measures in place to protect this sensitive data. This change adds several security enhancements to the system for alerting and throttling on excessive requests made to this endpoint. For now, we have set a conservative threshold of 5 requests per user per day, and 15 requests across all user per day. If an individual user exceeds this threshold and continues to make requests after being throttled, we deactivate their Cognito account and fire an alert.

If the max 15 requests per day is exceeded, we disable the endpoint altogether, requiring developer intervention to re-enable the endpoint after it has been determined if the system is being abused.

As part of this effort, we identified another attack vector for abuse, were a bad actor could use an admin's compromised credentials to create more admins with read ssn permissions to read many SSNs. With the global throttling now in place, the risk is now reduced, but we still want to be aware if excessive staff users are being created within the system, so this change adds some alarms around that endpoint as well.

### Requirements List
- This change is backwards compatible with the current system, no new requirements or migrations needed

### Description List
- Added rate-limiting logic to read ssn endpoint to throttle excessive requests
- Add alarms/alerting to report on ssn requests that exceed set thresholds
- Add alarms/alerting to POST staff users endpoint to notify if excessive staff users are being created

### Testing List
- Added smoke test to verify throttling logic works in sandbox environment
- Added functional/unit test coverage for read ssn lambda logic
- Code review

Closes #393


ChiefStief (approved):

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_providers.py
```diff
@@ -431,3 +462,75 @@ def test_get_provider_ssn_forbidden_without_correct_jurisdiction_level_scope(sel
...
+        """
+        The staff user has called this endpoint more than the set limit, so the endpoint returns a 429
+        If the user makes another request after this, their account should be deactivated and an alert should be
+        fired.
+        """
+        self._load_provider_data()
+        from handlers.providers import get_provider_ssn
+
+        test_provider_id = '89a6377e-c3a5-40e5-bca5-317ec854c570'
+        # add 5 previous calls to the endpoint
```
jusdino, line 476: 4 calls

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_providers.py
```diff
@@ -431,3 +462,75 @@ def test_get_provider_ssn_forbidden_without_correct_jurisdiction_level_scope(sel
...
+
+        with open('../common/tests/resources/api-event.json') as f:
+            event = json.load(f)
+            event['requestContext']['authorizer']['claims']['sub'] = staff_user_id
+
+        # The user has read permission for aslp
+        event['requestContext']['authorizer']['claims']['scope'] = 'openid email aslp/readGeneral aslp/readSSN'
+        event['pathParameters'] = {'compact': 'aslp', 'providerId': test_provider_id}
+
+        # the first request should succeed, being right at the limit
```
jusdino, line 487: `# the fifth request`, right?

backend/compact-connect/stacks/api_stack/v1_api/query_providers.py
```diff
@@ -283,18 +310,46 @@ def _add_get_provider_ssn(
             threshold_metric_id='ad1',
         )
 
-        # We'll also set a flat maximum access rate of 5 reads per hour to alarm on
+        # We'll also set a flat maximum access rate of 5 reads per day to alarm on
```
jusdino, line 313: 5 sounds a bit low for this. We don't want to encourage alarm fatigue by having an alarm that triggers well before action is required. I'm trying to think, from an operator's perspective, what should they look for to identify something suspicious before we hit our 15-call automatic intervention?

backend/compact-connect/stacks/api_stack/v1_api/query_providers.py
```diff
@@ -283,18 +310,46 @@ def _add_get_provider_ssn(
...
             comparison_operator=ComparisonOperator.GREATER_THAN_THRESHOLD,
             treat_missing_data=TreatMissingData.NOT_BREACHING,
-            alarm_description=(f'{self.api.node.path} max ssn reads alarm'),
+            alarm_description=f'{self.api.node.path} max ssn reads alarm. The GET provider SSN endpoint has been '
+            f'invoked more than an expected threshold within a 24 hour period. Investigation is '
+            f'required to ensure access is not the result of abuse.',
         )
         self.max_ssn_reads_alarm.add_alarm_action(SnsAction(self.api.alarm_topic))
+        # Add an alarm for 429 responses from the SSN endpoint
+        self.ssn_api_throttling_alarm = Alarm(
```
jusdino, line 329: Because this metric is generated for _any_ 4XX response, having it alarm on any metrics at all is likely to have a lot of false-positives, since any random unauthenticated bot call to this method will set off an alarm. We should probably dial this threshold up enough to filter out the random crawler traffic. We already alarm on our own internal throttling, so this would be a backup to detect authn/authz failures that would otherwise be missed. I think any legitimate attack vector caught by this alarm would need to have a pretty high call volume to have a chance at success, so maybe we bump it up to like 100 in 5 mins?

backend/compact-connect/stacks/api_stack/v1_api/staff_users.py
```diff
@@ -488,6 +497,86 @@ def _add_post_user(
...
+        self.max_daily_staff_users_created_alarm = Alarm(
+            self.api,
+            'MaxDailyStaffUserCreatedAlarm',
+            metric=staff_user_created_daily_count_metric,
+            threshold=20,
+            evaluation_periods=1,
+            comparison_operator=ComparisonOperator.GREATER_THAN_THRESHOLD,
+            treat_missing_data=TreatMissingData.NOT_BREACHING,
+            alarm_description=f'{self.api.node.path} max daily staff users created alarm. The POST staff user endpoint '
+            f'has been invoked more than an expected threshold within an hour period. '
```
jusdino, line 543: within a day

backend/compact-connect/lambdas/python/provider-data-v1/handlers/providers.py
```diff
@@ -153,3 +171,111 @@ def get_provider_ssn(event: dict, context: LambdaContext):  # noqa: ARG001 unuse
...
+                current_request_compact=compact,
+            )
+
+            # Set the lambda's reserved concurrency to 0 to throttle the endpoint
+            try:
+                config.lambda_client.put_function_concurrency(
+                    FunctionName=config.current_lambda_name, ReservedConcurrentExecutions=0
+                )
+                logger.critical('Lambda concurrency set to 0 due to excessive SSN requests')
+                metrics.add_metric(name='ssn-endpoint-throttled', value=1, unit='Count')
```
jusdino, line 236: Where is the alarm for this metric?

backend/compact-connect/tests/smoke/ssn_read_throttling_smoke_tests.py
```diff
@@ -0,0 +1,224 @@
...
+    Checks if a user is enabled or disabled in Cognito
+    Returns True if enabled, False if disabled
+    """
+    from config import config
+
+    user_data = config.cognito_client.admin_get_user(UserPoolId=config.cognito_staff_user_pool_id, Username=email)
+    return user_data.get('Enabled', True)
+
+
+def trigger_get_provider_ssn_endpoint_throttling():
```
jusdino, line 81: Nice!

jusdino (changes requested):
Nice work. Just a few tweaks, I think:

backend/compact-connect/lambdas/python/provider-data-v1/handlers/providers.py
```diff
@@ -153,3 +171,111 @@ def get_provider_ssn(event: dict, context: LambdaContext):  # noqa: ARG001 unuse
...
+                'Global SSN rate limit exceeded, throttling endpoint',
+                global_request_count=global_request_count,
+                current_request_user_id=user_id,
+                current_request_compact=compact,
+            )
+
+            # Set the lambda's reserved concurrency to 0 to throttle the endpoint
+            try:
+                config.lambda_client.put_function_concurrency(
+                    FunctionName=config.current_lambda_name, ReservedConcurrentExecutions=0
```
jusdino, line 233: You can get this from the `context` object as well, FWIW:
https://docs.aws.amazon.com/lambda/latest/dg/python-context.html

backend/compact-connect/lambdas/python/provider-data-v1/handlers/providers.py
```diff
@@ -153,3 +171,111 @@ def get_provider_ssn(event: dict, context: LambdaContext):  # noqa: ARG001 unuse
...
+    :return: True if rate limit is exceeded, False otherwise
+    """
+    now = config.current_standard_datetime
+    window_start = now - timedelta(hours=24)
+    window_start_timestamp = window_start.timestamp()
+    now_timestamp = now.timestamp()
+
+    # Generate a unique ID for this request
+    # This ensures every request is recorded, even for requests within the same second
+    request_sk = f'TIME#{now_timestamp}#UUID#{uuid.uuid4()}'
```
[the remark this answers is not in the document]
landonshumway-ia: That makes sense, my only concern with this approach is I believe it's possible for a client to override this value through the headers it sends to the endpoint, which could present an attack vector where the caller would hit the endpoint with many requests, all with the same request id and within the same second. I'm trying to dig through AWS documentation to clarify if this is true, but I can't find any definitive answer. 

As an alternative, we could simply log the SK as we are creating the record, which would give us traceability in the logs. Curious what your thoughts are with this approach.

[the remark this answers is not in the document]
landonshumway-ia: Looks like there is also a `aws_request_id` we can fetch from the lambda context itself, which can not be overridden by a client, so we can leverage this since I'm going to use the context anyway to get the name of the lambda function.

backend/compact-connect/stacks/api_stack/v1_api/query_providers.py
landonshumway-ia: Well, if we are to treat the max threshold of 15 as a capacity metric, we could alert if we go over 80% of our max expected volume, so set the threshold to 12? Or honestly we just get rid of this particular alert all together now that we have the other alerts in place.

landonshumway-ia: Actually, I forgot I added this metric `rate-limited-ssn-access` which increments if anyone hits the threshold and is rate limited. We could replace this alert for another that uses this metric.

backend/compact-connect/lambdas/python/provider-data-v1/handlers/providers.py
landonshumway-ia: It's now been added

backend/compact-connect/stacks/api_stack/v1_api/query_providers.py
```diff
@@ -283,18 +310,79 @@ def _add_get_provider_ssn(
...
+            alarm_description='SECURITY ALERT: SSN ENDPOINT DISABLED. The GET provider SSN endpoint has been disabled '
+                              'due to excessive requests. Immediate investigation required. Endpoint will need to be '
+                              'manually reactivated before any further requests can be processed.',
+        )
+        self.ssn_endpoint_disabled_alarm.add_alarm_action(SnsAction(self.api.alarm_topic))
+
+        # Add an alarm for 429 responses from the SSN endpoint
+        self.ssn_api_throttling_alarm = Alarm(
+            self.api,
+            'SSNApi429ThrottlingAlarm',
```
jusdino, line 364: We should probably call this something more like `SSNApi4XXAlarm`, since it isn't the alarm for throttling.

jusdino (changes requested):
A couple quick things, then let's call it good 👍

jusdino (approved):

jusdino:
@jlkravitz , this is ready for you!

backend/compact-connect/stacks/api_stack/v1_api/query_providers.py
```diff
@@ -283,18 +312,80 @@ def _add_get_provider_ssn(
...
+            comparison_operator=ComparisonOperator.GREATER_THAN_OR_EQUAL_TO_THRESHOLD,
+            treat_missing_data=TreatMissingData.NOT_BREACHING,
+            alarm_description=f'{self.api.node.path} SECURITY ALERT: SSN ENDPOINT DISABLED. The GET provider SSN '
+                              'endpoint has been disabled due to excessive requests. Immediate investigation required. '
+                              'Endpoint will need to be manually reactivated before any further requests can be '
+                              'processed.',
+        )
+        self.ssn_endpoint_disabled_alarm.add_alarm_action(SnsAction(self.api.alarm_topic))
+
+        # Add an alarm for 429 responses from the SSN endpoint
```
jlkravitz, line 364: Should this be `4xx`? Can we add a quick comment re: why we're tracking 4xx generally vs just 429?

backend/compact-connect/stacks/api_stack/v1_api/staff_users.py
```diff
@@ -488,6 +497,86 @@ def _add_post_user(
...
+        # Also create a daily metric
+        staff_user_created_daily_count_metric = Metric(
+            namespace='compact-connect',
+            metric_name='staff-user-created',
+            statistic='SampleCount',
+            period=Duration.days(1),
+            dimensions_map={'service': 'common'},
+        )
+
+        # Setting a flat rate of 20 Staff users created per day to alarm on
```
jlkravitz, line 533: I imagine this will be higher on some days- e.g., onboarding new states. Is the goal here to keep it low and adjust based on what we see in the real world?

backend/compact-connect/stacks/api_stack/v1_api/staff_users.py
```diff
@@ -488,6 +497,86 @@ def _add_post_user(
...
+            evaluation_periods=1,
+            comparison_operator=ComparisonOperator.GREATER_THAN_THRESHOLD,
+            treat_missing_data=TreatMissingData.NOT_BREACHING,
+            alarm_description=f'{self.api.node.path} max daily staff users created alarm. The POST staff user endpoint '
+            f'has been invoked more than an expected threshold within a day. '
+            f'Investigation is required to ensure requests are authorized.',
+        )
+        self.max_daily_staff_users_created_alarm.add_alarm_action(SnsAction(self.api.alarm_topic))
+
+        # We'll monitor longer access patterns to detect anomalies, over time
```
jlkravitz, line 548: Can we make this more specific re: what kinds of anomalies this is monitoring? e.g., "days where the # of created staff users is more than most other days", or is it weeks, windows of time, etc.

webroot/src/pages/LicensingDetail/LicensingDetail.ts
```diff
@@ -337,15 +337,13 @@ export default class LicensingDetail extends Vue {
 
         const { compact, licenseeId } = this;
         let isError = false;
-        const ssnFullResponse = await dataApi.getLicenseeSsn(compact, licenseeId).catch(() => {
+        const ssnFullResponse = await dataApi.getLicenseeSsn(compact, licenseeId).catch((error) => {
             isError = true;
-            this.licenseeFullSsnError = this.$t('serverErrors.networkError');
+            this.licenseeFullSsnError = error?.response?.data?.message || this.$t('serverErrors.networkError');
```
jlkravitz, line 342: What precautions do we need to take (or do we already take) to ensure this server-side error message does not contain sensitive data or information?

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_providers.py
```diff
@@ -431,3 +468,73 @@ def test_get_provider_ssn_forbidden_without_correct_jurisdiction_level_scope(sel
         resp = get_provider_ssn(event, self.mock_context)
 
         self.assertEqual(403, resp['statusCode'])
+
+    def test_get_provider_ssn_throttled_and_deactivated_if_staff_user_goes_beyond_rate_limit(self):
+        """
+        The staff user has called this endpoint more than the set limit, so the endpoint returns a 429
+        If the user makes another request after this, their account should be deactivated and an alert should be
+        fired.
```
jlkravitz, line 476: Small suggestion to wording here- from reading, my assumption was that the last request was not actually part of this test.

```suggestion
        The staff user has called this endpoint more than the set limit, so the endpoint returns a 429
        and, after one more request, their account is deactivated and an alert is
        fired.
```

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_providers.py
```diff
@@ -431,3 +468,73 @@ def test_get_provider_ssn_forbidden_without_correct_jurisdiction_level_scope(sel
...
+        self.assertEqual(429, resp['statusCode'])
+        # assert that the user's account has not been deactivated yet.
+        user = self.config.cognito_client.admin_get_user(UserPoolId=self.config.user_pool_id, Username=staff_user_id)
+        self.assertEqual(user['Enabled'], True)
+
+        # make another request to trigger deactivation
+        resp = self._make_ssn_request_with_unique_request_id(event)
+        self.assertEqual(429, resp['statusCode'])
+
+        # assert that the user's account has been deactivated.
```
jlkravitz, line 506: are we testing that an alert is fired too? if not, perhaps we update the function comment

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_providers.py
```diff
@@ -431,3 +468,73 @@ def test_get_provider_ssn_forbidden_without_correct_jurisdiction_level_scope(sel
...
+        user = self.config.cognito_client.admin_get_user(UserPoolId=self.config.user_pool_id, Username=staff_user_id)
+        self.assertEqual(user['Enabled'], False)
+
+    @patch('handlers.providers.config.lambda_client', autospec=True)
+    def test_get_provider_ssn_sets_reserved_concurrency_to_zero_and_deactivated_if_staff_user_goes_beyond_rate_limit(
```
jlkravitz, line 511:
```suggestion
    def test_get_provider_ssn_endpoint_throttled_if_all_staff_users_exceed_rate_limit(
```

backend/compact-connect/tests/smoke/ssn_read_throttling_smoke_tests.py
```diff
@@ -0,0 +1,224 @@
...
+    """
+    Verifies that the GET provider SSN endpoint will throttle and deactivate users that call the
+    endpoint too frequently.
+
+    Step 1: Create three test staff users with the aslp/readSSN scope.
+    Step 2: Have each user call the endpoint until throttled (6 requests each). The first two should be disabled
+    (asserted with the AdminGetUser api), and the last one should cause the lambda to throttle itself with a set
+    reserved concurrency limit of 0 (asserted using the boto3 lambda client
```
jlkravitz, line 89:
```suggestion
    reserved concurrency limit of 0 (asserted using the boto3 lambda client)
```

jlkravitz (changes requested):
Several nits, but generally looks good. Glad we're monitoring this closely!

One question: Have we discussed integrating MFA into the login system? I'm glad we're monitoring for the case where a staff member's credentials are compromised, but the risk of that happening in the first place would drop significantly with MFA in place. I might even require security keys or passkeys for staff with `readSSN` access to eliminate the risk of phishing attacks, which are likely to be the main source of compromised credentials, though SMS or standard 2FA would be better than nothing.

Another situation I could see happening– a staff user is invited but there's a typo in the email. The user gets access to the system. I'm not sure there's much we can do about that besides closely monitoring the read SSN endpoint (as we do here) and perhaps some UX/UI improvements to confirm the email is for the right person. We might also consider creating a staff user domain whitelist, such that only individuals with email addresses in one of those whitelisted domains can be invited as staff users.

Just food for thought!

backend/compact-connect/stacks/api_stack/v1_api/query_providers.py
landonshumway-ia: Unfortunately, Amazon's api gateway metrics do not include specific status codes, just general 4xx errors. This alarm's primary intent is for the case where we are getting a high number of throttles, but I'll update the comment to reflect 4xx because that is what it is actually checking.

backend/compact-connect/stacks/api_stack/v1_api/staff_users.py
landonshumway-ia: Yes, the primary goal is just to have visibility into the case where a high volume of users are being added to the system. When we are first onboarding new states, it is expected that this alarm will fire, and we will verify that the behavior is authorized and not the result of abuse. Over time, it is expected that this alert will not fire unless there is abuse taking place in the system.

webroot/src/pages/LicensingDetail/LicensingDetail.ts
jusdino: We have a pretty deliberate exception handling strategy around this. If any unexpected exceptions are raised, our `@api_handler` decorator will not pass any dynamic information back to the client. We log then let exceptions fly in the lambda, which causes API Gateway to return a generic `502` response. In some 4XX cases, we do explicitly set the response messages in the exceptions we raise, so that we can send meaningful feedback to the client. See `api_handler` in `backend/compact-connect/lambdas/python/common/cc_common/utils.py` for details.

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_providers.py
landonshumway-ia: The alert is triggered by CloudWatch, which is outside of the purview of the lambda. We have additional CDK testing to ensure the alerts are configured as expected.

backend/compact-connect/tests/smoke/ssn_read_throttling_smoke_tests.py
```diff
@@ -0,0 +1,224 @@
...
+        f'test-staff-user-1-{uuid.uuid4()}@example.com',
+        f'test-staff-user-2-{uuid.uuid4()}@example.com',
+        f'test-staff-user-3-{uuid.uuid4()}@example.com',
+    ]
+
+    test_user_subs = []
+
+    # Create staff users with permission to read SSNs
+    for email in test_emails:
+        user_sub = create_test_staff_user(
```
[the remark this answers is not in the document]
landonshumway-ia: sub is another term for the user's id (see the 'sub' section in https://docs.aws.amazon.com/cognito/latest/developerguide/amazon-cognito-user-pools-using-the-id-token.html#user-pool-id-token-payload) This is a common term used by IDPs.

backend/compact-connect/tests/smoke/ssn_read_throttling_smoke_tests.py
```diff
@@ -0,0 +1,224 @@
...
+                response = _make_ssn_request(email, test_provider_id)
+                # We expect a 404 since the provider ID doesn't exist, but that's fine
+                # The important thing is that it's not a 429
+                if response.status_code == 429:
+                    raise SmokeTestFailureException(
+                        f'User {email} was throttled on request {j + 1}, expected to succeed'
+                    )
+                logger.info(f'Request {j + 1} successful with status code {response.status_code}')
+                # Small delay to ensure requests are recorded properly
+                time.sleep(0.5)
```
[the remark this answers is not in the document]
landonshumway-ia: After reviewing this a second time, since the the query in the lambda is using a consistent read, this actually isn't needed. Good call out, I'll remove it.

backend/compact-connect/tests/smoke/ssn_read_throttling_smoke_tests.py
```diff
@@ -0,0 +1,224 @@
...
+    return user_data.get('Enabled', True)
+
+
+def trigger_get_provider_ssn_endpoint_throttling():
+    """
+    Verifies that the GET provider SSN endpoint will throttle and deactivate users that call the
+    endpoint too frequently.
+
+    Step 1: Create three test staff users with the aslp/readSSN scope.
+    Step 2: Have each user call the endpoint until throttled (6 requests each). The first two should be disabled
```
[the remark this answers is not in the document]
landonshumway-ia: Yes, I had originally had it set up to make 6 from each, but its now actually 7 requests from the first 2 users so they are deactivated, and 2 more from the third user. I'll update the comment

webroot/src/pages/LicensingDetail/LicensingDetail.ts
jlkravitz: Great. Thanks for the explanation!

backend/compact-connect/tests/smoke/ssn_read_throttling_smoke_tests.py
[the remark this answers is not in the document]
jlkravitz: Ah, yes. Forgot about that. Thanks!

backend/compact-connect/stacks/api_stack/v1_api/staff_users.py
jlkravitz: Thanks for the update on this one! Comment is very helpful.

backend/compact-connect/stacks/api_stack/v1_api/staff_users.py
jlkravitz: Got it.

jlkravitz (approved):
@jusdino Re: MFA, ah yes. I am so used to my sandbox environment that I forgot that it's set to disable MFA. Good reminder!

re: typos, a UI/UX update makes sense. Perhaps we can discuss next week ways to make the `readSSN` permission more difficult to assign to arbitrary staff users (Perhaps you cannot assign that permission until the user is active in the system, or assigning the permission requires two staff to 'approve', or only whitelisted domains can be assigned those permissions, etc. I'm sure there's a simpler approach that encompasses most of this.)

Flagging for @isabeleliassen as something to discuss on a future stand-up. (Not sure of appropriate process here – e.g., if I'm supposed to create a ticket for this given it requires additional discussion?)

Regardless, @isabeleliassen This is good to merge!

jlkravitz (approved):

isabeleliassen (approved):

PR #631, opened 2025-03-11
Feat/multi license core

### LINKED ISSUE #569: Multiple Licenses - update core data mode
We need to update our core data model to support multiple license types per jurisdiction. Primarily, this means including a license type abbreviation in record sort keys.

Tasks:

- Define license types and abbreviations in cdk.json (already done)
- Update CDK and lambda code for new compact data structure (already done)
- Add license type to sort keys for license/privilege/update records
- Remove license type from provider record
- Update ingest logic to handle multiple license types

jusdino:
### Description List
- Revised core data model to support a practitioner having more than one license/privilege licenseType per state
- Added data migration to update existing provider records to new format
- Enhanced how python tests are run to fix an issue with code coverage under-reporting

Closes #569


backend/compact-connect/lambdas/python/provider-data-v1/handlers/ingest.py
```diff
@@ -116,62 +115,75 @@ def ingest_license_message(message: dict):
...
+                privilege_records = []
 
-            # Which license do we use for provider data?
-            # If at least one active: last issued active license
-            # If all inactive: last issued inactive license
             # Set (or replace) the posted license for its jurisdiction
-            existing_license = licenses.get(posted_license_record['jurisdiction'])
+            existing_license = licenses_organized.get(posted_license_record['jurisdiction'], {}).get(
+                posted_license_record['licenseType']
+            )
```
landonshumway-ia, line 157: Is it possible that a license record could have trailing whitespace in the csv file? I'm not sure if our ingest pipeline is currently stripping our string values of this potential whitespace. This would be another reason to be validating that the licenseType value matches one of our expected types.

On a somewhat related note, case sensitivity. Since we do a string equality == comparison when matching these license types in the privilege purchase flow, I imagine we would want to ensure these are being stored lowercase. Also if were were to ever create a GSI on this field, we would want the search on the GSI to be case insensitive (though honestly we would likely introduce a separate GSI field for that if we crossed that bridge) At any rate, I think we need to add some logic to the ingest pipeline to ensure these license types are stored in a standardized way so we can ensure that the equality check matches up as expected downstream.

landonshumway-ia (changes requested):
This is looking really good. The main thing I think is missing right now is standardization in how we are storing these string values, which will be critical since we are performing equality checks on them.

backend/compact-connect/lambdas/python/provider-data-v1/handlers/ingest.py
jusdino: Yep, licenseType is validated against the exact values in our `cdk.json` on ingest, which are all lowercase.

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/license/record.py
```diff
@@ -46,7 +46,8 @@ class LicenseRecordSchema(CalculatedStatusRecordSchema, LicenseCommonSchema):
     @pre_dump
     def generate_pk_sk(self, in_data, **kwargs):  # noqa: ARG001 unused-argument
         in_data['pk'] = f'{in_data["compact"]}#PROVIDER#{in_data["providerId"]}'
-        in_data['sk'] = f'{in_data["compact"]}#PROVIDER#license/{in_data["jurisdiction"]}#'
+        license_type_abbr = config.license_type_abbreviations[in_data['compact']][in_data['licenseType']]
```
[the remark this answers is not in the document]
jusdino: We could do that, though I was planning to just add the same hook to the privilege records so they were fully-validated by the compact as well. The benefit of a `LicenseType` field at that point would mostly just be readability/documentation.

backend/compact-connect/lambdas/python/provider-data-v1/handlers/ingest.py
landonshumway-ia: And whitespace? Are we ensuring that is being stripped from the incoming data somewhere before we run it through the validation? Is that a valid concern with this csv files?

jusdino: We're ensuring that the value after parsing from the CSV exactly matches our allowed values. I'm honestly not positive whether there's any automatic stripping of values in our CSV parser, given the many dialects of CSV, but ultimate, if the value doesn't land in the Schema exactly matching, it'll be a problem with how the CSV file was generated. If whitespace becomes a common issue with states uploading files, we could look at adding some flexibility there but I think that's a separate conversation from validating what we have here.

PR #640, opened 2025-03-12
Feat/multi license core 2

### LINKED ISSUE #569: Multiple Licenses - update core data mode

jusdino:
### Description List
- Updated core data model to support a provider having licenses of multiple types in one state
- Added data migration for existing provider data
- Reworked our python test running to fix pytest coverage under-reporting

Closes #569


backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/privilege/record.py
```diff
@@ -25,8 +25,16 @@ class AttestationVersionRecordSchema(Schema):
     version = String(required=True, allow_none=False)
 
 
+class ValidatesLicenseTypeMixin:
+    @validates_schema
+    def validate_license_type(self, data, **kwargs):  # noqa: ARG001 unused-argument
+        license_types = config.license_types_for_compact(data['compact'])
+        if data['licenseType'] not in license_types:
+            raise ValidationError({'licenseType': [f'Must be one of: {", ".join(license_types)}.']})
```
landonshumway-ia, line 33: This logic is the same as the check for license schemas right? Can we not use this ValidatesLicenseTypeMixin class in both privileges and licenses for a single source of logic for this?

landonshumway-ia (commented):
Data migration looks good. Just one question about the license type validation mixin

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/privilege/record.py
jusdino: Sure, that makes sense.

landonshumway-ia (approved):

landonshumway-ia:
@jlkravitz This is now ready for your review. Thanks

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
```diff
@@ -401,13 +384,19 @@ def create_provider_privileges(
...
                             'dateOfRenewal': privilege_record['dateOfRenewal'],
                             'dateOfExpiration': privilege_record['dateOfExpiration'],
                             'compactTransactionId': compact_transaction_id,
                             'privilegeId': privilege_record['privilegeId'],
                             'attestations': attestations,
+                            **(
+                                {'persistedStatus': ProviderEligibilityStatus.ACTIVE}
+                                if original_privilege['persistedStatus'] == ProviderEligibilityStatus.INACTIVE
+                                else {}
+                            ),
```
jlkravitz, line 399: I _think_ this assumes that upstream validation code guarantees that the new license is not expired. True? If so, can we just add a brief comment here to explicitly state that assumption.

backend/compact-connect/lambdas/python/common/cc_common/data_model/provider_record_util.py
```diff
@@ -0,0 +1,212 @@
...
+        :param provider_records: The list of provider records to search through
+        :return: The provider's home state selection
+        """
+        home_state_selection_records = ProviderRecordUtility.get_records_of_type(
+            provider_records,
+            ProviderRecordType.HOME_JURISDICTION_SELECTION,
+        )
+        if not home_state_selection_records:
+            return None
+
```
jlkravitz, line 104: If we shouldn't have more than one home jurisdiction selection, worth asserting that here (or otherwise logging something)?

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/license/__init__.py
```diff
@@ -43,11 +40,5 @@ class LicenseCommonSchema(ForgivingSchema):
     homeAddressState = String(required=True, allow_none=False, validate=Length(2, 100))
     homeAddressPostalCode = String(required=True, allow_none=False, validate=Length(5, 7))
     militaryWaiver = Boolean(required=False, allow_none=False)
-    emailAddress = Email(required=False, allow_none=False, validate=Length(1, 100))
+    emailAddress = Email(required=False, allow_none=False)
```
jlkravitz, line 43: What's the motivation behind removing the maximum length constraint on emails? It seems valuable to have at least _some_ upper-bound, even if it's high?

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/license/api.py
```diff
@@ -53,7 +53,7 @@ class LicensePostRequestSchema(ForgivingSchema):
     homeAddressState = String(required=True, allow_none=False, validate=Length(2, 100))
     homeAddressPostalCode = String(required=True, allow_none=False, validate=Length(5, 7))
     militaryWaiver = Boolean(required=False, allow_none=False)
-    emailAddress = Email(required=False, allow_none=False, validate=Length(1, 100))
+    emailAddress = Email(required=False, allow_none=False)
```
jlkravitz, line 56: Same question here and in the other instances of this change in this file.

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/license/api.py
```diff
@@ -106,6 +105,7 @@ class LicenseUpdateGeneralResponseSchema(ForgivingSchema):
     providerId = Raw(required=True, allow_none=False)
     compact = Compact(required=True, allow_none=False)
     jurisdiction = Jurisdiction(required=True, allow_none=False)
+    licenseType = String(required=True, allow_none=False)
```
jlkravitz, line 108: Isn't this technically an enum, such that we could use the OneOf construct?

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/license/record.py
```diff
@@ -133,16 +134,17 @@ def generate_pk_sk(self, in_data, **kwargs):  # noqa: ARG001 unused-argument
...
         )
         return in_data
 
     @validates_schema
     def validate_license_type(self, data, **kwargs):  # noqa: ARG001 unused-argument
         license_types = config.license_types_for_compact(data['compact'])
-        if data['previous']['licenseType'] not in license_types:
-            raise ValidationError({'previous.licenseType': [f'Must be one of: {", ".join(license_types)}.']})
+        if data['licenseType'] not in license_types:
+            raise ValidationError({'licenseType': [f'Must be one of: {", ".join(license_types)}.']})
```
jlkravitz, line 147: Can we use the `ValidatesLicenseTypeMixin` instead?

backend/compact-connect/lambdas/python/migration/dummy_migration/main.py
```diff
@@ -0,0 +1,7 @@
+"""
+Dummy file to test the data migration construct.
```
jlkravitz, line 2: Did we intend to keep this dummy migration around?

backend/compact-connect/lambdas/python/provider-data-v1/handlers/__init__.py
```diff
@@ -15,55 +16,10 @@ def get_provider_information(compact: str, provider_id: str) -> dict:
...
-        logger.error("Failed to find a provider's primary record!", provider_id=provider_id)
-        raise CCInternalException('Unexpected provider data')
-
-    provider['licenses'] = list(licenses.values())
-    provider['privileges'] = list(privileges.values())
-    provider['militaryAffiliations'] = military_affiliations
-    if home_jurisdiction_selection:
-        provider['homeJurisdictionSelection'] = home_jurisdiction_selection
-    return provider
+    return ProviderRecordUtility.assemble_provider_records_into_object(provider_data['items'])
```
jlkravitz, line 25: nice cleanup!

backend/compact-connect/lambdas/python/provider-data-v1/handlers/privileges.py
```diff
@@ -20,28 +20,32 @@ def deactivate_privilege(event: dict, context: LambdaContext):  # noqa: ARG001 u
...
     with logger.append_context_keys(
-        compact=compact, provider_id=provider_id, jurisdiction=jurisdiction, license_type=license_type
+        compact=compact, provider_id=provider_id, jurisdiction=jurisdiction, license_type=license_type_abbr
     ):
         # Get the user's scopes to check for jurisdiction-specific admin permission
         scopes = get_event_scopes(event)
         jurisdiction_admin_scope = f'{jurisdiction}/{compact}.{CCPermissionsAction.ADMIN}'
         compact_admin_scope = f'{compact}/{CCPermissionsAction.ADMIN}'
 
+        # Validate the the license type is a supported abbreviation
```
jlkravitz, line 33: Should this comment go on L38 instead?

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_ingest.py
```diff
@@ -787,3 +789,255 @@ def test_inactive_privileges_included_in_privilege_jurisdictions(self):
...
+            'compact': 'aslp',
+            'providerId': provider_id,
+            'jurisdiction': 'oh',
+            'dateOfSelection': self.config.current_standard_datetime,
+        }
+
+        # Create the home jurisdiction selection record
+        schema = ProviderHomeJurisdictionSelectionRecordSchema()
+        serialized_record = schema.dump(home_jurisdiction_selection)
+        self.config.provider_table.put_item(Item=serialized_record)
```
jlkravitz, line 909: question- is there any reason this couldn't happen _before_ uploading the first OH license above?

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_ingest.py
```diff
@@ -787,3 +789,255 @@ def test_inactive_privileges_included_in_privilege_jurisdictions(self):
...
+        self.assertEqual('Björk', oh_license['givenName'])
+
+        self.assertEqual('audiologist', ky_license['licenseType'])
+        self.assertEqual('B0608337260', ky_license['licenseNumber'])
+        self.assertEqual('2020-06-06', ky_license['dateOfIssuance'])
+        self.assertEqual('Audrey', ky_license['givenName'])
+
+        # Verify that the provider data was copied from the audiologist license in 'ky'
+        # because it has a newer issuance date. We can verify this by checking the givenName.
+        self.assertEqual('ky', provider_data['licenseJurisdiction'])
```
jlkravitz, line 1041: Just confirming: This test is basically the same as the "same jurisdiction" test, except that we would expect the `licenseJurisdiction` to be KY here instead of OH. Is that right? Anything else that is different?

jlkravitz (changes requested):
good stuff! love the new ProviderRecordUtility- cleans things up a lot.

a few questions/items below.

backend/compact-connect/bin/update_postman_collection.py
```diff
@@ -96,6 +171,12 @@ def merge_collections(new_collection: dict[str, Any], existing_collection: dict[
         if new_provider_users and existing_provider_users and 'auth' in existing_provider_users:
             new_provider_users['auth'] = existing_provider_users['auth']
 
+    # Preserve the bulk-upload script
+    preserve_bulk_upload_script(new_collection, existing_collection)
+
+    # Copy the Upload Document request
+    copy_upload_document_request(new_collection, existing_collection)
+
```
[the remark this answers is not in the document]
jusdino: There are some endpoints in our Postman collection that have scripts attached, which facilitate things like authentication flows, were we set variables (like the access token) for re-use across other requests. In this case, we have scripts that take the fields returned from the GET bulk upload call, for resuse in the Upload Docuemnt call. If we don't explicitly copy them over, they'll be lost when we regenerate from our API spec.

backend/compact-connect/lambdas/python/migration/dummy_migration/main.py
jusdino: Yes. It allows us to test the CDK construct, even when we have no actual migrations in flight.

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/license/record.py
jusdino: We could, except that this particular record has a `licenseType` field in two locations, so we need to tweak the logic a bit.

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_ingest.py
jusdino: If we wanted to, we'd have to force the provider id on creation with a mock or something. Why would we want to, though?

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_ingest.py
jusdino: Yeah, that's it. These three, *same jurisdiction*, *home jurisdiction*, and *different jurisdiction* are just confirming that the order of precedence is behaving correctly in different cases. In particular, you'll see that *home jurisdiction* case is effectively overriding the outcome of what would otherwise be the same as *different jurisdiction*. We're being thorough here because the 'best license' logic is getting complex.

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/license/api.py
jusdino: It could be, though the true allowed types depend on the `compact` value, which is why we use the schema-level validation hooks to check that, generally. In this case, this is a ResponseSchema, which we don't really use for validation, so we don't worry about it.

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/license/__init__.py
jusdino: I was reviewing the API schema and the marshmallow schema to identify discrepancies and remove redundancies. In the case of `email` fields, I opted to remove length constraints in favor of the more specific `Email` field formatting, letting the api schema keep the lengths reasonable.

After looking at this closer now, I realized that the license data doesn't necessarily get captured by an API Gateway body validator, so I've added it back to the `LicensePostRequestSchema`, where they are first validated on their way into the system. While doing that, I also noticed that we had an email address validation gap on the provider registration endpoint, so I've tested and fixed that as well.

backend/compact-connect/bin/update_postman_collection.py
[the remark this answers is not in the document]
jlkravitz: Ah, didn't realize you could have custom scripts in postman. Cool!

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
jlkravitz: To clarify, my question was related to the fact that this code explicitly sets `persistedStatus` to `ACTIVE`, if it was originally `INACTIVE` – is that not making an assumption of the validity of the license?

Or perhaps I'm confusing myself– is `persistedStatus` the field that denotes whether a license or privilege has been deactivated by a staff user, such that this code is just ensuring that the new privilege starts in an active state?

jusdino: You have that right - `persistedStatus` the field on a privilege that is used when an admin explicitly deactivates a privilege. But any new/renewed privilege is explicitly `ACTIVE`, and this action makes no assumptions about the status of the license.

jlkravitz (approved):
@isabeleliassen good to go

isabeleliassen (approved):

PR #644, opened 2025-03-13
Feat/Privilege deactivation notification

### LINKED ISSUE #520: backend for privilege deactivated notification
- notification to states of deactivation: "This message is to notify you that privilege [privilege number] held by [licensee name] was deactivated and can no longer be used to practice." 
- notification to practitioners of deactivation: "This message is to notify you that your privilege [privilege number] is deactivated and can no longer be used to practice."

landonshumway-ia:
Whenever a privilege is deactivated by an admin, we need to send notifications to both the licensee that owned that privilege, as well as the staff of the jurisdiction for which that privilege was purchased.

### Requirements List
- This change is backwards compatible with the current system, no new requirements.

### Description List
- Added wrapper EmailServiceClient python class to abstract underlying contract with email service lambda.
- Add privilege deactivation email templates to Node email notification service lambda
- Add calls to email service in privilege deactivation endpoint

### Testing List
- Added functional/unit testing for node/python lambda implementation
- verified functionality through local sandbox testing

Closes #520


backend/compact-connect/lambdas/python/common/cc_common/email_service_client.py
```diff
@@ -0,0 +1,185 @@
...
+from typing import Any
+
+import boto3
+from aws_lambda_powertools.logging import Logger
+
+from cc_common.exceptions import CCInternalException
+
+
+class EmailServiceClient:
+    """
```
jusdino, line 12: Nice addition!

backend/compact-connect/lambdas/python/common/tests/function/test_data_client.py
```diff
@@ -830,12 +831,14 @@ def test_deactivate_privilege_on_inactive_privilege_has_no_effect(self):
         test_data_client = DataClient(self.config)
 
         # Now, deactivate the privilege
-        test_data_client.deactivate_privilege(
-            compact='aslp',
-            provider_id=provider_id,
-            jurisdiction='ne',
-            license_type_abbr='aud',
-        )
+        with self.assertRaises(CCInvalidRequestException) as context:
```
jusdino, line 834: I guess this will make more sense as an interface when we start accepting descriptions on a deactivation - admins will expect those to show up, so idempotency is less useful. 👍

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_privileges.py
```diff
@@ -129,6 +132,105 @@ def test_board_admin_can_deactivate_privilege(self):
...
+        self.assertEqual(400, resp['statusCode'])
+        self.assertEqual({'message': 'Privilege already deactivated'}, json.loads(resp['body']))
+
+    @patch('cc_common.config._Config.current_standard_datetime', datetime.fromisoformat('2024-11-08T23:59:59+00:00'))
+    @patch('cc_common.config._Config.email_service_client')
+    def test_deactivate_privilege_handler_still_sends_state_notification_if_provider_notification_failed_to_send(
+        self, mock_email_service_client
+    ):
+        """
+        If a board admin has admin permission in the privilege jurisdiction, they can deactivate a privilege
```
jusdino, line 185: Update docstring?

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_privileges.py
```diff
@@ -129,6 +132,105 @@ def test_board_admin_can_deactivate_privilege(self):
...
+            provider_last_name='Guðmundsdóttir',
+        )
+
+    @patch('cc_common.config._Config.current_standard_datetime', datetime.fromisoformat('2024-11-08T23:59:59+00:00'))
+    @patch('cc_common.config._Config.email_service_client')
+    def test_deactivate_privilege_handler_raises_exception_if_state_notification_failed_to_send(
+        self, mock_email_service_client
+    ):
+        """
+        If a board admin has admin permission in the privilege jurisdiction, they can deactivate a privilege
```
jusdino, line 220: Update docstring?

jusdino (approved):
Looks great! No significant notes but I did spot a couple of test docstrings that need updating.

landonshumway-ia:
@jlkravitz This is now ready for your review. Thank you

backend/compact-connect/lambdas/nodejs/lib/email/email-notification-service.ts
```diff
@@ -32,12 +32,33 @@ export class EmailNotificationService extends BaseEmailService {
         }
     }
 
+    private async getJurisdictionRecipients(compact: string,
```
jlkravitz, line 35: nit: Can the first parameter be on a new line here and anywhere else in this file? I missed it at first when reviewing code that called this function.

backend/compact-connect/lambdas/python/provider-data-v1/handlers/privileges.py
```diff
@@ -64,4 +64,50 @@ def deactivate_privilege(event: dict, context: LambdaContext):  # noqa: ARG001 u
...
+                provider_first_name=provider_first_name,
+                provider_last_name=provider_last_name,
+            )
+        except CCInternalException as e:
+            # Log the error but don't fail the deactivation process
+            logger.error('Failed to send jurisdiction privilege deactivation notifications', exception=str(e))
+            failed_to_send_notification = True
+
+    if failed_to_send_notification:
+        raise CCInternalException('Privilege was deactivated, but email notification failed to send.')
```
jlkravitz, line 111: Do we really want to throw an error if this happens? Will the frontend see this message? If not, what indication will they have that the failed sending of the email is the only issue?

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_privileges.py
```diff
@@ -129,6 +132,106 @@ def test_board_admin_can_deactivate_privilege(self):
...
+        with open('../common/tests/resources/dynamo/privilege.json') as f:
+            privilege = json.load(f)
+            # set persisted status to deactivated
+            privilege['persistedStatus'] = 'inactive'
+            self.config.provider_table.put_item(Item=privilege)
+
+        # The user has admin permission for oh
+        resp = self._request_deactivation_with_scopes('openid email ne/aslp.admin')
+        self.assertEqual(400, resp['statusCode'])
+        self.assertEqual({'message': 'Privilege already deactivated'}, json.loads(resp['body']))
```
jlkravitz, line 177: Is this testing admin permissions, or already-deactivated privileges? How are the permissions relevant?

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_privileges.py
```diff
@@ -129,6 +132,106 @@ def test_board_admin_can_deactivate_privilege(self):
...
+        )
+
+        # We expect the handler to raise an exception with a message
+        with self.assertRaises(CCInternalException) as context:
+            self._request_deactivation_with_scopes('openid email ne/aslp.admin')
+
+        self.assertEqual('Privilege was deactivated, but email notification failed to send.', str(context.exception))
+
+        # Even though the first notification failed, the handler should still have attempted to send both
+        # notifications
```
jlkravitz, line 201: The test name implies that you'd confirm the state notification is still sent, but this code just seems to confirm that _both_ email functions were called.

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_privileges.py
```diff
@@ -129,6 +132,106 @@ def test_board_admin_can_deactivate_privilege(self):
...
+            self.config.provider_table.put_item(Item=privilege)
+
+        # The user has admin permission for oh
+        resp = self._request_deactivation_with_scopes('openid email ne/aslp.admin')
+        self.assertEqual(400, resp['statusCode'])
+        self.assertEqual({'message': 'Privilege already deactivated'}, json.loads(resp['body']))
+
+    @patch('cc_common.config._Config.current_standard_datetime', datetime.fromisoformat('2024-11-08T23:59:59+00:00'))
+    @patch('cc_common.config._Config.email_service_client')
+    def test_deactivate_privilege_handler_still_sends_state_notification_if_provider_notification_failed_to_send(
```
jlkravitz, line 181: Should we stick to jurisdiction for consistency?

```suggestion
    def test_deactivate_privilege_handler_still_sends_jurisdiction_notification_if_provider_notification_failed_to_send(
```

backend/compact-connect/lambdas/python/purchases/tests/function/test_handlers/test_transaction_reporting.py
```diff
@@ -864,30 +789,33 @@ def test_generate_report_handles_unknown_jurisdiction(self, mock_lambda_client):
...
         ]
-        self.assertEqual(2, len(jurisdiction_calls))  # Only OH and KY should get reports
 
-        jurisdiction_report_payloads = [json.loads(call[1]['Payload']) for call in jurisdiction_calls]
-        reported_jurisdictions = {payload['jurisdiction'] for payload in jurisdiction_report_payloads}
-        self.assertEqual({'oh', 'ky'}, reported_jurisdictions)  # Verify only OH and KY got reports
+        # Verify only OH and KY got reports
+        self.assertIn('oh', jurisdiction_args)
+        self.assertIn('ky', jurisdiction_args)
+        self.assertNotIn('xx', jurisdiction_args)
```
jlkravitz, line 807: Why the change here? Seems the original would more completely capture what we are trying to capture? (i.e., that `reported_jurisdictions` include exactly OH and KY).

jlkravitz (commented):
few questions/comments!

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_privileges.py
landonshumway-ia: The user still has to have proper permission to call the endpoint, even if the privilege is already deactivated. Is it the comment that is confusing? If so I can just remove that comment.

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_privileges.py
landonshumway-ia: That _is_ how we confirm that the notification is sent for the purpose of this test. If the email client function is called to send the state notification even though the first call failed, then the exception handling is working as expected. Obviously, these are mocks, so there are no actual api calls being made. The intent with this test is to verify that the failure of one function call does not prevent the call of the other. If the name needs rewording to make that more clear. Let me know.

backend/compact-connect/lambdas/python/purchases/tests/function/test_handlers/test_transaction_reporting.py
landonshumway-ia: ah good catch, that was my co-pilot getting 'smart' on me. I'll revert it.

backend/compact-connect/lambdas/python/provider-data-v1/handlers/privileges.py
landonshumway-ia: This is a good point. The main purpose of raising the exception here was to ensure we have an alert that fires as this requires investigation into the cause of the failure. Given that this will result in a 502, unfortunately the user will not see this particular message, but if they try again, they will get back a 400 with the message 'Privilege already deactivated', and if they refresh the page they will see that the privilege is in an inactive status.

The alternative is to return a 200, which doesn't seem right either, since the full handler was not successful. If we did return a 200 though, we could send a metric with an associated cloudwatch alarm that will fire if this occurs so support can still investigate. Curious what your thoughts are for how we should approach error handling here.

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_privileges.py
jlkravitz: I guess I'm confused because the scopes are for nebraska?

landonshumway-ia: Yes, the test user here is a nebraska jurisdiction admin for the aslp compact, which has permission to deactivate aslp privileges within the nebraska jurisdiction.

jlkravitz: Right. The comment says OH, though? I think it's fine to just remove the comment, ha :) Or, update. Whatever your preference.

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_privileges.py
jlkravitz: Ah, I think I understand now. I think this is good as-is. Thanks for explaining.

backend/compact-connect/lambdas/python/provider-data-v1/handlers/privileges.py
jlkravitz: Got it. Ultimately, I defer to you and think either choice is reasonable. My only thought is regarding the 'importance' of the email to the business logic.

If the emails _were_ vital, I would expect the handler to raise an exception if it failed, _and_ perhaps to rollback all other logic (i.e., reactivate the privilege).

If it were less vital, I might expect the handler to follow your 2nd suggested pattern – return 200, log error separately as a custom metric.

It seems overly complicated to implement rollback logic, in which case I would opt for the 200 approach. But that's assuming the above is good logic, which may or may not be the case.

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_privileges.py
landonshumway-ia: I've updated it

backend/compact-connect/lambdas/python/provider-data-v1/handlers/privileges.py
landonshumway-ia: We can move to the 200 approach, since in either case if the email notification step fails the end user can't do anything about it. Support staff will need to figure out what went wrong and likely resend the email notification once the issue has been addressed. We expect this occurrence to be extremely rare, so I will set up a custom metric with a cloudwatch alert, and just return a 200

landonshumway-ia: Update has been pushed

landonshumway-ia: Verified in my sandbox env that the alert fires when this exception case is triggered

jlkravitz (approved):

isabeleliassen (approved):

PR #651, opened 2025-03-14
Remove all references to 'militaryWaiver' field in project

### LINKED ISSUE #618: remove militaryWaiver from license uploads
Since CompactConnect didn't end up using the `militaryWaiver` field that is provided via state license data, we should remove it to reduce confusion. We will need to update record schema to remove the field and run a migration to remove the field from existing records.

landonshumway-ia:
We allow licensees to upload their own military affiliation records into the CompactConnect system, so we no longer require the 'militaryWaiver' field to be specified by states when uploading license data. This change removes all reference to that field.

### Requirements List
- The 'militaryWaiver' field was being validated in the schemas, so this is technically a backwards incompatible change. A migration script has been added which will remove the field from all records. Between the script running and the full deployment of the api with the new schema, there will be a brief period where api calls to get provider data will fail (~5 - 10 mins).

### Description List
- Removed references to 'militaryWaiver' field from license and provider schemas
- Added migration script to remove field from existing records

### Testing List
- Verified migration script in local sandbox environment
- Updated all tests to stop referencing removed field
- Code review

Closes #618


jusdino (changes requested):
Looks great! One minor nit.

jusdino (approved):
Great!

landonshumway-ia:
@jlkravitz This is ready for your review. Thanks!

jlkravitz (approved):
Looks good!

@isabeleliassen Good to merge this one.

isabeleliassen (approved):

PR #652, opened 2025-03-18
Add field for enabling registration at compact/jurisdiction levels

### LINKED ISSUE #597: Registration Live Toggle
- toggle registration options at the compact level and at the state level
- can add specific error message so people understand that it's because a state/compact isn't accepting registrations
- no frontend anticipated

landonshumway-ia:
Some compacts/jurisdictions will need time to onboard onto the system, during which we do not want licensees being able
to register within that particular compact or jurisdiction. This adds a new field `licenseeRegistrationEnabledForEnvironments` which is a list of environments that allow registration within the respective compact/jurisdiction.

### Requirements List
- No new requirements

### Description List
- Added `licenseeRegistrationEnabledForEnvironments` field to all configuration files to gate registration
- Added checks in registration point to ensure registration is enabled for selected compact/jurisdiction pair

### Testing List
- Functional tests added to 
- Code review

Closes #597


backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_registration.py
```diff
@@ -7,13 +7,15 @@
...
-TEST_COMPACT = 'aslp'
+TEST_COMPACT_ABBR = 'aslp'
+TEST_COMPACT_NAME = 'Audiology and Speech Language Pathology'
 TEST_LICENSE_TYPE = 'speech-language pathologist'
 
 MOCK_SSN_LAST_FOUR = '1234'
 MOCK_GIVEN_NAME = 'Joe'
 MOCK_FAMILY_NAME = 'Dokes'
-MOCK_STATE = 'ky'
+MOCK_JURISDICTION_POSTAL_CODE = 'ky'
```
jusdino, line 17: Strictly speaking, a zip code is a kind of postal code, this is a postal abbreviation.

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/compact/__init__.py
```diff
@@ -107,3 +107,7 @@ def compact_adverse_actions_notification_emails(self) -> list[str] | None:
     @property
     def compact_summary_report_notification_emails(self) -> list[str] | None:
         return self.get('compactSummaryReportNotificationEmails')
+
+    @property
+    def licensee_registration_enabled_for_environments(self) -> list[str] | None:
```
jusdino, line 112: Unless I'm misunderstanding, this compact/jurisdiction registration behavior is different from the `activeEnvironments` behavior, which can suffer the same downside that we worked around for `activeEnvironments` - we don't want to change checked-in files to activate environments in our sandboxes. The difference in behavior can also just be a bit confusing.

Should we just automatically enable registration for sandboxes, like we do for compact/jurisdiction config in general?

jusdino (changes requested):
Looks great, but with one overall concern:

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/compact/__init__.py
landonshumway-ia: Yes, good call. I've updated the uploader process to automatically include sandbox environments in the list of registration environments.

jusdino (approved):
Great! @jlkravitz, this one is ready for you!

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/jurisdiction/record.py
```diff
@@ -45,6 +45,12 @@ class JurisdictionRecordSchema(BaseRecordSchema):
         required=True,
         allow_none=False,
     )
+    licenseeRegistrationEnabledForEnvironments = List(
+        String(required=True, allow_none=False, validate=OneOf(['test', 'prod', config.environment_name])),
+        required=True,
```
jlkravitz, line 50: should this be `required=False` like it is with the compact config record?

jlkravitz (commented):
one question!

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/jurisdiction/record.py
landonshumway-ia: This schema is a forgiving schema, as defined on line 24 right above. The compact record schema should be making this field required as well, so good catch there. I'll fix it.

jlkravitz: Got it. Was confusing required fields vs. the definition of a `ForgivingSchema`. Glad I caught that regardless! :)

jlkravitz (approved):
@isabeleliassen Good to go!

isabeleliassen (approved):

PR #670, opened 2025-03-24
Bug/work around authorizenet bug

jusdino:
Authorize.Net's python SDK, `authorizenet` currently has [a bug that makes it uninstallable in environments that use the latest of python-maintained library management tools](https://github.com/AuthorizeNet/sdk-python/issues/166). Specifically, `setuptools`, with version `78+` stopped supporting some long-deprecated configuration arguments, which the `authorizenet` library still uses. In the hopes that Authorize.Net will quickly remedy this bug, we have adopted a temporary work-around to unblock our deploy pipelines: We've custom-built a docker image for building our lambdas that includes a downgraded version of `setuptools`.

This work-around is far from an ideal solution, so we should discuss it carefully before approving. Unfortunately, it is also the only short-term solution I can see to getting our pipeline unblocked.

### Description List
- Add a `Dockerfile` to build a custom SAM python lambda builder image with downgraded setuptools
- Update `PythonLambda` to use this custom image, as hosted in a public registry we've created for the purpose

### Testing List
- Code review

### DIFF
--- a/backend/compact-connect/common_constructs/python_function.py
+++ b/backend/compact-connect/common_constructs/python_function.py
@@ -2,13 +2,13 @@ from __future__ import annotations
 
 import os
 
-from aws_cdk import Duration, Stack
+from aws_cdk import DockerImage, Duration, Stack
 from aws_cdk.aws_cloudwatch import Alarm, ComparisonOperator, Stats, TreatMissingData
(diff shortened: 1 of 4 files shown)


jusdino:
Work-around is OBE: `setuptools` released another update where they postponed pulling support for the deprecated `config.cfg` fields. So long as authorize.net updates before 2026, we should be OK.

PR #681, opened 2025-03-27
Feat/split license statuses

### LINKED ISSUE #667: Three license status fields
User feedback has indicated that we need to disambiguate our license statuses between three different concepts, which mostly represent to the answer to three questions:
- Is the license-holder eligible for participation in the compact (eligible/ineligible)?
- Is the license-holder allowed to practice their profession (active/inactive)?
- What is a descriptive name for the license's state (active, expired, revoked, suspended, etc)?

To accommodate this, we will split our current `status` (`jurisdictionStatus`, internally) field into _three_ status fields for states to provide on upload:
- `compactEligibility`
- `licenseStatus`
- `licenseStatusName`

We will need to revise all of our existing logic (and API) to key off from the appropriate new fields and make them available to the UI.

@becca-collins , @ChiefStief, @isabeleliassen - this may require some design/UI updates, since this will be a breaking API change and the license/privilege history may become more complex, if we want to represent both the `compactEligibility` and `licenseStatus` fields.

jusdino:
We split the concept of 'license status' into three more dedicated fields, to better reflect the reality of how the compacts operate.

🚨 This is a BREAKING API CHANGE that needs design and UI updates 🚨 

### Description List
- Moved all runtime references to `active` or `inactive` to an enum
- Split license/provider `status` into `licenseStatus`, `compactEligibility` and `licenseStatusName`
- Updated status-based logic to reference new fields
- Updated scripts to match new data model
- Added migration to move existing data to new model

### Testing List
- Code review
- Deploy this to a sandbox with data to independently verify the migration

Closes #667 

## Summary by CodeRabbit

- **New Features**
  - Introduced new migration and purging utilities to streamline provider record management.
  - Enhanced license data with explicit fields for status naming and compact eligibility, improving clarity on license validity.

- **Refactor**
  - Standardized status fields by transitioning from generic “status” and “persistedStatus” to detailed “licenseStatus,” “licenseStatusName,” and “administratorSetStatus.”
  - Updated environment configurations to support testing workflows.

- **Documentation**
  - Revised API specifications and license CSV documentation to reflect the new field definitions and updated server endpoints.

- **Tests**
  - Updated test suites to validate the refined status and eligibility mappings consistently.
  - Enhanced tests to ensure alignment with new eligibility checks and updated field names.


backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/license/record.py
```diff
@@ -94,7 +102,9 @@ class LicenseUpdateRecordPreviousSchema(StrictSchema):
     homeAddressPostalCode = String(required=True, allow_none=False, validate=Length(5, 7))
     emailAddress = Email(required=False, allow_none=False)
     phoneNumber = ITUTE164PhoneNumber(required=False, allow_none=False)
-    jurisdictionStatus = ActiveInactive(required=True, allow_none=False)
+    licenseStatusName = String(required=False, allow_none=False, validate=Length(1, 100))
+    persistedLicenseStatus = ActiveInactive(required=True, allow_none=False)
+    persistedCompactEligibility = CompactEligibility(required=True, allow_none=False)
```
landonshumway-ia, line 107: It would be beneficial to have these field names denote where they originated from. In this case, these fields are uploaded by jurisdiction IT departments. We can capture this through the following field names:
```
jurisdictionUploadedLicenseStatus
jurisdictionUploadedCompactEligibility
```

Similarly, the `persistedStatus` in the privilege record could follow this pattern to denote that the field is set by an administrator to either active or inactive:
```
administratorSetStatus
```
This will make it much clearer where these fields are coming from and how they are used to calculate the final status used by the system.

landonshumway-ia (changes requested):
Based on our offline discussion, adding this one refactoring item. Thanks

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/provider/api.py
```diff
@@ -55,7 +58,9 @@ class ProviderGeneralResponseSchema(ForgivingSchema):
     phoneNumber = ITUTE164PhoneNumber(required=False, allow_none=False)
     compactConnectRegisteredEmailAddress = Email(required=False, allow_none=False)
     cognitoSub = String(required=False, allow_none=False)
-    status = ActiveInactive(required=True, allow_none=False)
```
landonshumway-ia, line 58: Isn't this is going to break the frontend, which is currently depending on this 'status' field in the provider list view? I was under the impression that we were going to leave this internal field alone. If that is not the case, then the frontend code should be updated as part of this PR to start checking for the 'licenseStatus' value instead, if that is the intent.

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/provider/api.py
```diff
@@ -92,12 +97,13 @@ class ProviderPublicResponseSchema(ForgivingSchema):
...
     compact = Compact(required=True, allow_none=False)
     licenseJurisdiction = Jurisdiction(required=True, allow_none=False)
+    licenseStatus = ActiveInactive(required=True, allow_none=False)
+    compactEligibility = CompactEligibility(required=True, allow_none=False)
     npi = NationalProviderIdentifier(required=False, allow_none=False)
     givenName = String(required=True, allow_none=False, validate=Length(1, 100))
     middleName = String(required=False, allow_none=False, validate=Length(1, 100))
     familyName = String(required=True, allow_none=False, validate=Length(1, 100))
     suffix = String(required=False, allow_none=False, validate=Length(1, 100))
-    status = ActiveInactive(required=True, allow_none=False)
```
landonshumway-ia, line 100: same here, my understanding is the public provider list view references this field for one of the column displays. If that's true, we will need to update the frontend as part of this PR to match.

backend/compact-connect/lambdas/python/common/tests/resources/api/license-post.json
```diff
@@ -6,7 +6,9 @@
   "middleName": "Gunnar",
   "familyName": "Guðmundsdóttir",
   "licenseType": "speech-language pathologist",
-  "status": "active",
+  "licenseStatus": "active",
+  "licenseStatusName": "DEFINITELY_A_HUMAN",
```
landonshumway-ia, line 10: interesting status name 😆

backend/compact-connect/lambdas/python/common/tests/resources/licenses.csv
```diff
@@ -1,6 +1,6 @@
...
-2024-06-30,0608337260,A0608337260,2024-06-30,speech-language pathologist,Guðmundsdóttir,Birmingham,Gunnar,active,529-31-5408,123 A St.,Apt 321,2024-06-30,oh,35004,Björk,2024-06-30
-2024-06-30,0608337260,B0608337260,2024-06-30,audiologist,Scott,Huntsville,Patricia,active,529-31-5409,321 B St.,,2024-06-30,oh,35005,Elizabeth,2024-06-30
-2024-06-30,0608337260,C0608337260,2024-06-30,speech-language pathologist,毛,Hoover,泽,active,529-31-5410,10101 Binary Ave.,,2024-06-30,oh,35006,覃,2024-06-30
-2024-06-30,0608337260,D0608337260,2024-06-30,speech-language pathologist,Adams,Tuscaloosa,Michael,inactive,529-31-5411,1AB3 Hex Blvd.,,2024-06-30,oh,35007,John,2024-06-30
-2024-06-30,0608337260,E0608337260,2024-06-30,speech-language pathologist,Carreño Quiñones,Montgomery,José,active,529-31-5412,10 Main St.,,2024-06-30,oh,35008,María,2024-06-30
+dateOfIssuance,npi,licenseNumber,dateOfBirth,licenseType,familyName,homeAddressCity,middleName,licenseStatus,compactEligibility,ssn,homeAddressStreet1,homeAddressStreet2,dateOfExpiration,homeAddressState,homeAddressPostalCode,givenName,dateOfRenewal
```
landonshumway-ia, line 1: Is there any value to including the `licenseStatusName` as part of this test file?

backend/compact-connect/lambdas/python/migration/667_three_license_status_fields/main.py
```diff
@@ -0,0 +1,152 @@
...
+                        config.provider_table.update_item(
+                            Key=key,
+                            UpdateExpression='SET #ju_compact_eligibility = :compact_eligibility, '
+                            '#ju_license_status = :license_status, '
+                            '#license_status_name = :license_status_name '
+                            'REMOVE #jurisdiction_status',
+                            ExpressionAttributeValues={
+                                ':license_status': license_status,
+                                ':compact_eligibility': 'eligible' if license_status == 'active' else 'ineligible',
+                                ':license_status_name': 'SOMETHING_USEFUL',
```
landonshumway-ia, line 76: `licenseStatusName` is optional right? so we could just disregard it altogether since we're just putting a mock value in it. It's fine to leave it in though I guess since we're only in the test environment for now. If we had data in prod I would recommend we leave the mock value out.

backend/compact-connect/lambdas/python/migration/667_three_license_status_fields/main.py
```diff
@@ -0,0 +1,152 @@
...
+                            UpdateExpression='SET #previous.#ju_compact_eligibility = :compact_eligibility, '
+                            '#previous.#ju_license_status = :license_status, '
+                            '#previous.#license_status_name = :license_status_name '
+                            'REMOVE #previous.#jurisdiction_status',
+                            ExpressionAttributeValues={
+                                ':license_status': previous_license_status,
+                                ':compact_eligibility': 'eligible'
+                                if previous_license_status == 'active'
+                                else 'ineligible',
+                                ':license_status_name': 'SOMETHING_USEFUL',
```
landonshumway-ia, line 103: Ditto here. This is effectively going to add a junk value to every update record, even though we don't really need it.

backend/compact-connect/lambdas/python/migration/667_three_license_status_fields/main.py
```diff
@@ -0,0 +1,152 @@
...
+                                '#updatedValues.#ju_compact_eligibility = :compact_eligibility, '
+                                '#updatedValues.#ju_license_status = :license_status, '
+                                '#updatedValues.#license_status_name = :license_status_name '
+                                'REMOVE #updatedValues.#jurisdiction_status',
+                                ExpressionAttributeValues={
+                                    ':license_status': updated_license_status,
+                                    ':compact_eligibility': 'eligible'
+                                    if updated_license_status == 'active'
+                                    else 'ineligible',
+                                    ':license_status_name': 'SOMETHING_ELSE_USEFUL',
```
landonshumway-ia, line 133: ditto here

backend/compact-connect/lambdas/python/provider-data-v1/handlers/ingest.py
```diff
@@ -272,7 +273,7 @@ def _populate_update_record(*, existing_license: dict, updated_values: dict, rem
         ):
             update_type = UpdateCategory.RENEWAL
             logger.info('License renewal detected')
-    elif updated_values == {'jurisdictionStatus': ProviderEligibilityStatus.INACTIVE.value}:
+    if updated_values.get('jurisdictionUploadedLicenseStatus') == ActiveInactiveStatus.INACTIVE:
```
landonshumway-ia, line 276: What about the case where the `compactEligibility` goes from eligible to ineligible? Is that going to be handled in a separate ticket?

landonshumway-ia (changes requested):
Looking good, found a couple areas that would be good to get clarification on.

backend/compact-connect/bin/purge_provider_table.py
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/env python3
+"""
+This script is used to purge the provider table of all items.
+It is intended to be run from the command line and will delete all items from the table. Obviously, we never want to run
+this in production.
```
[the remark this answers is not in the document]
jusdino: I couldn't think of an easy way to do that, other than what I have here: In order to purge the production table, the person who runs this script would need to:
1) Reference the production table name in the env var and
2) Configure their CLI to authenticate against the production account (assuming they have access).

Hopefully, between reduced head count of people who can even do that, and the two-factor configuration, it would be hard for someone to accidentally do this.

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/license/record.py
```diff
@@ -144,6 +154,3 @@ def validate_license_type(self, data, **kwargs):  # noqa: ARG001 unused-argument
         license_types = config.license_types_for_compact(data['compact'])
         if data['licenseType'] not in license_types:
             raise ValidationError({'licenseType': [f'Must be one of: {", ".join(license_types)}.']})
-        # We have to check for existence here to allow for the updatedValues partial case
-        if data['updatedValues'].get('licenseType') and data['updatedValues']['licenseType'] not in license_types:
-            raise ValidationError({'updatedValues.licenseType': [f'Must be one of: {", ".join(license_types)}.']})
```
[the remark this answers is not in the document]
jusdino: That field no longer exists.

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/provider/api.py
jusdino: Yes, this is the big breaking change that I was mentioning - we need the design/ui tickets to go with this work because this update will break the UI. I deliberately renamed this API field because `status` is just way too ambiguous, now that we have so much nuance to different types of statuses in the system. The UI update is not so simple as just moving `status` to `licenseStatus`, since this split of statuses requires the whole design change that displays licenses differently for both license status and eligibility, and needs to move privilege logic to eligibility, not license status, etc.

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/provider/api.py
jusdino: Yep. We have a separate ticket for the corresponding UI updates. They're substantial, so I won't be doing them in this ticket.

backend/compact-connect/lambdas/python/migration/667_three_license_status_fields/main.py
jusdino: Yeah, I wanted to add something so that the value can show up in the license cards that the web devs will be working up.

backend/compact-connect/lambdas/python/migration/667_three_license_status_fields/main.py
jusdino: I could set it to a more realistic mock status - I wrote this before we had a handful of examples to work with. I do want the UI to have something to display, though.

backend/compact-connect/lambdas/python/provider-data-v1/handlers/ingest.py
jusdino: I opted to not call that out as a separate update type, since we don't have any known use cases for eligibility status change updates yet. The logic in this whole categorization section is still a bit fuzzy, since we don't have settled definitions of what actually counts as a deactivation, so I'm trying to avoid over complicating it.

backend/compact-connect/bin/purge_provider_table.py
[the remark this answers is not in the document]
landonshumway-ia: Fair enough... I'm just having flashbacks of similar scenarios where accidents happened 👁️ 👁️ The name of the table should start with the name 'Prod' shouldn't it? Looking at our test env the table name starts with 'Test'

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/provider/api.py
landonshumway-ia: If that's the case, we should leave the status field in place for now... with a big fat TODO for us to remove it. As this stands we won't be able to actually merge this for quite some time until the frontend is updated. Were you planning to just leave the PR in a pending state until the front end changes were ready to be merged? I agree that this is a good change to have, but it doesn't seem like we're quite ready to rip this field out just yet.

backend/compact-connect/lambdas/python/migration/667_three_license_status_fields/main.py
landonshumway-ia: Ok, in that case perhaps we should consider something a little more 'demo' worthy 😄 if the intent is to show it to the EDs for example.

backend/compact-connect/lambdas/python/migration/667_three_license_status_fields/main.py
landonshumway-ia: That makes sense. Yeah now that we have a couple examples it would be good to use one of those.

backend/compact-connect/lambdas/python/provider-data-v1/handlers/ingest.py
landonshumway-ia: Yep, that makes sense. I agree this section is going to be updated again once we get more clarification on this.

backend/compact-connect/bin/purge_provider_table.py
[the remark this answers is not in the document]
jusdino: Yeah, the prod table starts with Prod.

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/provider/api.py
jusdino: I was expecting that we would have to put this on a shelf until the UI piece is ready. There's also the fact that privilege purchase validation is now being based on `compactEligibility` instead of `status`, which will make for subtle inconsistencies anyway. If we want to get this out sooner rather than later, I could add it back and make this update a _less_ breaking of a change. That's just not what I was expecting.

landonshumway-ia: Ok, let's discuss at standup this morning. My concern is that if this is sitting for a while it's going to get stale and cause conflicts with all the adverse action items we will be working on soon, so it would be good to get the frontend devs insight into how soon these changes are going to be implemented on their side.

backend/compact-connect/tests/resources/snapshots/QUERY_PROVIDERS_RESPONSE_SCHEMA.json
```diff
@@ -252,8 +273,10 @@
           "providerId",
           "givenName",
           "familyName",
-          "jurisdictionStatus",
-          "status",
+          "licenseStatus",
+          "compactEligibility",
+          "jurisdictionUploadedLicenseStatus",
+          "jurisdictionUploadedCompactEligibility",
```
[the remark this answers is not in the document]
jusdino: This is a good find, however, the `status` field is deprecated and is only being added back in temporarily, to give the web team time to migrate to the new `licenseStatus` and `providerEligibility` fields. Given that the field is now deprecated, we're leaving it out of the API documentation.

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
```diff
@@ -852,8 +837,8 @@ def deactivate_privilege(
         """
         Deactivate a privilege for a provider in a jurisdiction.
 
-        This will update the privilege record to have a persistedStatus of 'inactive' and will remove the jurisdiction
-        from the provider's privilegeJurisdictions set.
+        This will update the privilege record to have a administratorSetStatus of 'inactive' and will remove the
+        jurisdiction from the provider's privilegeJurisdictions set.
```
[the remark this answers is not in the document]
jusdino: Good find. Removing that statement.

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/provider/record.py
```diff
@@ -51,11 +60,11 @@ class ProviderRecordSchema(CalculatedStatusRecordSchema):
 
     # Generated fields
     birthMonthDay = String(required=False, allow_none=False, validate=Regexp('^[0-1]{1}[0-9]{1}-[0-3]{1}[0-9]{1}'))
-
-    # Generated fields
     privilegeJurisdictions = Set(String, required=False, allow_none=False, load_default=set())
     providerFamGivMid = String(required=False, allow_none=False, validate=Length(2, 400))
     providerDateOfUpdate = DateTime(required=True, allow_none=False)
+    licenseStatus = ActiveInactive(required=True, allow_none=False)
+    compactEligibility = CompactEligibility(required=True, allow_none=False)
```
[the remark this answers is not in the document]
jusdino: Our migration included in this PR guarantees that all records will have these values, once it is complete.

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/provider/api.py
```diff
@@ -37,8 +38,10 @@ class ProviderGeneralResponseSchema(ForgivingSchema):
     dateOfUpdate = Raw(required=True, allow_none=False)
     compact = Compact(required=True, allow_none=False)
     licenseJurisdiction = Jurisdiction(required=True, allow_none=False)
+    licenseStatus = ActiveInactive(required=True, allow_none=False)
+    compactEligibility = CompactEligibility(required=True, allow_none=False)
```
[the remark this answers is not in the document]
jusdino: We opted to define these api response schema in flat non-hierarchical structures so that all of the fields allowed out for each case are explicitly spelled out, reducing the chances of confusion, but raising the effort involved in making a common change across the schema. With that in mind, this duplication is deliberate and appropriate.

landonshumway-ia (approved):
Looks good. @jlkravitz This is ready for your review. Thanks

backend/compact-connect/bin/generate_mock_license_csv_upload_file.py
```diff
@@ -27,7 +27,9 @@
 
 os.environ['COMPACTS'] = json.dumps(COMPACTS)
 os.environ['JURISDICTIONS'] = json.dumps(JURISDICTIONS)
+os.environ['ENVIRONMENT_NAME'] = 'test'
```
jlkravitz, line 30: If this is true, can we add this comment here, too?

```
# The environment name has no bearing on the staff user creation process, but we need a value to be set
# for the data model to work.
```

backend/compact-connect/bin/generate_mock_license_csv_upload_file.py
```diff
@@ -150,15 +154,30 @@ def _set_dates(license_data: dict) -> dict:
         date_of_renewal = max(date_of_issuance, now - randint(1, 20) * timedelta(days=365))
         # Their license expired a year after renewal, but no later than yesterday.
         date_of_expiry = min(date_of_renewal + timedelta(days=365), now - timedelta(days=1))
+
+    # Licensees can only be eligible if they are also active
+    if active == ActiveInactiveStatus.ACTIVE:
```
jlkravitz, line 159: Is ActiveInactiveStatus an enum of type boolean? This might be easier to read as just `if active:`

backend/compact-connect/bin/purge_provider_table.py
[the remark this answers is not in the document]
jlkravitz: CodeRabbit suggested adding the below, which I think is a good idea.

```python
if table_name.startswith('Prod'):
    raise ValueError('This script should not be run against production tables. Aborting.')
```

backend/compact-connect/docs/README.md
```diff
@@ -45,7 +45,9 @@ The following table describes all available fields for the license CSV file. Req
 | homeAddressStreet1* | First line of provider's street address | String (max 100 chars) | 123 Main St |
 | licenseType* | Type of professional license. Types you provide must be associated with the compact you are uploading for. | One of: `audiologist`, `speech-language pathologist`, `occupational therapist`, `occupational therapy assistant`, `licensed professional counselor` | occupational therapist |
 | ssn* | Social Security Number | Format: XXX-XX-XXXX | 123-45-6789 |
-| status* | Current status of the licensee. "active" means they will be allowed to purchase privileges that expire on the same day as their license. "inactive" means they are not eligible to purchase privileges. | One of: `active`, `inactive` | active |
+| licenseStatus* | Current status of the license. "active" means they are allowed to practice their profession. *Note: licenses will automatically be displayed as `inactive` on or after their date of expiration, even if the last updload still showed them as `active`.* | One of: `active`, `inactive` | active |
```
jlkravitz, line 48: I'm probably misremembering, but didn't @isabeleliassen mention that licenses _are_ active on their expiration date?

```suggestion
| licenseStatus* | Current status of the license. "active" means they are allowed to practice their profession. *Note: licenses will automatically be displayed as `inactive` after their date of expiration, even if the last upload still showed them as `active`.* | One of: `active`, `inactive` | active |
```

backend/compact-connect/docs/README.md
```diff
@@ -45,7 +45,9 @@ The following table describes all available fields for the license CSV file. Req
...
+| licenseStatusName | An optional more descriptive name of the license status. | String (max 100 chars) | SUSPENDED |
+| compactEligibility* | Whether this license makes the licensee eligible to participate in the compact. Cannot be `eligible` if licenseStatus is `inactive`. *Note: licenses will automatically be displayed as `ineligible` on or after their date of expiration, even if the last upload still showed them as `eligible`.* | One of: `eligible`, `ineligible` | eligible |
```
jlkravitz, line 50: What does "the" compact mean in this case? Can we spell that out a bit more? Also, are there docs somewhere describing why this may be eligible or ineligible? Could we at least add a sentence, e.g., "...makes the license eligible to participate in the compact based on the compact's requirements."

I just got a little confused seeing a 'compact' field on what should be raw license data from the state. I may be overthinking this, though.

backend/compact-connect/docs/design/README.md
```diff
@@ -207,64 +230,163 @@ across jurisdictions, subject to their eligibility.
...
+   - SSN last four digits (full SSN is stored separately for security)
+   - National Provider Identifier (NPI)
+   - Set of privilege jurisdictions
+   - Status (calculated based on expiration date and jurisdiction status)
+
+2. **License Record** (`license`): Represents professional licenses held by the provider:
+   - License number and type
+   - Issuing jurisdiction
+   - Issuance, renewal, and expiration dates
+   - License status (active/inactive, calculated at load time)
```
jlkravitz, line 280:
```suggestion
   - License status (active/inactive, calculated at load time based on current date and license expiration date)
```

backend/compact-connect/docs/design/README.md
```diff
@@ -207,64 +230,163 @@ across jurisdictions, subject to their eligibility.
...
+   - Provider's name and contact details at time of issuance
+
+3. **Privilege Record** (`privilege`): Represents authorizations to practice in other jurisdictions:
+   - Jurisdiction where privilege is granted
+   - Dates of issuance, renewal, and expiration
+   - Status (active/inactive, calculated at load time)
```
jlkravitz, line 286: How is this calculated?

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
```diff
@@ -878,7 +862,7 @@ def deactivate_privilege(
         privilege_record = privilege_record_schema.load(privilege_record)
 
         # If already inactive, do nothing
-        if privilege_record.get('persistedStatus', 'active') == 'inactive':
+        if privilege_record.get('administratorSetStatus', ActiveInactiveStatus.ACTIVE) == ActiveInactiveStatus.INACTIVE:
```
jlkravitz, line 865: I like the new name!

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/license/api.py
```diff
@@ -31,9 +33,13 @@ class LicensePostRequestSchema(ForgivingSchema):
     ssn = SocialSecurityNumber(required=True, allow_none=False)
     npi = NationalProviderIdentifier(required=False, allow_none=False)
     licenseNumber = String(required=False, allow_none=False, validate=Length(1, 100))
-    # This status field is required when posting a license record. It will be transformed into the
-    # jurisdictionStatus field when the record is ingested.
-    status = ActiveInactive(required=True, allow_none=False)
+    licenseStatusName = String(required=False, allow_none=False, validate=Length(1, 100))
+    # The two fields below are stored in the database as fields, renamed to include 'persisted'
+    # because, when they are returned via the API, the named values are calculated based on
+    # the logic that includes the current time and the license expiration date.
```
jlkravitz, line 39: I am confused by this comment. Where are they "renamed"?

backend/compact-connect/lambdas/python/common/tests/function/test_data_client.py
```diff
@@ -74,20 +44,21 @@ def test_get_provider_garbage_in_db(self):
...
-                compact='aslp',
-                provider_id=provider_id,
-            )
+        # The field should not be allowed out via API
+        resp = client.get_provider(
+            compact='aslp',
+            provider_id=provider_id,
+        )
+        for item in resp['items']:
+            self.assertNotIn('something_unexpected', item)
```
jlkravitz, line 61: The mechanics changed here such that an exception is no longer thrown?

backend/compact-connect/lambdas/python/common/tests/unit/test_data_model/test_data_client.py
```diff
@@ -24,29 +24,6 @@ def setUp(self):
...
-        )
-
-    def test_get_provider_id_not_found(self):
-        # Mock response from DynamoDB for non-existent item
-        self.mock_ssn_table.get_item.return_value = {}
-
-        # Verify it raises CCNotFoundException
-        with self.assertRaises(CCNotFoundException):
-            self.client.get_provider_id(compact='aslp', ssn='123456789')
-
```
jlkravitz, line 49: Why are we removing these tests?

backend/compact-connect/lambdas/python/common/tests/unit/test_data_model/test_schema/test_license.py
```diff
@@ -226,3 +232,38 @@ def test_hash_is_unique(self):
 
         # The hashes should now be different
         self.assertNotEqual(change_hash, schema.hash_changes(schema.dump(alternate_record)))
+
+
+class TestLicenseIngestSchema(TstLambdas):
+    def test_calculated_status_to_persisted(self):
```
jlkravitz, line 238: I'm a bit confused by this test name?

backend/compact-connect/lambdas/python/common/tests/unit/test_provider_record_util.py
```diff
@@ -0,0 +1,137 @@
...
+    def test_find_best_license_raises_exception_when_no_licenses(self):
+        """Test that find_best_license raises an exception when no licenses are provided."""
+        from cc_common.data_model.provider_record_util import ProviderRecordUtility
+        from cc_common.exceptions import CCInternalException
+
+        with self.assertRaises(CCInternalException):
+            ProviderRecordUtility.find_best_license([])
+
+    def test_find_best_license_complex_scenario(self):
+        """Test a complex scenario with multiple licenses of different types and statuses."""
```
jlkravitz, line 110: nit but this seems to be only testing different statues, not types–

```suggestion
        """Test a complex scenario with multiple licenses of different statuses."""
```

backend/compact-connect/lambdas/python/migration/667_three_license_status_fields/main.py
```diff
@@ -0,0 +1,154 @@
...
+                        config.provider_table.update_item(
+                            Key=key,
+                            UpdateExpression='SET #ju_compact_eligibility = :compact_eligibility, '
+                            '#ju_license_status = :license_status, '
+                            '#license_status_name = :license_status_name '
+                            'REMOVE #jurisdiction_status',
+                            ExpressionAttributeValues={
+                                ':license_status': license_status,
+                                ':compact_eligibility': 'eligible' if license_status == 'active' else 'ineligible',
+                                ':license_status_name': 'ACTIVE' if license_status == 'active' else 'INACTIVE',
```
jlkravitz, line 76: should this be ALL CAPS?

backend/compact-connect/lambdas/python/provider-data-v1/handlers/ingest.py
jlkravitz: Agreed this makes sense, though I think worth flagging with a comment at the top of the function. If there are any significant license/privilege updates that are currently not covered, then worth writing those in now as well.

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/__init__.py
```diff
@@ -329,7 +329,9 @@ def _generate_providers(
                     self.mock_context,
                 )
                 # we need to get the provider id from the ssn table so it can be used in the ingest message
-                provider_id = data_client.get_provider_id(compact='aslp', ssn=ssn)
+                provider_id = self._ssn_table.get_item(Key={'pk': f'aslp#SSN#{ssn}', 'sk': f'aslp#SSN#{ssn}'})['Item'][
+                    'providerId'
+                ]
```
jlkravitz, line 334: `get_provider_id` seemed much cleaner- why the change?

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_data_model/test_provider_transformations.py
```diff
@@ -81,12 +81,11 @@ def test_transformations(self, mock_license_preprocessing_queue):
...
-        # We'll use the data client to get the resulting provider id
+        # We'll fetch the provider id from the ssn table
         client = DataClient(self.config)
-        provider_id = client.get_provider_id(
-            compact='aslp',
-            ssn=license_ssn,
-        )
+        provider_id = self._ssn_table.get_item(Key={'pk': f'aslp#SSN#{license_ssn}', 'sk': f'aslp#SSN#{license_ssn}'})[
+            'Item'
+        ]['providerId']
```
jlkravitz, line 88: Here, too.

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_ingest.py
```diff
@@ -65,31 +65,18 @@ def _get_provider_via_api(self, provider_id: str) -> dict:
...
-            event = json.load(f)
-
-        event['pathParameters'] = {'compact': 'aslp'}
-        event['requestContext']['authorizer']['claims']['scope'] = 'openid email stuff aslp/readGeneral'
-        event['body'] = json.dumps({'query': {'ssn': '123-12-1234'}})
-
-        # Find the provider's id from their ssn
-        resp = query_providers(event, self.mock_context)
-        self.assertEqual(resp['statusCode'], 200)
-        provider_id = json.loads(resp['body'])['providers'][0]['providerId']
```
jlkravitz, line 91: This seems like a useful thing to test. Why remove it?

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_ingest.py
```diff
@@ -712,8 +721,8 @@ def test_preprocess_license_ingest_creates_ssn_provider_record(self):
         self.assertEqual({'batchItemFailures': []}, resp)
 
         # Find the provider's id from their ssn
-        provider_id = self.config.data_client.get_provider_id(compact='aslp', ssn=test_ssn)
-
+        provider = self._ssn_table.get_item(Key={'pk': f'aslp#SSN#{test_ssn}', 'sk': f'aslp#SSN#{test_ssn}'})['Item']
+        provider_id = provider['providerId']
```
jlkravitz, line 725: Here, too.

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_provider_users.py
```diff
@@ -194,7 +194,12 @@ def test_post_provider_military_affiliation_sets_previous_record_status_to_inact
 
         event = self._when_testing_post_provider_user_military_affiliation_event_with_custom_claims()
 
-        resp = provider_user_me_military_affiliation(event, self.mock_context)
+        # We'll set the 'current datetime' to a date after the initial military affiliation record
```
jlkravitz, line 197: after the record expires? is uploaded?

```suggestion
        # We'll set the 'current datetime' to a date after the initial military affiliation record expires/was uploaded
```

jlkravitz (commented):
sorry for the delay! a few questions/comments.

backend/compact-connect/bin/generate_mock_license_csv_upload_file.py
jusdino: It's not. It is a StrEnum.

backend/compact-connect/lambdas/python/common/tests/function/test_data_client.py
jusdino: Yes, we switched most of our schema from `StrictSchema`, where we raise an exception if there are any unexpected fields (which provides a lot of safety against data mismatches, for an early development phase of an API) to a `ForgivingSchema`, where we instead just strip off unexpected fields (which provides flexibility and simplifies deliberate data migrations, api enhancements, for an API that is stabilizing and requires more up-time). The latter is more consistent with REST API conventions, where adding a new field to an API is not considered breaking, so doesn't raise exceptions. This test update reflects that change.

backend/compact-connect/lambdas/python/common/tests/unit/test_data_model/test_data_client.py
jusdino: We removed the tested method, as it is no longer needed (and is also undesired, since it's an access pattern we don't want on our SSN table).

backend/compact-connect/lambdas/python/migration/667_three_license_status_fields/main.py
jusdino: It doesn't really matter. The values are whatever a state IT system may provide, which I'm just guessing seems like something that could be all caps.

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/__init__.py
jusdino: We removed the method from the client - it's not longer a supported access pattern on the SSN table.

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_data_model/test_provider_transformations.py
jusdino: Same.

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_ingest.py
jusdino: We don't support querying by SSN anymore.

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_ingest.py
jusdino: Same.

backend/compact-connect/lambdas/python/provider-data-v1/handlers/ingest.py
jusdino: I'm not sure what comment to leave - this logic is just deliberately simple. @jlkravitz , can you suggest something specific?

backend/compact-connect/bin/purge_provider_table.py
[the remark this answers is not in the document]
jusdino: Of course. Good call. Where did CodeRabbit suggest that? Did I miss it, somewhere?

backend/compact-connect/bin/generate_mock_license_csv_upload_file.py
jlkravitz: Ah. It looks like `active` is set to be `True` or `False` above. Should that be updated?

backend/compact-connect/docs/README.md
```diff
@@ -45,9 +45,9 @@ The following table describes all available fields for the license CSV file. Req
 | homeAddressStreet1* | First line of provider's street address | String (max 100 chars) | 123 Main St |
 | licenseType* | Type of professional license. Types you provide must be associated with the compact you are uploading for. | One of: `audiologist`, `speech-language pathologist`, `occupational therapist`, `occupational therapy assistant`, `licensed professional counselor` | occupational therapist |
 | ssn* | Social Security Number | Format: XXX-XX-XXXX | 123-45-6789 |
-| licenseStatus* | Current status of the license. "active" means they are allowed to practice their profession. *Note: licenses will automatically be displayed as `inactive` on or after their date of expiration, even if the last updload still showed them as `active`.* | One of: `active`, `inactive` | active |
+| licenseStatus* | Current status of the license. "active" means they are allowed to practice their profession. *Note: licenses will automatically be displayed as `inactive` after their date of expiration, even if the last updload still showed them as `active`.* | One of: `active`, `inactive` | active |
```
jlkravitz, line 48: `updload` -> `upload`

backend/compact-connect/docs/README.md
```diff
@@ -45,9 +45,9 @@ The following table describes all available fields for the license CSV file. Req
...
 | licenseStatusName | An optional more descriptive name of the license status. | String (max 100 chars) | SUSPENDED |
-| compactEligibility* | Whether this license makes the licensee eligible to participate in the compact. Cannot be `eligible` if licenseStatus is `inactive`. *Note: licenses will automatically be displayed as `ineligible` on or after their date of expiration, even if the last upload still showed them as `eligible`.* | One of: `eligible`, `ineligible` | eligible |
+| compactEligibility* | Whether this license makes the licensee eligible to participate in the compact based on the compact's requirements. Cannot be `eligible` if licenseStatus is `inactive`. *Note: licenses will automatically be displayed as `ineligible` on or after their date of expiration, even if the last upload still showed them as `eligible`.* | One of: `eligible`, `ineligible` | eligible |
```
jlkravitz, line 50: Should this also say, "~on or~ after their date of expiration"?

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/license/api.py
```diff
@@ -34,8 +34,8 @@ class LicensePostRequestSchema(StrictSchema):
     npi = NationalProviderIdentifier(required=False, allow_none=False)
     licenseNumber = String(required=False, allow_none=False, validate=Length(1, 100))
     licenseStatusName = String(required=False, allow_none=False, validate=Length(1, 100))
-    # The two fields below are stored in the database as fields, renamed to include 'persisted'
-    # because, when they are returned via the API, the named values are calculated based on
+    # The two fields below are stored in the database as fields, renamed in the ingest chain to include
+    # 'jurisdictionUploaded' because, when they are returned via the API, the named values are calculated based on
```
jlkravitz, line 38: Sorry to be nit-picky, this still took a minute for me to parse. If the below is accurate, perhaps we can use something like it instead?

```python
# Note that the two fields below, `licenseStatus` and `compactEligibility`, are stored 
# in the database as `jurisdictionUploadedLicenseStatus` and `jurisdictionUploadedCompactEligibility`.
# This is to distinguish them from the `licenseStatus` and `compactEligibility` fields returned via the 
# API, which are dynamically calculated based on logic that includes the current time and the
# license expiration date.
```

backend/compact-connect/bin/purge_provider_table.py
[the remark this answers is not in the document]
jlkravitz: It seems to be selective with the feedback it leaves as individual Github comments vs. as a big chunk of "nitpick comments" embedded within a single Github comment. There's a comment from CodeRabbit that looks like the below. If you expand, you can see a list of additional comments it has.

image

backend/compact-connect/lambdas/python/provider-data-v1/handlers/ingest.py
jlkravitz: Oh my bad. I agree and was not thinking straight in my late-night stupor.

jlkravitz (commented):
few follow-up items

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/license/api.py
jusdino: Yeah, that's clearer. Thanks!

jusdino:
Good morning @jlkravitz , any remaining items?

jlkravitz (approved):
@isabeleliassen Good to merge, in coordination with @jusdino since this is a breaking change (not sure if we need to take care to merge?).

Issue #705, opened 2025-04-02
Create 'Beta' Environment for state and compact onboard testing


landonshumway-ia:
As we prepare to onboard more state IT departments so they can integrate their systems with CompactConnect, we need another environment for these states to upload their production data into with the expectation that if something breaks, the data in the CompactConnect database for that environment can be deleted without the risk of downtime. 

This involves updating the pipeline to add a new 'Beta' environment (final environment name TBD) which the pipeline will automatically deploy to whenever a change is merged to the main branch, to keep the prod and 'beta' environments in sync in regards to available features.

This will require creating a new AWS account in the CSG organization. It will also require changes to both the GitHub actions configuration for the frontend deployment, as well as the CDK code for adding the new pipeline environment.

### Estimate
3

### Tasks
- [ ] TODO
- [ ] Automated tests
- [ ] API Docs
- [ ] Postman collection
- [ ] PR opened with labels / reviewers / assignee / linked-issue

landonshumway-ia:
As part of this change, we also need to update the CSP lambda to support the beta environment.

landonshumway-ia:
Marking this as blocked for now since we cannot deploy the Beta and Prod pipeline resources until the next Sprint PR with the pipeline change is merged. The following steps will need to be performed for the beta and prod environments once the main branch contains the pipeline change:

```
- Bootstrap deploy backend pipelines, let them run through deployment
- Bootstrap deploy frontend pipelines, let them run through deployment

Test the following elements of each environment
- User login (staff, provider)
- Provider registration
- License upload

- Request beta account be removed from SES Sandbox
- Request prod account be removed from SES Sandbox (if not done so already)
```

Issue #580, opened 2025-02-20
Move UI building into CDK pipeline


jusdino:
We'll move the building and distributing of the UI resources into CDK, so we can provide environment-specific configuration directly into the UI build, rather than having an extra step of maintaining GitHub secrets. This will also automate deploying of UIs to developer sandbox environments.

landonshumway-ia:
As part of this change, we need to ensure that we still invalidate the CloudFront cache on deploy

Issue #424, opened 2025-01-06
Split backend pipelines into separate Stacks


jusdino:
We currently define both the *test* and *prod* pipelines in the same Stack, which is controlled by one branch (`development` in this case). When we have a situation where there is a stack in `development` that doesn't exist in `main` yet, the *prod* pipeline winds up trying to deploy a stack that doesn't exist yet. This two pipelines, one stack approach works better in projects where any differences between `development` and `main` are minor and short-lived, but the complexity and review process of this project is straining its limits.

To better support larger, longer-lived differences between `development` and `main` branches, we should split the two pipelines into two stacks, each self-mutated based on triggers from their respective branches.

PR #731, opened 2025-04-12
Reference data event bus through SSM parameter

landonshumway-ia:
While testing the pipeline migration, we hit an issue where the event bus needs to be renamed, as it is using the pipeline stack as part of its name. This creates a cross stack reference issue since it is being referenced by several other stacks, and you cannot change a stack export value while it is still being referenced by other stacks. In order to address this, this PR replaces the direct reference with an SSM parameter, which can be changed without causing the stacks to fail deployment. This PR will need to be merged and deployed first, before the subsequent pipeline change can be run through.

This change is a precursor to the following tickets: https://github.com/csg-org/CompactConnect/issues/705 https://github.com/csg-org/CompactConnect/issues/580 https://github.com/csg-org/CompactConnect/issues/424

## Summary by CodeRabbit

- **New Features**
	- Centralized configuration management for event integration has been introduced, enabling uniform retrieval of event bus parameters across components.
- **Refactor**
	- Updated several service integrations to rely on the new parameter-based configuration approach, improving decoupling and consistency in event operations.
	- Introduced private attributes for better encapsulation and management of event bus resources.

### DIFF
--- a/backend/compact-connect/stacks/persistent_stack/__init__.py
+++ b/backend/compact-connect/stacks/persistent_stack/__init__.py
@@ -109,12 +109,13 @@ class PersistentStack(AppStack):
 
         # This resource should not be reference directly as a cross stack reference, any reference should
         # be made through the SSM parameter
-        self._data_event_bus = EventBus( self, 'DataEventBus')
+        self._data_event_bus = EventBus(self, 'DataEventBus')
         # We Store the data event bus name in SSM Parameter Store
         # to avoid issues with cross stack references due to the fact that
         # you can't update a CloudFormation exported value that is being referenced by a resource in another stack.
-        self.data_event_bus_arn_ssm_parameter = (SSMParameterUtility
-                                                 .set_data_event_bus_arn_ssm_parameter(self, self._data_event_bus))
+        self.data_event_bus_arn_ssm_parameter = SSMParameterUtility.set_data_event_bus_arn_ssm_parameter(
+            self, self._data_event_bus
+        )
         # TODO - these are needed until pipeline migration effort is complete  # noqa: FIX002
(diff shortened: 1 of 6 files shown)


jusdino (approved):
Looks great! 🚀

backend/compact-connect/stacks/persistent_stack/__init__.py
```diff
@@ -106,7 +107,18 @@ def __init__(
             auto_delete_objects=removal_policy == RemovalPolicy.DESTROY,
         )
 
-        self.data_event_bus = EventBus(self, 'DataEventBus')
+        # This resource should not be reference directly as a cross stack reference, any reference should
```
jlkravitz, line 110:
```suggestion
        # This resource should not be referenced directly as a cross stack reference, any reference should
```

jlkravitz (commented):
one typo, otherwise looks good.

Is this PR blocking others? If not, it might be more straightforward to merge this in with the sprint PR (next week, say). If it is blocking, though, we can merge per your suggestion to both `development` and `main`. @landonshumway-ia

landonshumway-ia:
@jlkravitz Unfortunately, this PR is a blocker for our pipeline migration effort, which we hope to roll out no later than this week as several states are looking to test out the system in the beta environment. So we will want to get this rolled out to both branches sooner rather than later. Likewise, we have another PR upcoming for the actual pipeline architecture itself, which will also need to follow a similar process of being pulled into development as well as the main branch in order to get everything wired up correctly.

jlkravitz (approved):
@isabeleliassen good to merge this!

isabeleliassen (approved):

PR #736, opened 2025-04-15
Reference data event bus through SSM parameter (#731)

landonshumway-ia:
While testing the pipeline migration, we hit an issue where the event bus needs to be renamed, as it is using the pipeline stack as part of its name. This creates a cross stack reference issue since it is being referenced by several other stacks, and you cannot change a stack export value while it is still being referenced by other stacks. In order to address this, this PR replaces the direct reference with an SSM parameter, which can be changed without causing the stacks to fail deployment. This PR will need to be merged and deployed first, before the subsequent pipeline change can be run through.

This change is a precursor to the following tickets: https://github.com/csg-org/CompactConnect/issues/705 https://github.com/csg-org/CompactConnect/issues/580 https://github.com/csg-org/CompactConnect/issues/424

### DIFF
--- a/backend/compact-connect/common_constructs/ssm_parameter_utility.py
+++ b/backend/compact-connect/common_constructs/ssm_parameter_utility.py
@@ -1,0 +1,46 @@
+from aws_cdk.aws_events import EventBus
+from aws_cdk.aws_ssm import StringParameter
+from constructs import Construct
+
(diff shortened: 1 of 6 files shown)


jlkravitz (approved):
@isabeleliassen Good to merge this.

isabeleliassen (approved):

jlkravitz:
@landonshumway-ia looks like tests are failing. mind taking a look?

PR #752, opened 2025-04-21
fix/Prevent sandbox UI without domain name

landonshumway-ia:
If a developer wants to test the UI deployments in their sandbox environment, they must have a domain name configured. This adds the needed check to prevent developers from deploying the UI stack if they are missing a domain name field.

## Summary by CodeRabbit

- **Documentation**
  - Clarified deployment instructions for the sandbox UI, including guidance for running locally when no domain name is configured.

- **Bug Fixes**
  - Added a check to prevent deployment of the sandbox UI without a configured domain name, providing a clear error message and alternative instructions.


jlkravitz (approved):

isabeleliassen (approved):

Issue #786, opened 2025-05-01
"end military status" only shows up for someone who has an actively enabled military status


isabeleliassen:

jsandoval81:
- Also fix keyboard nav for Enter on this element. Currently it flashes during keyboard nav.

jsandoval81:
Also, look into effort to detect & show user if they are in the "stuck initializing" state of their API request succeeded but their document upload is pending. A simple message on this page would suffice for now.

jsandoval81:
Also:
- Handle S3 XML errors
- Set frontend max to 8 or 9 MB (S3 config has a hard limit of ~10 MB)

[Reported in Sprint 23 PR](https://github.com/csg-org/CompactConnect/pull/790#discussion_r2078630321)

jsandoval81:
Also, as part of the state config FE ticket #782 it was commented:

> As part of this, we will need to update the deprecated military discount related logic to use the new military rate logic. Rather than subtracting the military discount amount from the base rate, we will now just show the new 'militaryRate' amount in place of the standard rate.

It may be better to handle that frontend update in this ticket since it's already touching the military handling.

ChiefStief:
Allow ending of military affiliation when stuck in initializing state (likely backend bug to fix)

PR #790, opened 2025-05-05
Sprint 23

jlkravitz:

### DIFF
--- a/backend/compact-connect/bin/create_staff_user.py
+++ b/backend/compact-connect/bin/create_staff_user.py
@@ -45,9 +45,9 @@ user_table = boto3.resource('dynamodb').Table(USER_TABLE_NAME)
 schema = UserRecordSchema()
 
 
-def create_compact_ed_user(*, email: str, compact: str, user_attributes: dict):
-    sys.stdout.write(f"Creating Compact ED user, '{email}', in {compact}")
-    sub = create_cognito_user(email=email)
+def create_compact_ed_user(*, email: str, compact: str, user_attributes: dict, permanent_password: str | None = None):
+    sys.stdout.write(f"Creating Compact ED user, '{email}', in {compact}\n")
+    sub = create_cognito_user(email=email, permanent_password=permanent_password)
     user_table.put_item(
         Item=schema.dump(
             {
@@ -62,9 +62,11 @@ def create_compact_ed_user(*, email: str, compact: str, user_attributes: dict):
     )
 
 
-def create_board_ed_user(*, email: str, compact: str, jurisdiction: str, user_attributes: dict):
-    sys.stdout.write(f"Creating Board ED user, '{email}', in {compact}/{jurisdiction}")
-    sub = create_cognito_user(email=email)
+def create_board_ed_user(
+    *, email: str, compact: str, jurisdiction: str, user_attributes: dict, permanent_password: str | None = None
+):
+    sys.stdout.write(f"Creating Board ED user, '{email}', in {compact}/{jurisdiction}\n")
+    sub = create_cognito_user(email=email, permanent_password=permanent_password)
     user_table.put_item(
         Item=schema.dump(
             {
@@ -79,7 +81,13 @@ def create_board_ed_user(*, email: str, compact: str, jurisdiction: str, user_at
     )
 
 
-def create_cognito_user(*, email: str):
+def create_cognito_user(*, email: str, permanent_password: str | None):
+    """Create a Cognito user with the given email address and password.
+
+    If provided, sets the password as the user's permanent password. Since this circumvents default password policies
+    (i.e., password reset), this should only be used in testing/sandbox environments.
+    """
+
     def get_sub_from_attributes(user_attributes: list):
         for attribute in user_attributes:
             if attribute['Name'] == 'sub':
@@ -87,18 +95,35 @@ def create_cognito_user(*, email: str):
         raise ValueError('Failed to find user sub!')
 
     try:
+        # By including the TemporaryPassword on user creation, we avoid creating a user if the desired permanent
+        # password does not adhere to the password policy. Either no user is created, or a user is created with
+        # the desired password.
+        kwargs = {'TemporaryPassword': permanent_password} if permanent_password is not None else {}
         user_data = cognito_client.admin_create_user(
             UserPoolId=USER_POOL_ID,
             Username=email,
             UserAttributes=[{'Name': 'email', 'Value': email}],
             DesiredDeliveryMediums=['EMAIL'],
+            **kwargs,
         )
+
+        if permanent_password is not None:
+            cognito_client.admin_set_user_password(
+                UserPoolId=USER_POOL_ID, Username=email, Password=permanent_password, Permanent=True
+            )
         return get_sub_from_attributes(user_data['User']['Attributes'])
 
     except ClientError as e:
         if e.response['Error']['Code'] == 'UsernameExistsException':
+            sys.stdout.write('User already exists, returning existing user')
             user_data = cognito_client.admin_get_user(UserPoolId=USER_POOL_ID, Username=email)
             return get_sub_from_attributes(user_data['UserAttributes'])
+        if e.response['Error']['Code'] == 'InvalidPasswordException':
+            sys.stderr.write(f'Invalid password: {e.response["Error"]["Message"]}')
+            sys.exit(2)
+        else:
+            sys.stderr.write(f'Failed to create user: {e.response["Error"]["Message"]}')
+            sys.exit(2)
 
 
 if __name__ == '__main__':
--- a/backend/compact-connect/bin/sandbox_bootstrap_staff_users.py
+++ b/backend/compact-connect/bin/sandbox_bootstrap_staff_users.py
@@ -1,0 +1,71 @@
+#!/usr/bin/env python3
+"""Script to bootstrap staff users in a sandbox environment with predefined credentials to simplify testing.
+Run this script from `backend/compact-connect`.
+
+The AWS CLI must be configured with AWS credentials that have appropriate access to Cognito and DynamoDB.
+"""
+
+import os
+import sys
+from argparse import ArgumentParser
+
+import sandbox_fetch_aws_resources
+
+SANDBOX_USER_PASSWORD = 'Test12345678'  # noqa: S105 this script is used in sandbox environments only
+
+
+def bootstrap_board_ed_user(compact: str, jurisdiction: str, email_username: str, email_domain: str):
+    email = f'{email_username}+board-ed-{compact}-{jurisdiction}@{email_domain}'
+    _user_attributes = {
+        'email': email,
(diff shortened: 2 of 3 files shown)


backend/compact-connect/app_clients/README.md
```diff
@@ -59,7 +59,10 @@ Add a new app client yaml file to `/app_clients` following the schema of the exa
 
    This command creates an app client with the ability to generate access tokens that expire after 15 minutes. This expiration time can be adjusted according to the needs of the consuming team up to 1 day (see [AccessTokenValidity](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_CreateUserPoolClient.html#CognitoUserPools-CreateUserPoolClient-request-AccessTokenValidity)), though it is strongly recommended to keep this value as short as possible to limit the amount of time an access token is valid for if it is compromised. The consuming team will need to implement logic to generate new access tokens before previous tokens expire.
 
-   If the consuming team plans to use both the test and production environments, you will need to create two separate app clients, one in each respective AWS accounts for those environments.
+   If the consuming team plans to use both the beta and production environments, you will need to create two separate app clients- one in each respective AWS account.
```
jlkravitz, line 62: # (6)

Military affiliation documentation seems to when the file is too large. This is expected, but it might be nice to show a better error message.

image

It seems like the backend returns XML, like 

```xml
<Error><Code>EntityTooLarge</Code><Message>Your proposed upload exceeds the maximum allowed size</Message><ProposedSize>10489621</ProposedSize><MaxSizeAllowed>10485760</MaxSizeAllowed><RequestId>ZY452TMSZHVTP4NF</RequestId><HostId>l3MVMSUvGnS49ceQWxS0F0lhwVK6jOVB0RZ9PxzeUpWBgCVPtyutchRSw/Jhe5Z+CWZuFG07exM=</HostId></Error>
```

backend/compact-connect/app_clients/README.md
```diff
@@ -59,7 +59,10 @@ Add a new app client yaml file to `/app_clients` following the schema of the exa
...
+
+   The cognito token URL for the beta environment is `https://compact-connect-staff-beta.auth.us-east-1.amazoncognito.com/oauth2/token`
```
jlkravitz, line 64: # (7)

After failing to upload my military affiliation documentation, I do not have military affiliation (expected).

If I try to purchase a privilege, I get this error message, though, which seems confusing given I never uploaded anything and my military affiliation is inactive. I tried explicitly "ending" my military affiliation, but I still wasn't able to purchase a privilege.

image

backend/compact-connect/app_clients/README.md
```diff
@@ -59,7 +59,10 @@ Add a new app client yaml file to `/app_clients` following the schema of the exa
...
 
-   If the consuming team plans to use both the test and production environments, you will need to create two separate app clients, one in each respective AWS accounts for those environments.
+   If the consuming team plans to use both the beta and production environments, you will need to create two separate app clients- one in each respective AWS account.
+
+   The cognito token URL for the beta environment is `https://compact-connect-staff-beta.auth.us-east-1.amazoncognito.com/oauth2/token` 
+   The cognito token URL for the prod environment is `https://compact-connect-staff.auth.us-east-1.amazoncognito.com/oauth2/token`
```
jlkravitz, line 66: # (8)

Just FYI, but I'm on train wifi, which goes in and out, and I somehow ended up in this indeterminate state on the frontend. I was a logged in provider and was able to purchase privileges before this happened.

image

**Update:** Once wifi was back, a refresh seems to have done the trick. So really just FYI, perhaps there's room for a smoother experience here on bad connections.

jlkravitz (commented):
@jusdino @jsandoval81 @landonshumway-ia @ChiefStief 

Two items related to military affiliation might involve backend work.

Several other items relate to keyboard navigation (space vs enter), so it may be worth discussing our broader strategy / philosophy on that before proceeding so we can take a uniform approach in future sprints. Let me know what you think.

My preference would be to hotfix (1) and (2), and (7) if it's not too much work.

backend/compact-connect/app_clients/README.md
landonshumway-ia: @jlkravitz This XML response is coming from AWS S3. The CompactConnect API returns temporary upload credentials for the client to upload their file, if they upload a file that is too large, the best we could probably do is to have the frontend display the entire XML response so the user could determine the cause of the error and see the file is too large. Our API doesn't have any control over the response from S3 in this case. We can only set the max file size we allow to be uploaded.

backend/compact-connect/app_clients/README.md
landonshumway-ia: @jlkravitz If the user attempted to start the military affiliation process but the document was not uploaded, your user's military status is stuck in an initializing state, we do not allow you to purchase privileges until you go back and complete the document upload process as we cannot determine if you are active military and need to apply the military discount. If you attempted to end the military status and then still could not purchase privileges, then that is a bug on the backend. We will need to create a ticket to address that. My guess is that because your status is stuck in 'initializing', the backend is not deactivating the military data record on your user account. This was not a case we anticipated, but will want to address.

jsandoval81: @landonshumway-ia After looking at this, I think the backend should just interpret this state as "not military affiliated". The purchase should probably just succeed in this case, without any military discount / fee.

landonshumway-ia: Unfortunately, this is not the only way that the user's military status can get stuck in the 'initializing' state. There is also the possibility that AWS is having an outage, and the file upload process doesn't trigger the lambda that finalizes the user's military state. If we were to ignore this warning and move forward with the purchase, the user may believe they are getting the military rate when they will not. We spent a solid few days walking through the implications of this military purchase flow, and determined that if they are stuck in a initializing state, we don't want them to move forward until they go back and either cancel their affiliation, or upload a valid record.

landonshumway-ia: So given that the backend endpoint is not successfully setting their 'initializing' status to inactive is something we need to address.

backend/compact-connect/app_clients/README.md
```diff
@@ -20,7 +20,7 @@ Add a new app client yaml file to `/app_clients` following the schema of the exa
 
 ##### **Compact-Level Scopes:**
    These are the scopes that are scoped to a specific compact. Granting these scopes will allow the app client to perform actions across all jurisdictions within that compact.
-   Generally, the only scope that should be granted at the compact level is the `{compact}/readGeneral` scope.
+   Generally, the only scope that should be granted at the compact level is the `{compact}/readGeneral` scope if needed.
```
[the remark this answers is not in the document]
jsandoval81: @jlkravitz Forms on most web pages just attempt to execute the first button when clicking "enter". Example in JCC app is the licensing search form - if you click enter while highlighted on an form element it will submit the form since submit is the first button.

In this case, the "Add state..." element is actually a button under the hood (for semantics), and so "enter" is operating on that first button. As I'm navigating this form this feels reasonable / expected.

Typically space or arrow keys would be used to interact with select drop-downs. You can also use character keys to try and match select options.

backend/compact-connect/app_clients/README.md
```diff
@@ -20,7 +20,7 @@ Add a new app client yaml file to `/app_clients` following the schema of the exa
...
 
    The following scopes are available at the compact level:
    ```
```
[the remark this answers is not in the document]
jsandoval81: @jlkravitz I'm actually not 100% sure the universally expected behavior here. You could be right that it should mimic mouse behavior exactly. I created #803 to track.

backend/compact-connect/app_clients/README.md
```diff
@@ -59,7 +59,10 @@ Add a new app client yaml file to `/app_clients` following the schema of the exa
 
    This command creates an app client with the ability to generate access tokens that expire after 15 minutes. This expiration time can be adjusted according to the needs of the consuming team up to 1 day (see [AccessTokenValidity](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_CreateUserPoolClient.html#CognitoUserPools-CreateUserPoolClient-request-AccessTokenValidity)), though it is strongly recommended to keep this value as short as possible to limit the amount of time an access token is valid for if it is compromised. The consuming team will need to implement logic to generate new access tokens before previous tokens expire.
```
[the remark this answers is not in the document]
jsandoval81: I also see that the enter is not behaving as expected here. We're going to be touching this exact UI this sprint via #786  - I've added this as an item to that ticket.

_For semantics, this element is actually a button (rather than an anchor link). Buttons generally respond to enter or space._

backend/compact-connect/app_clients/README.md
jsandoval81: @landonshumway-ia @jlkravitz We may need to work this in a new ticket. I think there are some discussions worth having about how to close this gap, that could involve the backend & frontend. For instance:

1. We may want to figure out how to message the user their military status is incomplete, but still allow them to make purchases if they want?
1. It may be worth changing the expected order of API interaction to attempt to upload documents first, then make the call to our API if upload succeeded?
1. Other ideas / separating out MVP vs. longer-term updates / etc.

backend/compact-connect/app_clients/README.md
jsandoval81: @jlkravitz Similar to a previous item, we're going to be touching the military UI in #786 shortly, so I've added this as an item in that ticket.

backend/compact-connect/app_clients/README.md
jsandoval81: @jlkravitz I created #805 to track this.

backend/compact-connect/app_clients/README.md
jsandoval81: @landonshumway-ia @jlkravitz I created #807 to track this, with the `refinement needed` label.

backend/compact-connect/app_clients/README.md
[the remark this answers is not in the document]
jlkravitz: Ah ok - good to know. That makes sense.

backend/compact-connect/app_clients/README.md
jlkravitz: Sounds good- thanks all!

backend/compact-connect/app_clients/README.md
jlkravitz: Thanks for the discussion on this one. Tracking this and dealing with it more holistically makes sense to me.

backend/compact-connect/app_clients/README.md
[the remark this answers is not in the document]
jlkravitz: To be honest, I'm not sure, either. This was just my intuition, but it may be worth a (quick) search.

jlkravitz:
@isabeleliassen This is good to merge after merging #806. This is a Sprint PR, so should be merged using a merge commit.

isabeleliassen (approved):

PR #813, opened 2025-05-21
Sprint 24

jlkravitz:


backend/bin/run_python_tests.py
```diff
@@ -22,15 +22,15 @@
 
 # Define the test directories to include
 TEST_DIRS = (
+    'compact-connect/lambdas/python/common',
```
jlkravitz, line 25: @jsandoval81 It seems the print-friendly view on mobile shows the navigation bar still, but the hamburger menu doesn't open when clicked. Mind doing a hotfix for that? (I assume it makes most sense to just remove the navigation menu completely for parity with the desktop version.)

image

backend/bin/run_python_tests.py
```diff
@@ -46,6 +46,29 @@ def get_coverage():
     )
 
 
+def clean_modules():
```
jlkravitz, line 49: @jsandoval81 One small keyboard issue on Safari -

image

backend/bin/run_python_tests.py
jsandoval81: @isabeleliassen I think we're going to be overhauling the print page based on the last sprint review. I'd suggest adding this item to that ticket. Do you happen to know the ticket # for that?

cc @jlkravitz

backend/bin/run_python_tests.py
jsandoval81: @jlkravitz I believe the chrome scenario you described is expected. The space key can be used to select the initial focus item, and the arrow keys can be used to move-focus-and-select.

https://www.w3.org/WAI/ARIA/apg/patterns/radio/

The up-arrow-as-the-first-input-action thing on Safari is a small bummer, in that Apple has implemented it slightly different than other browsers. But I still think it's ok since space-to-initially-select works as expected on Safari too.

jlkravitz: Hmm, I was looking at that page and maybe interpreted it differently. The examples they provide seem to work without spacebar- is there a distinction between the example(s) they provide and ours that I'm not considering?

jsandoval81: @jlkravitz I think tab is just focus, not selection. If you tab to the group, the first item is what will be selected if you then press space.

jlkravitz: @jsandoval81 right, but isn't the group as a whole being focused on tab, rather than the first radio button?

jlkravitz:
Ah gotcha. I was treating outline as equivalent to “focused” but that may
be wrong. I’ll research and think on it, but let’s keep as-is for now

backend/bin/run_python_tests.py
jlkravitz: Ah gotcha. I was treating outline as equivalent to “focused” but that may be wrong. I’ll research and think on it, but let’s keep as-is for now

jlkravitz:
@isabeleliassen This is good to merge (with a merge commit, rather than a squash & merge). I left one comment unresolved to make sure you saw it, but feel free to resolve it when you've read it. Thanks!

backend/bin/run_python_tests.py
isabeleliassen: Yes, added it to that ticket: https://github.com/csg-org/CompactConnect/issues/816

PR #835, opened 2025-06-02
Feat/military affiliation updates

ChiefStief:
### Requirements List
-

### Description List
-

### Testing List
- Code review
- ???

Closes #


backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
```diff
@@ -608,6 +608,17 @@ def _rollback_privilege_transactions(
...
+    def _get_all_military_affiliation_records_for_provider(self, compact: str, provider_id: str):
+        military_affiliation_records = self.config.provider_table.query(
+            KeyConditionExpression=Key('pk').eq(f'{compact}#PROVIDER#{provider_id}')
+            & Key('sk').begins_with(
+                f'{compact}#PROVIDER#military-affiliation#',
+            ),
+        ).get('Items', [])
+
+        schema = MilitaryAffiliationRecordSchema()
+        return [schema.load(record) for record in military_affiliation_records]
```
landonshumway-ia, line 620: Nit - we are attempting to move towards using the data class pattern for all of our serializing/deserializing rather than referencing the marshmallow schemas directly. We have a data class for military affiliation records that you should be able to leverage:
```suggestion
        return [MilitaryAffiliationData.from_database_record(record) for record in military_affiliation_records]
```
This would mean you are now returning the data class instances, rather than dictionaries. So the 'inactivate_military_affiliation_status' method would need to be updated to account for this. I will show an example for this in the next comment.

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
```diff
@@ -701,28 +712,28 @@ def create_military_affiliation(
...
         """
-        active_military_affiliation_records = self._get_active_military_affiliation_records(compact, provider_id)
+        military_affiliation_records = self._get_all_military_affiliation_records_for_provider(compact, provider_id)
         schema = MilitaryAffiliationRecordSchema()
         with self.config.provider_table.batch_writer() as batch:
-            for record in active_military_affiliation_records:
+            for record in military_affiliation_records:
                 record['status'] = MilitaryAffiliationStatus.INACTIVE.value
                 serialized_record = schema.dump(record)
                 batch.put_item(Item=serialized_record)
```
landonshumway-ia, line 739: Now here, you would use the data class to set the status value using the update method, and then serializing it into a database record:
```suggestion
                record.update({'status': MilitaryAffiliationStatus.INACTIVE.value})
                serialized_record = record.serialize_to_database_record()
                batch.put_item(Item=serialized_record)
```

landonshumway-ia (approved):
This looks good. I just noticed one thing that could improve the implementation and move us towards our desired pattern of using data classes, but if you don't feel it's worth tackling at this time, the current implementation you have works just fine. I leave it up to you to decide if you want to implement the suggested change.

webroot/src/components/PrivilegePurchaseFinalize/PrivilegePurchaseFinalize.ts
```diff
@@ -224,12 +228,17 @@ export default class PrivilegePurchaseFinalize extends mixins(MixinForm) {
         const licenseTypeKey = this.licenseTypeSelected;
 
         this.selectedStatePurchaseDataList.forEach((stateSelected) => {
-            if (stateSelected?.fees?.[licenseTypeKey]) {
-                total += stateSelected.fees[licenseTypeKey];
-            }
+            const feeConfig = stateSelected?.fees?.[licenseTypeKey];
+
+            if (feeConfig) {
+                const hasMilitaryRate = Boolean(feeConfig && (feeConfig.militaryRate || feeConfig.militaryRate === 0));
```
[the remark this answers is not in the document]
jsandoval81: Yeah, still getters. Just not sure if the getters would benefit from using a method that does common stuff.

webroot/src/components/MilitaryAffiliationInfoBlock/MilitaryAffiliationInfoBlock.ts
```diff
@@ -174,12 +188,13 @@ class MilitaryAffiliationInfoBlock extends mixins(MixinForm) {
...
+    }
 
-            button.focus();
-        });
+    focusOnModalCancelButton() {
+        const buttonComponent = this.$refs.noBackButton as any;
+        const button = buttonComponent.$refs.button as HTMLElement;
+
+        button.focus();
     }
```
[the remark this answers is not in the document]
ChiefStief: Thats not working for me either, at some point I would like to find the solution to this because it seems like there's gotta be something somewhat obvious

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
```diff
@@ -9,7 +9,11 @@
 from botocore.exceptions import ClientError
 
 from cc_common.config import _Config, config, logger, metrics
-from cc_common.data_model.provider_record_util import ProviderRecordUtility, ProviderUserRecords
+from cc_common.data_model.provider_record_util import (
+    MilitaryAffiliationData,
```
landonshumway-ia, line 13: This `MilitaryAffiliationData` import should actually come from the data class module directly
```
from cc_common.data_model.schema.military_affiliation import MilitaryAffiliationData
```

landonshumway-ia (changes requested):
Looks good, just one quick import fix

landonshumway-ia (approved):

landonshumway-ia:
@jlkravitz This PR is now ready for your review whenever you have capacity. Thanks.

webroot/src/components/MilitaryAffiliationInfoBlock/MilitaryAffiliationInfoBlock.ts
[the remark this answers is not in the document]
jlkravitz: What about `InstanceType<typeof InputButton>`?

webroot/src/components/PrivilegePurchaseFinalize/PrivilegePurchaseFinalize.ts
```diff
@@ -533,6 +541,13 @@ export default class PrivilegePurchaseFinalize extends mixins(MixinForm) {
         formInput.value = formInput.value.replace(/[^\d]/g, '');
     }
 
+    getShouldUseMilitaryRate(feeConfig) {
```
jlkravitz, line 544: Right?

```suggestion
    shouldUseMilitaryRate(feeConfig) {
```

webroot/src/models/Licensee/Licensee.model.ts
```diff
@@ -196,12 +196,18 @@ export class Licensee implements InterfaceLicensee {
...
-    public isMilitary(): boolean {
+    public isMilitaryStatusActive(): boolean {
         // The API does not return the affiliations in the get-all endpoint therefore all users will appear unaffiliated
         // if only that endpoint has been called
         return this.militaryAffiliations?.some((affiliation) => ((affiliation as MilitaryAffiliation).status as any) === 'active') || false;
     }
 
+    public isMilitaryStatusInitializing(): boolean {
+        // The API does not return the affiliations in the get-all endpoint therefore all users will appear unaffiliated
+        // if only that endpoint has been called
```
jlkravitz, line 207: i know this is a c/p of the above comment... but what does this mean?

webroot/src/models/Licensee/Licensee.model.ts
```diff
@@ -196,12 +196,18 @@ export class Licensee implements InterfaceLicensee {
...
+        return this.militaryAffiliations?.some((affiliation) => ((affiliation as MilitaryAffiliation).status as any) === 'initializing') || false;
```
jlkravitz, line 208: We might have discussed this way long ago, but do we need all these casts? I would expect this to work without any casting, no?

jlkravitz (commented):
few comments!

webroot/src/models/Licensee/Licensee.model.ts
ChiefStief: When we make a GET Licensees (Get all) API call the military affiliation objects will not be included in the response, so Licensees retrieved using the GET all endpoint will always appear to have no military affiliation even if they do.

When we make a GET Licensee (Get one) we are returned the military affiliation objects. We are using the same model for both versions of the data so we are commenting on those methods that they may not return the correct values if only the get all endpoints have been called.

That being said, I think the wording of the comment can be tweaked for each method so as to be more specific to the method even though it is always caused by the same lack of military affiliation objects.

jlkravitz: Got it, helpful. Thanks!

PR #840, opened 2025-06-03
Sprint 25

jlkravitz:


backend/compact-connect/common_constructs/queued_lambda_processor.py
```diff
@@ -28,6 +28,7 @@ def __init__(
         batch_size: int,
```
jlkravitz, line 28: Weird.. as a state admin for KY, when i update the jurisprudence information link to use 'http://localhost:3018/', I get a network error when I try to click save. (400 unauthorized)

Any other URL seems to be working fine.

image

backend/compact-connect/common_constructs/queued_lambda_processor.py
```diff
@@ -28,6 +28,7 @@ def __init__(
...
         encryption_key: IKey,
         alarm_topic: ITopic,
+        dlq_count_alarm_threshold: int = 10,
```
jlkravitz, line 31: Do we care about the scenario in which the following occurs–

1. A State ED sets the rate at X
2. A provider purchases a privilege, making it to the final page before clicking "Complete Purchase"
3. State ED sets the rate to Y
4. Provider clicks "Complete Purchase" but has now paid the wrong price.

I'm guessing no, but just wanted to flag.

I'd also flag the same thing for changes in the jurisprudence exam requirement. e.g., what if it starts as not required, a provider gets part way through purchasing but does not testify that they've done the exam, the setting flips, and then the provider completes the purchase. Do we catch that?

backend/compact-connect/common_constructs/queued_lambda_processor.py
```diff
@@ -28,6 +28,7 @@ def __init__(
...
     ):
         super().__init__(scope, construct_id)
```
jlkravitz, line 34: @ChiefStief i don't seem to be getting an email back after purchasing (in this case, renewing) privileges for a provider. any insight?

backend/compact-connect/common_constructs/queued_lambda_processor.py
landonshumway-ia: My understanding is these rates will be updated very infrequently, as updates need to go through an approval process which can take time to complete. Likely if these scenarios you've described become a concern in the future, we can roll out such updates in a maintenance window in which no users are able to purchase privileges.

backend/compact-connect/common_constructs/queued_lambda_processor.py
landonshumway-ia: @jlkravitz Cognito related emails are still sent out from Cognito's domain, even if you don't have a custom domain name configured. Do you have a custom domain for your sandbox environment? If not that would explain why you are not receiving this particular email.

landonshumway-ia: If the README is not clear on the process for setting that up, let us know and we can improve the README instructions as needed.

backend/compact-connect/common_constructs/queued_lambda_processor.py
jlkravitz: Makes sense!

backend/compact-connect/common_constructs/queued_lambda_processor.py
jlkravitz: OK got it working. One thing that tripped me up is needing to go to the `app.` subdomain of the configured domain to view the app.

Could we add that to the README on this PR, perhaps in step 10 (last sentence is my suggested addition)?

```
10) *Optional:* If you have a domain name configured for your sandbox environment, once the backend stacks have successfully deployed, you can deploy the frontend UI by setting the 'deploy_sandbox_ui' context field to `true` and run `cdk deploy 'SandboxUI/*'`. If you do not have a domain name configured, you can still run the UI from local host (see the README.md under the webroot folder for more information about running the app on localhost). If you *do* have a domain name configured, the application will be accessible at the 'app' subdomain of the configured domain name (e.g., app.[configured_domain.com]).
```

backend/compact-connect/common_constructs/queued_lambda_processor.py
jlkravitz: @landonshumway-ia Was this reproducible for you? Mostly want to make sure we're not missing a weird edge case that covers other URLs.

landonshumway-ia: In this case, we should be good to move forward on this, given that other URLs other than localhost are able to pass through the WAF filter. In a future case where some other URL ends up being filtered by this, we can move the payload validation from the API Gateway layer into the lambda layer (where we are already performing a separate schema validation anyway).

backend/compact-connect/common_constructs/queued_lambda_processor.py
landonshumway-ia: Sure I'll add a hotfix PR for this now.

backend/compact-connect/common_constructs/queued_lambda_processor.py
jlkravitz: Sounds good! Thanks for looking into this.

PR #847, opened 2025-06-05
Feat/authorize net payment opaque data

### LINKED ISSUE #793: authorize.net update for compliance FE
From Landon: I've come across some information this morning that is relevant to the payment processor decision and I wanted to make you aware of it. Right now the Compact Connect frontend is passing credit card information directly into our backend system, which then forwards that information to [Authorize.net](http://authorize.net/) It turns out because we are passing this credit card information into our system, we are subject to more stringent [PCI DSS (Payment Card Industry Data Security Standard)](https://www.pcisecuritystandards.org/standards/) compliance requirements, as we are handling sensitive payment data directly. The alternative approach to what we are doing now would be to update the frontend to pass the payment information to [Authorize.net](http://authorize.net/) directly, and then get a token back from them which would then be passed to our backend, so no credit card information would ever touch our API.
If the compact directors do decide to stay with [Authorize.net](http://authorize.net/), we will likely want to make an update to the system so that we do not pass credit card information directly through the backend. There was a similar effort we were going to make as part of moving over to Stripe, which 
[@John](https://csghq.slack.com/team/U9P7WSW5P)
 estimated at about 5 points, so it would likely take the same amount of effort for [Authorize.net](http://authorize.net/)

### LINKED ISSUE #794: authorize.net update for compliance BE

landonshumway-ia:
Previously, our initial design for processing payment information when purchasing privileges involved sending card data directly to the backend, which was then used to call Authorize.net to make a transaction. In order to improve the security of the system, we are updating the front end to send the payment card information to Authorize.net directly using their Accept UI solution, and then sending the payment nonce token to the backend instead. This updates the backend to support that pattern.

As part of supporting this, the frontend will now need the api login id and the public client key to generate the payment collection form. This updates the system so that when a compact admin uploads their payment processing creds into the system, the public key is stored in the compact configuration object, which will then be accessible by the frontend when practitioners go to purchase privileges.

Closes #794 
Closes #793

## Summary by CodeRabbit

- **New Features**
  - Added support for uploading and validating payment processor credentials, including public fields for secure payment collection.
  - Payment privilege purchasing now uses secure opaque payment tokens instead of raw card details.
  - Compact privilege options and purchase APIs now include public payment processor fields and sandbox status.
  - Introduced a new Vue component integrating Authorize.net AcceptUI for payment input.

- **Improvements**
  - Enhanced Content Security Policy to allow necessary Authorize.Net domains for payment processing.
  - Updated API schemas and documentation to reflect new opaque payment data structure and public credential fields.
  - Anchored transaction history processing to scheduled invocation time for replayability.
  - Refactored payment UI and form handling to use opaque payment data and simplify input management.
  - Improved UI behavior by auto-scrolling to action buttons after mock data population.
  - Updated infrastructure to grant Lambda permissions and environment variables for payment processor features.

- **Bug Fixes**
  - Ensured idempotent transaction storage to prevent duplicate records during transaction processing.

- **Tests**
  - Added and updated tests for payment processor credential uploads, privilege purchasing with opaque data, and transaction history idempotency.
  - Expanded smoke and load tests to use secure payment tokens and validate new credential flows.
  - Enhanced unit tests with AWS mocking and coverage for credential validation errors.

- **Chores**
  - Updated environment configuration and test resources to support new payment processor integration and secure data handling.
  - Cleaned up styling and localization keys related to payment inputs.


backend/compact-connect/tests/resources/snapshots/BetaFrontend-FrontendDeploymentStack-UI_DISTRIBUTION_LAMBDA_FUNCTION.json
```diff
@@ -2,7 +2,7 @@
...
...pe-Options\n    headers['x-content-type-options'] = [{\n        key: 'X-Content-Type-Options',\n        value: 'nosniff',\n    }];\n    // X-Frame-Options\n    headers['x-frame-options'] = [{\n        key: 'X-Frame-Options',\n        value: 'DENY',\n    }];\n    // Referrer-Policy\n    headers['referrer-policy'] = [{\n        key: 'Referrer-Policy',\n        value: 'strict-origin-when-cross-origin',\n    }];\n    // Server\n    headers.server = [{\n        key: 'Server',\n        value: 'CompactConnect',\n    }];\n    // Content-Security-Policy\n    setCspHeader(headers);\n};\n\n// ============================================================================\n//                                 LAMBDA ENTRY                               =\n// ============================================================================\nexports.handler = async (event) => {\n    // https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/lambda-event-structure.html\n    const eventRecord = event?.Records[0]?.cf || {};\n    const response = eventRecord.response || {};\n    const responseHeaders = response.headers || {};\n\n    setSecurityHeaders(responseHeaders);\n\n    return response;\n};\n"
```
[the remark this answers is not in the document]
landonshumway-ia: What is your source for this? We could not find any documentation from Authorize.net regarding which domains to add.

jusdino (approved):
Great! No notes ✔️

webroot/src/components/PrivilegePurchaseInformationConfirmation/PrivilegePurchaseInformationConfirmation.ts
```diff
@@ -264,6 +264,10 @@ export default class PrivilegePurchaseInformationConfirmation extends mixins(Mix
         this.formData.homeState.value = true;
 
         this.validateAll({ asTouched: true });
+        await nextTick();
+        const formButtons = document.getElementById('button-row');
+
+        formButtons?.scrollIntoView({ behavior: 'smooth', block: 'center' });
```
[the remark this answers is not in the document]
jsandoval81: Vue refs do not work as smoothly with TypeScript, forcing us to have additional type casts or double type casts. In this particular case, the native `document` method is sufficient.

webroot/src/models/CompactFeeConfig/CompactFeeConfig.model.ts
```diff
@@ -28,6 +35,7 @@ export class CompactFeeConfig implements InterfaceCompactFeeConfigCreate {
     public compactCommissionFeeType? = null;
     public perPrivilegeTransactionFeeAmount? = 0;
     public isPerPrivilegeTransactionFeeActive? = false;
+    public paymentSdkConfig? = {};
```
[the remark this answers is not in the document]
jsandoval81: The `CompactFeeConfig` class implements `InterfaceCompactFeeConfigCreate`, which specifies the type for the `paymentSdkConfig` property. Instantiating the default as `{}` falls within the type structure of `InterfaceCompactFeeConfigCreate`.

backend/compact-connect/stacks/api_stack/v1_api/api_model.py
```diff
@@ -948,6 +898,24 @@ def _purchase_privilege_options_items_schema(self):
...
+                                'publicClientKey': JsonSchema(
+                                    type=JsonSchemaType.STRING,
+                                    description='The public client key for the payment processor',
+                                ),
+                                'apiLoginId': JsonSchema(
+                                    type=JsonSchemaType.STRING,
+                                    description='The API login ID for the payment processor',
+                                ),
+                            },
+                        ),
```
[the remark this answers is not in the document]
landonshumway-ia: This is a response object, which our API controls and we perform internal validation on the values as they are sent out. This definition here is purely for documentation purposes.

webroot/src/components/PrivilegePurchaseFinalize/PrivilegePurchaseFinalize.ts
```diff
@@ -204,24 +176,52 @@ export default class PrivilegePurchaseFinalize extends mixins(MixinForm) {
...
+    get currentCompactPaymentSdkConfig(): PaymentSdkConfig | null {
+        return this.currentCompact?.fees?.paymentSdkConfig || null;
     }
 
-    get privCount(): number {
-        return this.selectedStatePurchaseDataList?.length || 0;
+    get currentCompactCommissionFee(): number | null {
+        return this.currentCompact?.fees?.compactCommissionFee || null;
+    }
+
```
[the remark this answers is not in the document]
jsandoval81: I don't think this method was altered in this PR. The logic we have in place has been tested previously and is in an expected state. I understand the comment here, but I think there are more subtle backend data reasons for this that have not changed. We'll leave as-is for now.

rmolinares (changes requested):
Nice job! Minor questions / comments but I also wanted to mention that I'm not able to view the "Purchase successful" screen because the suggested CC info in the QA notes trigger an error and redirect back to the "Payment Summary" screen.

webroot/src/components/PrivilegePurchaseFinalize/PrivilegePurchaseFinalize.vue
```diff
@@ -6,70 +6,13 @@
 -->
 
 <template>
-    <div class="finalize-privilege-purchase-container">
-        <form class="complete-purchase-form" @submit.prevent="handleSubmit">
+    <div id="finalize-privilege-purchase-container" class="finalize-privilege-purchase-container">
+        <form class="complete-purchase-form" @submit.prevent="() => null">
```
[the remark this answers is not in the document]
jsandoval81: Yeah this is just the first one-off deviation from our normal form handling framework - a very specific case where the "submit" button is actually the element placed on the page required by the authorize.net framework.

webroot/src/components/PrivilegePurchaseFinalize/PrivilegePurchaseFinalize.vue
```diff
@@ -6,70 +6,13 @@
...
-                                    :formInput="formData.zip"
-                                    @input="handleZipInput(formData.cvv)"
-                                />
-                            </div>
-                        </div>
-                    </div>
+                <div class="finalize-purchase-title-row">
+                    <h1 class="finalize-purchase-title">{{$t('payment.payment')}}</h1>
                 </div>
+                <MockPopulate :isEnabled="isMockPopulateEnabled" @selected="mockPopulate" />
```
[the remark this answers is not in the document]
jsandoval81: On this screen, it checks the checkbox near the bottom of the card; it also copies the mock CC number to your clipboard for easy pasting on the credit card form that pops up.

jsandoval81:

_FYI the unresolved comments from @rmolinares above are actually resolved, but neither him nor I have the ability to mark them resolved. If you're able to then it's a green light to do so._

backend/compact-connect/lambdas/nodejs/cloudfront-csp/index.js
```diff
@@ -193,6 +197,9 @@ const setCspHeader = (headers = {}) => {
             buildSrcString('style-src-elem', [
                 'self',
                 'https://fonts.googleapis.com',
+                'https://jstest.authorize.net/',
+                'https://js.authorize.net/',
+                `'sha256-YwWQHXh4Vw0oD2Oo8pV9huEF2sE9mD8i5nZUuHzEg9A='`, // <style> tag injected by authorize.net widget
```
jlkravitz, line 202: Is the authorize.net widget that we're using pinned to a specific version such that it won't change unexpectedly on us?

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/compact/record.py
```diff
@@ -37,6 +38,9 @@ class CompactRecordSchema(BaseRecordSchema):
         allow_none=False,
     )
     licenseeRegistrationEnabled = Boolean(required=True, allow_none=False)
+    # This is not set until a compact admin uploads credentials for their payment processor
```
jlkravitz, line 41: Why can't these all be uploaded "together"?

backend/compact-connect/lambdas/python/purchases/handlers/privileges.py
```diff
@@ -87,6 +87,8 @@ def get_purchase_privilege_options(event: dict, context: LambdaContext):  # noqa
         if item['type'] == JURISDICTION_TYPE:
             serlialized_options.append(JurisdictionOptionsResponseSchema().load(item))
         elif item['type'] == COMPACT_TYPE:
+            # we determine at run-time if the payment processor is running in sandbox mode
+            item['isSandbox'] = True if config.environment_name != 'prod' else False
```
jlkravitz, line 91: nit: small preference for this.

```suggestion
            item['isSandbox'] = config.environment_name != 'prod'
```

backend/compact-connect/lambdas/python/purchases/purchase_client.py
```diff
@@ -14,18 +14,26 @@
...
-AUTHORIZE_DOT_NET_CLIENT_TYPE = 'authorize.net'
 OK_TRANSACTION_MESSAGE_RESULT_CODE = 'Ok'
 MAXIMUM_TRANSACTION_API_LIMIT = 1000
 
+# Authorize.net does not have a clear way to distinguish between an error that is caused by an issue with the card
+# information passed in by the user, and an internal issue caused by the API itself. To account for this, we
+# pulled the list of known issues from their transaction response code lookup and put the list of error codes that are
+# likely caused by the user. These include issues such as an invalid or unsupported card number, the expiration date
+# being expired, or the card being declined. You can review the description of these codes by searching for them at
+# https://developer.authorize.net/api/reference/responseCodes.html
```
jlkravitz, line 34: helpful comment!

webroot/src/components/PrivilegePurchaseFinalize/PrivilegePurchaseFinalize.ts
```diff
@@ -479,86 +366,20 @@ export default class PrivilegePurchaseFinalize extends mixins(MixinForm) {
...
-        this.formData.zip.value = `46214`;
+    async mockPopulate(): Promise<void> {
+        // It's handy to have certain values copied to the clipboard during local development;
+        // The following are applicable depending on the testing scenario:
+        navigator.clipboard.writeText(`5424 0000 0000 0015`); // Test CC number
+        // navigator.clipboard.writeText(`01/30`); // Test CC expiry
+        // navigator.clipboard.writeText(`900`); // Test CC CVV
+        // navigator.clipboard.writeText(`Test`); // Test CC first name
+        // navigator.clipboard.writeText(`User`); // Test CC last name
+        // navigator.clipboard.writeText(`46214`); // Test CC zip code
```
jlkravitz, line 377: Why copy to the clipboard vs actually populating the text fields like other mock populate links?

webroot/src/models/CompactFeeConfig/CompactFeeConfig.model.ts
```diff
@@ -11,12 +11,19 @@ import { deleteUndefinedProperties } from '@models/_helpers';
 // ========================================================
 // =                       Interface                      =
 // ========================================================
+export interface PaymentSdkConfig {
+    loginId?: string;
+    clientKey?: string;
+    isProductionMode?: boolean;
```
jlkravitz, line 17: nit, but any reason to not have this naming align with the backend naming (or vice versa)? e.g., `isSandbox` is the backend name

webroot/src/components/PrivilegePurchaseFinalize/PrivilegePurchaseFinalize.spec.ts
```diff
@@ -17,50 +15,4 @@ describe('PrivilegePurchaseFinalize page', async () => {
         expect(wrapper.exists()).to.equal(true);
         expect(wrapper.findComponent(PrivilegePurchaseFinalize).exists()).to.equal(true);
     });
-    it('should not allow letters or symbols', async () => {
```
jlkravitz, line 20: Does this validation no longer happen on the frontend with the Authorize.net Accept UI?

webroot/README.md
```diff
@@ -109,6 +109,12 @@
     - **`VUE_APP_MOCK_API`** :arrow_heading_down:
         - Only used for local development
         - `true` if mock API should be used
+    - **`VUE_APP_MOCK_API_PAYMENT_LOGIN_ID`** :arrow_heading_down:
+        - Only used for local development
+        - Public login ID from payment gateway, if payment testing locally using mock API
```
jlkravitz, line 114: if not, I assume this can be blank then? (e.g., if testing using a sandbox environment or with any 'real' API)?

webroot/src/components/PrivilegePurchaseAcceptUI/PrivilegePurchaseAcceptUI.ts
```diff
@@ -0,0 +1,160 @@
...
+        };
+    }
+
+    unloadPaymentDetailsUi(): void {
+        (window as any).handlePaymentDetailsResponse = undefined;
+    }
+
+    handleClicked(): void {
+        this.resetMessages();
+        this.adjustFramePosition();
```
jlkravitz, line 117: Mind adding a quick comment here explaining why this is necessary?

backend/compact-connect/tests/smoke/load_test_privilege_purchasing.py
```diff
@@ -165,8 +144,8 @@ def run_load_test(num_iterations: int):
         logger.info(f'Starting iteration {iteration + 1} of {num_iterations}')
 
         try:
-            # Delete existing privileges
-            delete_existing_privileges()
+            # Deactivate existing privileges
+            deactivate_existing_privileges()
```
jlkravitz, line 148: Won't this result in slightly different runtime behavior the first vs. the second time you run this? (My sense was that deleting the privileges provided a clean slate.)

backend/compact-connect/tests/smoke/compact_configuration_smoke_tests.py
```diff
@@ -270,8 +271,69 @@ def test_jurisdiction_configuration():
...
+
+
 if __name__ == '__main__':
     load_smoke_test_env()
-    test_active_member_jurisdictions()
-    test_compact_configuration()
-    test_jurisdiction_configuration()
+    # test_active_member_jurisdictions()
+    # test_compact_configuration()
+    # test_jurisdiction_configuration()
```
jlkravitz, line 338: Should these be commented out?

backend/compact-connect/lambdas/python/purchases/tests/unit/test_purchase_client.py
```diff
@@ -52,6 +55,13 @@
 MOCK_SETTLEMENT_TIME_UTC_2 = '2024-12-26T15:15:20.007Z'
 MOCK_SETTLEMENT_TIME_LOCAL_2 = '2024-12-26T11:15:20.007Z'
 
+# mock order information
+MOCK_DATA_DESCRIPTOR = 'COMMON.ACCEPT.INAPP.PAYMENT'
```
jlkravitz, line 59: is this particular string meaningful or important in any way? If not, I'd be inclined to either add a comment like "// common descriptor"

or change to `mockDataDescriptor` or similar.

backend/compact-connect/lambdas/python/purchases/tests/function/test_handlers/test_transaction_history.py
```diff
@@ -287,6 +296,42 @@ def test_process_settled_transactions_stores_transactions_in_dynamodb(self, mock
             stored_transactions['Items'],
         )
 
+    @patch('handlers.transaction_history.PurchaseClient')
+    def test_process_settled_transactions_does_not_duplicate_identical_transaction_records(
+        self, mock_purchase_client_constructor
+    ):
+        """Test that transactions are stored in DynamoDB."""
```
jlkravitz, line 303: c/p comment?

backend/compact-connect/lambdas/python/purchases/tests/function/test_handlers/test_transaction_history.py
```diff
@@ -287,6 +296,42 @@ def test_process_settled_transactions_stores_transactions_in_dynamodb(self, mock
...
+        from handlers.transaction_history import process_settled_transactions
+
+        # in this test, there is one transaction, and one privilege. These should map together using the default
```
jlkravitz, line 306: Perhaps?

```suggestion
        # in this test, there is one transaction associated with the purchase of one privilege. These should map together using the default
```

jlkravitz (commented):
a few small comments/questions, but generally looks good!

webroot/src/components/PrivilegePurchaseFinalize/PrivilegePurchaseFinalize.ts
jsandoval81: This specific instance is a little different, in that the user is actually popping open an iframe hosted at `authorize.net`, which is more challenging to perform a direct input. So we've opted to simply copy the most onerous sample field to the OS context for the developers to easily use in the iframe (if they wish).

webroot/src/components/PrivilegePurchaseFinalize/PrivilegePurchaseFinalize.spec.ts
jsandoval81: @jlkravitz The hosted authorize.net iframe handles it's own input validation. We (correctly) no longer have any visibility or control over credit card details.

webroot/README.md
jsandoval81: @jlkravitz Correct, not required unless both of the following are true:

1. You are in a local developer environment
2. You wish to run the frontend against the mock API

This is just a way for developers to have this info locally, without having to commit the values to the repo.

webroot/src/components/PrivilegePurchaseAcceptUI/PrivilegePurchaseAcceptUI.ts
jsandoval81: @jlkravitz Good call. Added!

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/compact/record.py
landonshumway-ia: We keep them separate for security reasons. We want to minimize exposure to these creds as well as have stricter IAM policies for the lambda function that manages them. We don't want to have the compact admin have to specify or accidentially expose these credentials every time they need to make a configuration change.

webroot/src/models/CompactFeeConfig/CompactFeeConfig.model.ts
landonshumway-ia: @jsandoval81 In this case, I actually agree we should have the naming align. When `IsSandbox` is set to true, `isProductionMode` would be set to false, meaning you would have to inverse whatever response you get from the API right? If the value aligned with the backend without needing to inverse it then I would be fine naming it whatever, but if we need to have logic in place to change the values to align with a different name, that could get confusing.

backend/compact-connect/tests/smoke/compact_configuration_smoke_tests.py
landonshumway-ia: ah no, good catch, that was me running one specific test. I'll put those back in.

webroot/src/models/CompactFeeConfig/CompactFeeConfig.model.ts
jlkravitz: Yeah I think I lean towards changing. I don't necessarily think frontend/backend schemas need to be totally aligned, but it's helpful to at least use consistent terminology and conventions, even if the data schemas themselves are different.

jsandoval81: @jlkravitz Sounds good - updated.

webroot/src/components/PrivilegePurchaseAcceptUI/PrivilegePurchaseAcceptUI.ts
```diff
@@ -77,8 +77,8 @@ class PrivilegePurchaseAcceptUI extends Vue {
         return this.paymentSdkConfig?.clientKey || '';
     }
 
-    get isProductionMode(): boolean {
-        return this.paymentSdkConfig?.isProductionMode || false;
+    get isSandboxMode(): boolean {
+        return this.paymentSdkConfig?.isSandboxMode || false;
```
jlkravitz, line 81: Should this be `|| true` now?

webroot/src/models/CompactFeeConfig/CompactFeeConfig.model.ts
```diff
@@ -71,7 +71,7 @@ export class CompactFeeConfigSerializer {
             paymentSdkConfig: {
                 loginId: paymentProcessorPublicFields?.apiLoginId || '',
                 clientKey: paymentProcessorPublicFields?.publicClientKey || '',
-                isProductionMode: (typeof isSandbox === 'boolean') ? !isSandbox : false,
+                isSandboxMode: (typeof isSandbox === 'boolean') ? isSandbox : false,
```
jlkravitz, line 74: should the default be true now?

webroot/src/components/PrivilegePurchaseAcceptUI/PrivilegePurchaseAcceptUI.ts
jsandoval81: @jlkravitz As I was originally working on the initial POC, defaulting prod-mode to false seemed ok at that time. After the update today to rename things, I reconsidered that defaulting to prod-mode to true at this point was probably the right thing to do. With the reasoning that the prod environment would be the least impacted by any bugs or oddities in the future (e.g. Prod defaulting to sandbox-mode for some reason has more negative impact than Pre-Prod defaulting to prod-mode for some reason). This is actually something we usually do with other environment-type configs and I should have caught before opening the PR.

isabeleliassen (approved):

PR #861, opened 2025-06-16
Frontend/move military status

### LINKED ISSUE #838: "add military status" moved to account settings page
- add military status button currently shows at top of practitioner dashboard. The button should be removed from there and added to the account settings page

rmolinares:
### Requirements List
- _None_

### Description List
- Move military status button from dashboard to account screen matching latest [Figma designs](https://www.figma.com/design/SYM0uWszsu8Sf0YfxAhIMY/JCC?node-id=1-12&p=f&t=4aWemQs57euP96qd-0)
- Update button label
- Update back link on military status to return to account screen
- Add test to check the section exists on the account component

### Testing List
- Code review
- Testing
    - Log in as practitioner, go to the dashboard screen, and confirm that the military status button is no longer present anywhere
    - Go to the account screen and confirm that there is a military status section below the change password and overall appearance matches Figma designs above
        - Confirm that clicking the button will take you to the military status screen
            - Confirm that clicking the back link in the screen will redirect you back to the account screen 
        - Confirm that the military status flow works as expected and there are no redirects back to the dashboard 

Closes #838

### DIFF
--- a/webroot/src/components/UserAccount/UserAccount.ts
+++ b/webroot/src/components/UserAccount/UserAccount.ts
@@ -13,6 +13,7 @@ import {
 } from 'vue-facing-decorator';
 import { reactive, computed } from 'vue';
 import { AuthTypes } from '@/app.config';
+import InputButton from '@components/Forms/InputButton/InputButton.vue';
 import MixinForm from '@components/Forms/_mixins/form.mixin';
 import Card from '@components/Card/Card.vue';
 import InputText from '@components/Forms/InputText/InputText.vue';
@@ -28,6 +29,7 @@ import Joi from 'joi';
 @Component({
     name: 'UserAccount',
     components: {
+        InputButton,
         Card,
         InputText,
         InputSubmit,
@@ -86,6 +88,14 @@ class UserAccount extends mixins(MixinForm) {
         });
 
         return enabledInputKeys.length > 0;
+    }
+
+    get militaryStatusLabel(): string {
+        return `${this.$t('common.add')}/${this.$t('common.edit')}`;
+    }
+
+    get currentCompactType(): string | null {
+        return this.userStore?.currentCompact?.type || null;
     }
 
     //
@@ -165,6 +175,13 @@ class UserAccount extends mixins(MixinForm) {
                     this.setError(err.message);
                 });
         }
+    }
+
+    viewMilitaryStatus() {
+        this.$router.push({
+            name: 'MilitaryStatus',
+            params: { compact: this.currentCompactType }
+        });
     }
 
     //
--- a/webroot/src/components/UserAccount/UserAccount.vue
+++ b/webroot/src/components/UserAccount/UserAccount.vue
@@ -23,6 +23,18 @@
                 />
             </form>
             <ChangePassword />
+            <div class="military-status-container">
+                <div class="form-title">{{ $t('military.militaryStatusTitle') }}</div>
+                <div class="btn-container military-status-btn">
+                    <InputButton
+                        :label="militaryStatusLabel"
+                        :aria-label="militaryStatusLabel"
+                        :isTransparent="false"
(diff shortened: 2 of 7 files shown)


webroot/src/components/UserAccount/UserAccount.ts
```diff
@@ -88,6 +90,14 @@ class UserAccount extends mixins(MixinForm) {
         return enabledInputKeys.length > 0;
     }
 
+    get militaryStatusLabel(): string {
+        return `${this.$t('common.add')}/${this.$t('common.edit')}`;
+    }
```
jsandoval81, line 95: Some early components got quite cluttered by having the template labels in the TS modules. We've updated to just putting non-dynamic text directly in the templates. If it makes it easier, you can also add an item in the translations for this - but only if it helps.

webroot/src/components/UserAccount/UserAccount.ts
```diff
@@ -88,6 +90,14 @@ class UserAccount extends mixins(MixinForm) {
...
+
+    get currentCompactType(): string | null {
+        return this.userStore?.currentCompact?.type || null;
+    }
```
jsandoval81, line 99: We're trying to keep getters roughly in order of importance and / or consistent between components; rather than just adding them to the bottom as they come up. So this getter might go up closer to `get userStore()`.

webroot/src/components/UserAccount/UserAccount.ts
```diff
@@ -167,6 +177,16 @@ class UserAccount extends mixins(MixinForm) {
         }
     }
 
+    viewMilitaryStatus(): void {
+        const route = { name: 'MilitaryStatus' } as any;
```
jsandoval81, line 181: We're trying to avoid `any` types where possible.

webroot/src/components/UserAccount/UserAccount.ts
```diff
@@ -167,6 +177,16 @@ class UserAccount extends mixins(MixinForm) {
...
+
+        if (this.currentCompactType) {
+            route.params = { compact: this.currentCompactType };
+        }
+
+        this.$router.push(route);
```
jsandoval81, line 187: Would it be better to make sure `currentCompactType` exists, rather than doing `$router.push()` regardless?

I'm wondering if something like this would work:
```
viewMilitaryStatus(): void {
    if (this.currentCompactType) {
        this.$router.push({ name: 'MilitaryStatus', params: { compact: this.currentCompactType }});
    }
}
```

If you felt like there could ever be a case where the button might be clickable before `this.currentCompactType` is ready, you could additionally add a `v-if` to the template.

webroot/src/components/UserAccount/UserAccount.vue
```diff
@@ -23,6 +23,20 @@
...
             <ChangePassword />
+            <section class="military-status-container" aria-labelledby="military-status-title">
+                <h2 class="section-title" id="military-status-title">
+                    {{ $t('military.militaryStatusTitle') }}
+                </h2>
+                <div class="btn-container military-status-btn">
+                    <InputButton
+                        :label="militaryStatusLabel"
+                        :aria-label="militaryStatusLabel"
+                        :isTransparent="false"
```
jsandoval81, line 34: Can probably remove `:isTransparent="false"`. That's a pretty well-established default. If you feel strongly about leaving it in you can, we just typically haven't in this app.

webroot/src/components/UserAccount/UserAccount.vue
```diff
@@ -23,6 +23,20 @@
                 />
             </form>
             <ChangePassword />
+            <section class="military-status-container" aria-labelledby="military-status-title">
```
jsandoval81, line 26: I think we'd need some sort of `v-if` conditional here to only show this new section if the user is a licensee. This is likely as simple as using the existing `isLicensee` computed.

PR #870, opened 2025-06-19
Feat/deactivate privileges when home state license deactivated

landonshumway-ia:
When a home state license is marked as inactive by a jurisdiction (via `jurisdictionUploadedLicenseStatus` field) but is not expired according to the `dateOfExpiration` field, all privileges associated with that specific license type and jurisdiction should be automatically deactivated. This feature leverages our existing event-driven architecture to listen for license deactivation events and perform the privilege deactivation.

Closes #423

## Summary by CodeRabbit

* **New Features**
  * Added automated deactivation of privileges when a provider's home state license is deactivated by the jurisdiction.
  * Introduced event-driven processing for license deactivation, including new event listener infrastructure and event schemas.
  * Privilege records now display a deactivation status when affected by license deactivation.
  * Enhanced privilege purchase validation to allow renewal of privileges previously deactivated due to license deactivation.
  * Added detailed smoke tests verifying end-to-end license deactivation privilege workflow.

* **Bug Fixes**
  * Prevented privilege deactivation events from being published for licenses that are already expired.

* **Tests**
  * Added comprehensive unit and smoke tests for license deactivation workflows and event processing.
  * Extended test coverage for privilege renewal after deactivation.

* **Chores**
  * Updated jurisdiction mappings to include U.S. Territories.
  * Improved documentation and type safety in test utilities.

PR #874, opened 2025-06-20
Add create date and effective date to privilege events

### LINKED ISSUE #887: Add createDate and effectiveDate to privilege history db events
### Tasks
- [ ] TODO
- [ ] Automated tests
- [ ] API Docs
- [ ] Postman collection

ChiefStief:
### Requirements List
-

### Description List
- Added createDate and effectiveDate to all privilege updates and modified events and tests as needed
- Added createDate and effectiveDate migration script and infra

### Testing List
- Test in your sandbox
- Code review

### Notes
These changes were initially going to be part of the synthetic history ticket, 770. Having these values in the db is required to be able to generate a timeline of all the events.

When we halted progress on that ticket, I pulled these standalone changes out with the goals of:

1) Getting some immediate value from the work done here
2) Merging them sooner than that ticket so that we have the dbs populated as correctly as possible as soon as possible because we cant always correctly back populate this information

Closes #887

## Summary by CodeRabbit

* **New Features**
  * Added explicit `createDate` and `effectiveDate` fields to privilege update records and related API responses for improved timestamp tracking.
  * Introduced a data migration to backfill these new date fields for existing privilege update records.

* **Bug Fixes**
  * Updated privilege and license update operations to consistently include creation and effective dates.

* **Tests**
  * Enhanced test data and assertions to validate presence of new date fields in privilege update records.

* **Chores**
  * Improved import formatting for better code readability.


landonshumway-ia (commented):
From what I can tell, this implementation looks solid. Nice work. I think we will want to hold off on merging this though until we get #870 in, since we will also need to add these new fields to those events as well.

Also to verify, it seems the intent is to only include these new fields specifically for privilege updates, and not licenses right? I have no issue with this for now, since we do not display license history anywhere. However, it is very likely that this will become a need in the future, and if we don't collect that information until after the system is live, we will not be able to just 'guess' at what the effective date for the old license update records was, since that could have legal implications when messing with live data like that. We will probably have to allow null for those fields on license records when we get to that point, to denote that we simply don't know the effective date for those particular records.

backend/compact-connect/lambdas/python/common/cc_common/event_bus_client.py
```diff
@@ -120,6 +120,7 @@ def publish_license_encumbrance_event(
         provider_id: str,
         jurisdiction: str,
         license_type_abbreviation: str,
+        effective_date: str,
```
jusdino, line 123: This should be a `datetime`, not a `str`. We generally should aim to serialize objects as late as possible, so we're passing the richer objects around internally.

backend/compact-connect/lambdas/python/common/cc_common/event_bus_client.py
```diff
@@ -156,12 +160,14 @@ def publish_license_encumbrance_lifting_event(
...
         :param jurisdiction: The jurisdiction of the license
         :param license_type_abbreviation: The license type abbreviation
+        :param effective_date: The effective lift date of the encumbrance
         """
         event_detail = {
             'compact': compact,
             'providerId': provider_id,
             'jurisdiction': jurisdiction,
             'licenseTypeAbbreviation': license_type_abbreviation,
+            'effectiveDate': effective_date,
```
jusdino, line 170: It's funny - I think I made this same change to the event interfaces as part of my [AA-notification PR](https://github.com/csg-org/CompactConnect/pull/864). We'll have to watch out for merge conflicts / duplication here and to be sure we approach it consistently across the interface.

backend/compact-connect/lambdas/python/migration/create_effective_date_7/main.py
```diff
@@ -0,0 +1,154 @@
...
+    def on_update(self, properties: dict) -> None:
+        do_migration(properties)
+
+    def on_delete(self, _properties: dict) -> CustomResourceResponse | None:
+        """
+        No roll-back on delete.
+        """
+
+
+on_event = CreateEffectiveDateMigration('create-effective-date-1')
```
jusdino, line 23: `-1`?

backend/compact-connect/stacks/persistent_stack/__init__.py
```diff
@@ -360,6 +360,29 @@ def _add_migrations(self):
             ],
         )
 
+        update_dates_migration = DataMigration(
+            self,
+            '7CreateEffectiveDate',
```
jusdino, line 365: What are the `7`'s for? We were tagging these with the issue number.

jusdino (commented):
I think I'm missing some context around this change - is there an associated ticket? I'm guessing that this has to do with the privilege history work you were doing. A little explanation of your full intent here in the PR description would be helpfu.

backend/compact-connect/lambdas/python/migration/create_effective_date_7/main.py
ChiefStief: I did this to line up with the what I thought was happening with the numbers explained below. But apparently forgot to update this one and it doesn't matter generally.

Ill make a ticket for this work generally and change this to just be the ticket number everywhere

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
```diff
@@ -2651,6 +2707,8 @@ def deactivate_license_privileges(
                     'jurisdiction': privilege_data.jurisdiction,
                     'licenseType': privilege_data.licenseType,
                     'previous': privilege_data.to_dict(),
+                    'effectiveDate': now,
+                    'createDate': now,
                     'updatedValues': {
                         'licenseDeactivatedStatus': LicenseDeactivatedStatusEnum.LICENSE_DEACTIVATED,
                     },
```
landonshumway-ia, line 2714: Any chance you can also add these same fields for license update records when they are deactivated? If not that's fine, but we will eventually need it for the frontend effort to determine the last known date for which a license has been active.

landonshumway-ia (approved):
Looks good, nice work 👍  a couple of nits but they will not block my approval should you choose to pass on them.

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
ChiefStief: I think we should add this as a detail to the end point (931), I scoped this ticket to just privilege updates to specifically support the privilege history ticket

backend/compact-connect/lambdas/python/common/common_test/test_data_generator.py
```diff
@@ -330,6 +330,8 @@ def generate_default_privilege_update(
             'jurisdiction': DEFAULT_PRIVILEGE_JURISDICTION,
             'licenseType': DEFAULT_LICENSE_TYPE,
             'previous': previous_privilege.to_dict(),
+            'createDate': datetime.fromisoformat(DEFAULT_PRIVILEGE_UPDATE_DATE_OF_UPDATE),
+            'effectiveDate': datetime.fromisoformat(DEFAULT_PRIVILEGE_UPDATE_DATE_OF_UPDATE),
```
jlkravitz, line 334: should this use the constant you newly defined, `DEFAULT_LIFT_EFFECTIVE_DATE`?

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
```diff
@@ -1350,6 +1378,8 @@ def encumber_privilege(self, adverse_action: AdverseActionData) -> None:
                     'compact': adverse_action.compact,
                     'jurisdiction': adverse_action.jurisdiction,
                     'licenseType': privilege_data.licenseType,
+                    'createDate': now,
+                    'effectiveDate': adverse_action.effectiveStartDate,
```
jlkravitz, line 1382: Without documentation, my assumption is that `createDate`  indicates when the privilege update record was created? How does that differ from `dateOfUpdate`? (Presumably, the record is created when the privilege was updated.)

And then, I'm a bit confused about `effectiveDate`. Is it only relevant for encumbrances?

jlkravitz (commented):
couple small notes!

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
ChiefStief: `dateOfUpdate` is an automatically generated field whenever the privilegeUpdate record is updated. At the moment I think this ends up being the same as `createDate` because they are not updated outside of creation, but they could be theoretically. I was told to not rely on `dateOfUpdate` being identical to `createDate` and we need the `createDate` so Im adding the field.

`effectiveDate` is when the update is effective on the provider which is not the same conceptually as the `createDate`. As you said, at the moment the only time these differ is in encumbrance and encumbrance lifted events because they can be backdated in `effectiveDate` when uploaded to the system. Storing separate fields is important for these events because it allows us to sort by and display `effectiveDate` which is what the average user cares about while also maintaining records of when these entered the system which will be helpful for general audit purposes, potential legal issues, and creating a chronological timeline of events entering the system which will be important when working on the history timeline to disentangle edge cases caused by the `effectiveDate` backdating of license encumbrance events.

jlkravitz: Aha, got it. Thanks for the explanation.

jlkravitz (approved):
@isabeleliassen good to merge

@ChiefStief To avoid missing anything, I'll let you resolve the CodeRabbit comments that remain. If you make changes because of any of them, let me know and I'll reapprove. Thanks!

Issue #863, opened 2025-06-17
Stabilize test coverage for branches FE


jsandoval81:
Frontend branches coverage has taken a small dip recently, causing a little extra time as each PR is now hovering near our coverage goal. 

Statements / Functions / Lines coverage are in solid shape, this is just for Branches coverage.

rmolinares:
### QA Notes
- See https://github.com/csg-org/CompactConnect/issues/855#issuecomment-3013135797 as this ticket was as part of that effort

PR #878, opened 2025-06-23
Frontend/form submission validation visibility

### LINKED ISSUE #855: gray out next or submit buttons until all fields are filled
- right now some submit/next buttons appear clickable even when not all required fields are filled. They should remain gray until all required fields are filled
- If possible, if someone tries to click the button when it is gray, they should get a message saying that not all the required fields have been filled.
- add message if missing
- for both registration and privilege purchase flow

rmolinares:
### Requirements List
- _None_

### Description List
- Update practitioner registration and obtain privileges form to include a feature that scrolls and focuses the user to the first invalid required input in the form (from top to bottom)
    - Undoes the prior flow which includes disabled continue buttons until all required fields are valid 
- Improve unrelated test coverage
- Fix mobile bug where nav would scroll behind the fixed next / previous buttons in the obtain privileges flow
- Update payment summary screen title

### Testing List
- yarn test:unit:coverage` confirm the branches line item has an updated acceptable percentage
- Code review
- Testing
    - Proceed to the practitioner registration form and confirm that the continue button is no longer disabled and will auto position screen to the first invalid required field in the form, as well as direct focus on that input element, with the exception of radio buttons which will focus the entire input group
    - Confirm that same experience in the obtain privileges form flow with a slight exception to the select privileges screen which disables the continue button initially until at least 1 state is selected

Closes #855 #863

### DIFF
--- a/webroot/src/components/Forms/InputDate/InputDate.vue
+++ b/webroot/src/components/Forms/InputDate/InputDate.vue
@@ -17,7 +17,7 @@
     >
         <label
             v-if="!formInput.shouldHideLabel"
-            :for="`dp-input-${formInput.id}`"
+            :for="formInput.id"
         >
             {{ formInput.label }}
             <span v-if="isRequired" class="required-indicator">*</span>
@@ -63,7 +63,7 @@
             >
                 <div class="dp__input_wrap">
                     <input
-                        :id="`dp-input-${formInput.id}`"
+                        :id="formInput.id"
                         type="text"
                         v-model="localValue"
                         @input="onInput"
--- a/webroot/src/components/Forms/InputRadioGroup/InputRadioGroup.vue
+++ b/webroot/src/components/Forms/InputRadioGroup/InputRadioGroup.vue
@@ -12,6 +12,7 @@
             'has-error': !!formInput.errorMessage,
             'disabled': formInput.isDisabled,
         }"
+        :id="formInput.id"
     >
         <div
             v-if="!formInput.shouldHideLabel"
--- a/webroot/src/components/Forms/_mixins/form.mixin.ts
+++ b/webroot/src/components/Forms/_mixins/form.mixin.ts
@@ -211,6 +211,67 @@ class MixinForm extends Vue {
         this.updateFormSubmitError(errorMessage);
     }
 
+    showInvalidFormError(): void {
+        const { formData } = this;
+
+        // Find the first required invalid input
+        const firstInvalidKey = this.formKeys.find((key) => {
+            const formInput = formData[key];
+
+            // Skip submit inputs
+            if (formInput.isSubmitInput) {
+                return false;
+            }
+
+            const isInvalid = !formInput.isValid;
+
+            return isInvalid;
+        });
+
+        // If we found an invalid input, scroll to it
+        if (firstInvalidKey) {
+            this.scrollToInput(formData[firstInvalidKey]);
+        }
+    }
+
+    protected scrollToInput(formInput: FormInput): void {
+        const element = document.getElementById(formInput.id);
+
+        if (element) {
+            // Scroll to the element
+            element.scrollIntoView({
+                behavior: 'smooth',
+                block: 'center'
+            });
+
+            // Check if this is a radio button group by looking for radio inputs inside
+            const radioInputs = element.querySelectorAll('input[type="radio"]');
+
+            console.log('radioInputs', radioInputs);
+
+            if (radioInputs.length > 0) {
+                // This is a radio button group, focus on the first radio input
+                const firstRadioInput = document.getElementById(`${formInput.id}-1`);
+
+                if (firstRadioInput) {
+                    console.log('firstRadioInput', firstRadioInput);
+
+                    firstRadioInput.focus();
+                } else {
+                    console.log('fallback to first radio input');
+
+                    // Fallback to the first radio input found
+                    (radioInputs[0] as HTMLElement).focus();
+                }
+            } else {
+                console.log('non-radio button group, focus on the input');
+
+                // Non-radio button group, focus on the input
+                element.focus();
+            }
+        }
+    }
+
     @Watch('locale') localeChanged() {
         // @TODO: For now we just brute force the form re-init on language change. Making all the layers reactive (messages inside Joi schemas, etc) was messy.
         this.initFormInputs();
--- a/webroot/src/components/Forms/_mixins/form.mixin.ts
+++ b/webroot/src/components/Forms/_mixins/form.mixin.ts
@@ -247,26 +247,17 @@ class MixinForm extends Vue {
             // Check if this is a radio button group by looking for radio inputs inside
             const radioInputs = element.querySelectorAll('input[type="radio"]');
 
-            console.log('radioInputs', radioInputs);
-
             if (radioInputs.length > 0) {
                 // This is a radio button group, focus on the first radio input
                 const firstRadioInput = document.getElementById(`${formInput.id}-1`);
 
                 if (firstRadioInput) {
-                    console.log('firstRadioInput', firstRadioInput);
-
                     firstRadioInput.focus();
                 } else {
-                    console.log('fallback to first radio input');
-
                     // Fallback to the first radio input found
                     (radioInputs[0] as HTMLElement).focus();
                 }
             } else {
-                console.log('non-radio button group, focus on the input');
-
-                // Non-radio button group, focus on the input
                 element.focus();
             }
         }
--- a/webroot/src/components/Forms/_mixins/mixins.spec.ts
+++ b/webroot/src/components/Forms/_mixins/mixins.spec.ts
@@ -124,6 +124,185 @@ describe('Form mixin', async () => {
 
         expect(component.isFormError).to.equal(true);
     });
+    it('should show invalid form error when there is an invalid input', async () => {
+        const wrapper = await mountShallow(FormMixin);
+        const component = wrapper.vm;
+        const formInput = new FormInput();
+
+        formInput.id = 'test-input';
+        formInput.isValid = false;
+        formInput.isSubmitInput = false;
+        component.formData.testInput = formInput;
+
+        // Mock scrollToInput method to verify it gets called
+        let scrollToInputCalled = false;
+
+        component.scrollToInput = () => {
+            scrollToInputCalled = true;
+        };
+
+        component.showInvalidFormError();
+
+        expect(scrollToInputCalled).to.equal(true);
+    });
+    it('should not call scrollToInput when all inputs are valid', async () => {
+        const wrapper = await mountShallow(FormMixin);
+        const component = wrapper.vm;
+        const formInput = new FormInput();
+
+        formInput.id = 'test-input';
+        formInput.isValid = true;
+        formInput.isSubmitInput = false;
+        component.formData.testInput = formInput;
+
+        // Mock scrollToInput method to verify it doesn't get called
+        let scrollToInputCalled = false;
+
+        component.scrollToInput = () => {
+            scrollToInputCalled = true;
+        };
+
+        component.showInvalidFormError();
+
+        expect(scrollToInputCalled).to.equal(false);
+    });
+    it('should skip submit inputs when finding invalid form inputs', async () => {
+        const wrapper = await mountShallow(FormMixin);
+        const component = wrapper.vm;
+        const submitInput = new FormInput();
+        const regularInput = new FormInput();
+
+        submitInput.id = 'submit-input';
+        submitInput.isValid = false;
+        submitInput.isSubmitInput = true;
+
+        regularInput.id = 'regular-input';
+        regularInput.isValid = false;
+        regularInput.isSubmitInput = false;
+
+        component.formData.submitInput = submitInput;
+        component.formData.regularInput = regularInput;
+
+        // Mock scrollToInput method to verify it gets called with the regular input
+        let scrollToInputCalledWith = null;
+
+        component.scrollToInput = (input) => {
+            scrollToInputCalledWith = input;
+        };
+
+        component.showInvalidFormError();
+
+        expect(scrollToInputCalledWith).to.not.be.null;
+        expect((scrollToInputCalledWith as any).id).to.equal(regularInput.id);
+    });
+    it('should scroll to input element when element exists', async () => {
+        const wrapper = await mountShallow(FormMixin);
+        const component = wrapper.vm;
+        const formInput = new FormInput();
+
+        formInput.id = 'test-input';
+
+        // Mock DOM element and its methods
+        const mockElement = {
+            scrollIntoView: () => { /* mock implementation */ },
+            querySelectorAll: () => [],
+            focus: () => { /* mock implementation */ }
+        } as unknown as HTMLElement;
+
+        // Mock document.getElementById
+        const originalGetElementById = document.getElementById;
+
+        document.getElementById = (id: string): HTMLElement | null => {
+            if (id === 'test-input') {
+                return mockElement;
+            }
+            return null;
+        };
+
+        let scrollIntoViewCalled = false;
+        let focusCalled = false;
+
+        (mockElement as any).scrollIntoView = () => {
+            scrollIntoViewCalled = true;
+        };
+        (mockElement as any).focus = () => {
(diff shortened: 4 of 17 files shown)


webroot/src/styles.common/mixins/purchase-flow-buttons.less
```diff
@@ -34,12 +34,32 @@
...
+                z-index: 1;
+                justify-content: start;
+                width: 100%;
+                height: 100%;
+
+                :deep(.input-button) {
+                    width: 100%;
+                }
+            }
+        }
```
jsandoval81, line 61: I feel like this fairly specific new style should just go in `PrivilegePurchaseFinalize.less` for now.

jsandoval81 (changes requested):

webroot/src/components/Forms/_mixins/form.mixin.ts
```diff
@@ -211,6 +211,58 @@ class MixinForm extends Vue {
...
+        });
+
+        // If we found an invalid input, scroll to it
+        if (firstInvalidKey) {
+            this.scrollToInput(formData[firstInvalidKey]);
+        }
+    }
+
+    protected scrollToInput(formInput: FormInput): void {
+        const element = document.getElementById(formInput.id);
```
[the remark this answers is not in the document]
rmolinares: `getElementsByName` works pretty well actually!

webroot/src/components/Forms/InputDate/InputDate.vue
```diff
@@ -63,7 +63,7 @@
             >
                 <div class="dp__input_wrap">
                     <input
-                        :id="`dp-input-${formInput.id}`"
+                        :id="formInput.id"
```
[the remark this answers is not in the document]
rmolinares: Yeah, I'll set it back.

webroot/src/components/Forms/_mixins/form.mixin.ts
```diff
@@ -211,6 +211,58 @@ class MixinForm extends Vue {
...
+            return isInvalid;
+        });
+
+        // If we found an invalid input, scroll to it
+        if (firstInvalidKey) {
+            this.scrollToInput(formData[firstInvalidKey]);
+        }
+    }
+
+    protected scrollToInput(formInput: FormInput): void {
```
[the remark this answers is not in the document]
rmolinares: Yeah, this works really well. Since you're more familiar with the app, can you think of any forms that wouldn't want the `validateAll` config to include `skipErrorScroll = false`. I can't think of any initially.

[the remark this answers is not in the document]
jsandoval81: None come to mind. Just give a smoke test around the app to make sure there isn't anything obvious.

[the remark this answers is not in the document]
rmolinares: After checking it looks like they all could benefit from this feature. I made an adjustment to a few more to remove the disabled submit button to match the new form behavior.

webroot/src/components/Forms/InputRadioGroup/InputRadioGroup.vue
```diff
@@ -12,6 +12,7 @@
             'has-error': !!formInput.errorMessage,
             'disabled': formInput.isDisabled,
         }"
+        :id="formInput.id"
```
jsandoval81, line 15: Minor: We generally put `id` above `class`.

webroot/src/components/Forms/_mixins/form.mixin.ts
```diff
@@ -158,20 +158,29 @@ class MixinForm extends Vue {
...
         this.formKeys.forEach((key) => {
-            if (config.asTouched) {
-                formData[key].isTouched = true;
+            // Only validate if it's a FormInput (has a validate function)
+            if (formData[key] && typeof formData[key].validate === 'function') {
+                if (config.asTouched) {
+                    formData[key].isTouched = true;
+                }
+                formData[key].validate();
             }
```
jsandoval81, line 167: Is it possible to just more straightforwardly check the type - something like `instanceof FormInput` or similar?

If not, prefer to use optional chaining to reduce the statement checks; e.g.:

`if (typeof formData[key]?.validate === 'function') {`

webroot/src/components/Forms/_mixins/form.mixin.ts
```diff
@@ -158,20 +158,29 @@ class MixinForm extends Vue {
...
+            this.showInvalidFormError();
+        }
     }
 
     checkValidForAll(): void {
         const { formData } = this;
 
-        this.isFormValid = this.formKeys.every((key) => formData[key].isValid);
+        this.isFormValid = this.formKeys
+            .filter((key) => formData[key] && typeof formData[key].validate === 'function')
```
jsandoval81, line 182: Same as above. Check by type, or reduce using optional chaining operator.

webroot/src/components/Forms/_mixins/form.mixin.ts
```diff
@@ -211,6 +220,47 @@ class MixinForm extends Vue {
         this.updateFormSubmitError(errorMessage);
     }
 
+    showInvalidFormError(): void {
+        const { formData } = this;
+
+        // Find the first required invalid input that has a validation schema
```
jsandoval81, line 226: I think we can remove the word "required" here. I think we're just finding the first invalid input.

webroot/src/components/Forms/_mixins/form.mixin.ts
```diff
@@ -211,6 +220,47 @@ class MixinForm extends Vue {
...
+        const firstInvalidKey = this.formKeys.find((key) => {
+            const formInput = formData[key];
+
+            return (
+                !formInput.isSubmitInput
+                && !formInput.isValid
+                && typeof formInput.validate === 'function'
+                && formInput.validation // Only consider inputs with validation
```
jsandoval81, line 234: Do we need these 2 additional checks here? Would the first 2 be sufficient?

webroot/src/components/Forms/_mixins/form.mixin.ts
```diff
@@ -211,6 +220,47 @@ class MixinForm extends Vue {
...
+            );
+        });
+
+        // If we found an invalid input, try to scroll to and focus its input by name
+        if (firstInvalidKey) {
+            const formInput = formData[firstInvalidKey];
+
+            this.scrollToInput(formInput);
```
jsandoval81, line 242: This is probably an acceptable place for a one-liner.

webroot/src/components/Forms/_mixins/form.mixin.ts
```diff
@@ -211,6 +220,47 @@ class MixinForm extends Vue {
...
+        }
+    }
+
+    protected scrollToInput(formInput: FormInput): void {
+        const elements = document.getElementsByName(formInput.name);
+
+        if (elements.length > 0) {
+            const element = elements[0] as HTMLElement;
+
+            if (element) {
```
jsandoval81, line 252: These 2 `if` checks feel like they overlap in a way that could be cleaned up. Maybe something like:

```
const elements = document.getElementsByName(formInput.name);

if (elements[0]) {
    const element = elements[0] as HTMLElement;

    // Scroll to the element
    ...
```

Or:

```
const element = document.getElementsByName(formInput.name)[0] as HTMLElement | undefined;

if (element) {
    // Scroll to the element
    ...
```

jsandoval81 (changes requested):
Nice work - the form setup on that page feels a _LOT_ better 🚀

webroot/src/pages/RegisterLicensee/RegisterLicensee.ts
```diff
@@ -308,6 +308,8 @@ class RegisterLicensee extends mixins(MixinForm) {
             await nextTick();
 
             this.scrollIntoView('summary-heading');
+        } else {
+            this.showInvalidFormError();
```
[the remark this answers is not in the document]
rmolinares: It's not. For some reason this update didn't save. Good catch.

webroot/src/components/SelectedStatePurchaseInformation/SelectedStatePurchaseInformation.ts
```diff
@@ -312,12 +312,22 @@ class SelectedStatePurchaseInformation extends mixins(MixinForm) {
...
     }
 
     setScopeInputValue(newValue): void {
         if (this.scopeOfPracticeCheckInput) {
             (this.scopeOfPracticeCheckInput.value as any) = newValue; // any use required here because of outstanding ts bug regarding union type inference
+
+            // Clear error message when checkbox becomes valid
+            if (newValue) {
+                this.scopeOfPracticeCheckInput.errorMessage = '';
+            }
```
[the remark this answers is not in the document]
rmolinares: We could, and I removed it now, but it still needs a `this.scopeOfPracticeCheckInput.validate();` line because this is a programmatic value change (triggered from the modal), so it needs to call .validate() explicitly.

The only other way would be to add a custom watcher but that could also have potential side effects. Still better and more in-line than what I had originally.

webroot/src/components/Forms/InputRadioGroup/InputRadioGroup.vue
```diff
@@ -8,6 +8,7 @@
 <template>
     <div
         class="input-container radio-group-container"
+        :id="formInput.id"
```
jsandoval81, line 11: Just one more line up (above `class` to be consistent with the rest of the app) 🙂

webroot/src/components/Forms/_mixins/form.mixin.ts
```diff
@@ -158,20 +158,29 @@ class MixinForm extends Vue {
         const { formData } = this;
 
         this.formKeys.forEach((key) => {
-            if (config.asTouched) {
-                formData[key].isTouched = true;
+            // Only validate if it's a FormInput
+            if (formData[key] instanceof FormInput) {
```
jsandoval81, line 162: Since we make this evaluation ~3 times now, I'm thinking it would be best to just create a computed - underneath `formValues` and above `formSubmitInputs`. That way you can just reference a new computed like `this.formInputs`; which will also clean up a lot of the existing `formData[key]` verbosity. Thoughts?

webroot/src/components/Forms/_mixins/mixins.spec.ts
```diff
@@ -124,6 +126,102 @@ describe('Form mixin', async () => {
...
+        formInput.name = 'test-input';
+        formInput.isValid = false;
+        formInput.isSubmitInput = false;
+        formInput.validation = Joi.string().required();
+        component.formData.testInput = formInput;
+
+        const spy = sinon.spy();
+
+        component.scrollToInput = spy;
+        component.showInvalidFormError();
```
jsandoval81, line 143: Minor: Let's move the variable declaration up with the rest of the declarations & also remove any of these blank lines we don't need. I'm good if you want to leave a blank line between the setup lines and the `component.showInvalidFormError();` line if you like separation between the setup & the execution.

This would apply to most of the new test blocks in this file.

webroot/src/components/SelectedStatePurchaseInformation/SelectedStatePurchaseInformation.ts
```diff
@@ -312,12 +312,16 @@ class SelectedStatePurchaseInformation extends mixins(MixinForm) {
...
+            this.jurisprudenceCheckInput.validate();
         }
     }
 
     setScopeInputValue(newValue): void {
         if (this.scopeOfPracticeCheckInput) {
-            (this.scopeOfPracticeCheckInput.value as any) = newValue; // any use required here because of outstanding ts bug regarding union type inference
+            (this.scopeOfPracticeCheckInput.value as any) = newValue;
+
+            this.scopeOfPracticeCheckInput.validate();
```
jsandoval81, line 324: The comment explaining `any` should also be kept in `setScopeInputValue()`. We can also get rid of the empty lines in these.

jsandoval81 (changes requested):

webroot/src/components/Forms/_mixins/form.mixin.ts
jsandoval81: Got it. Ideally we don't have non-FormInputs in `formData`, but I also don't want to 100% rule them out for cases where we need flexibility. I think the `stateChecklist` could also be cleaned up to be top-level dynamic FormInputs too, but we can do that another time. We can resolve this note.

webroot/src/components/Forms/_mixins/form.mixin.ts
rmolinares: Yeah, good suggestion. I've added that.

jsandoval81 (approved):

jsandoval81:

webroot/src/components/Forms/_mixins/form.mixin.ts
jlkravitz: Bumping this comment- Per John's comment, I think this would be more correct?
```suggestion
        // Find the first invalid input that has a validation schema
```

webroot/src/components/Forms/_mixins/form.mixin.ts
```diff
@@ -42,6 +42,12 @@ class MixinForm extends Vue {
         return values;
     }
 
+    get formInputs(): Array<FormInput> {
+        return this.formKeys
+            .map((key) => this.formData[key])
+            .filter((input) => input instanceof FormInput);
```
jlkravitz, line 48: What are we filtering out here? (i.e., what else is in `formKeys` other than FormInputs? Or is this just a TS thing to clarify the type?

webroot/src/components/PrivilegePurchaseFinalize/PrivilegePurchaseFinalize.ts
```diff
@@ -265,7 +265,7 @@ export default class PrivilegePurchaseFinalize extends mixins(MixinForm) {
                 id: 'no-refunds-check',
                 name: 'no-refunds-check',
                 label: this.$t('licensing.noRefundsMessage'),
-                validation: Joi.boolean().required(),
+                validation: Joi.boolean().invalid(false).messages(this.joiMessages.boolean),
```
jlkravitz, line 268: should this still be 'required' to avoid `undefined` being a valid value?

webroot/src/components/PrivilegePurchaseSelect/PrivilegePurchaseSelect.ts
```diff
@@ -346,8 +345,61 @@ export default class PrivilegePurchaseSelect extends mixins(MixinForm) {
...
+                name: `jurisprudence-${stateAbbrev}`,
+                label: this.jurisprudenceExplanationText,
+                value: false,
+                validation: Joi.boolean().invalid(false).required().messages(this.joiMessages.boolean),
+            });
+        } else {
+            delete this.formData[`jurisprudence-${stateAbbrev}`];
+        }
+
+        // Always add scope attestation (or add a check if that's conditional)
```
jlkravitz, line 368: What's '(or add a check if that's conditional)' referring to and/or mean?

jlkravitz (commented):
couple small things!

webroot/src/components/Forms/_mixins/form.mixin.ts
rmolinares: In certain cases there could be non-`FormInput` data in `formData`. For example, `PrivilegePurchaseSelect` has a `stateCheckList` array of `FormInput` types. The array itself was causing runtime issues since it isn't of `FormInput` type. There may be other cases such as this one where non-`FormInputs` are lumped into the `formData` and we want to provide that flexibility so we filter those out here.

webroot/src/components/PrivilegePurchaseFinalize/PrivilegePurchaseFinalize.ts
rmolinares: Good catch. It was removed and never re-introduced by accident. Added now.

webroot/src/components/Forms/_mixins/form.mixin.ts
rmolinares: Correct. Word has now been removed.

jlkravitz (approved):

isabeleliassen (approved):

PR #885, opened 2025-06-25
Frontend/storage switch and redirect preserve

### LINKED ISSUE #866: Allow multi-tab localstorage sessions
It was requested on a recent standup to allow frontend auth tokens to persist across tabs (`localStorage`). The context was so that new notification emails can take authenticated users straight to a desired page in the app (#675).

Currently we are using single-tab sessions (`sessionStorage`). 

- **`sessionStorage`**
    - Pros:
        - Security; tokens are cleared automatically on tab close (or overwritten on successful re-login)
    - Cons:
        - UX; each new tab requires separate authentication
- **`localStorage`**
    - Pros:
        - UX; authentication persists across tabs without having to re-login on each tab
    - Cons:
        - Security; tokens are persisted indefinitely. This is largely mitigated by the auth server setting a short expiry on the auth tokens

---
[Optional]
As part of this, we should also evaluate the effort to attempt to honor the original URI if the user is not already authenticated and is forced to complete the login flow when first visiting an authenticated URI. If the effort is small, we should include this. If not, it can be a separate ticket.
- router.beforeEach would need to be updated to add goto query to Logout route call
- Logout page would need to add logoutChecklist to logged out case
- Mock login would need to be updated to check for a storage redirect
- Would need some way to only route for authType, extra storage value or something
- Make sure all storage values are cleared after login

jsandoval81:
### Requirements List
- _None_

### Description List
- Changed the storage for auth from `sessionStorage` to `localStorage`
- Added the ability to preserve the requested path through the auth workflow
    - e.g. If users get logged out by the router nav guard, the app will attempt to route them to that page after the next time they (re)authenticate

### Testing List
- Code review
- Testing
    - After authenticating, additional browser tabs can be opened to the app without having to re-authenticate
    - If attempting to access an authenticated page while logged out, that page should be the landing page after logging in. Otherwise, logging in should redirect to the same default pages as before.
        - Example 1 (Staff)
            1. Login as a Staff user
            1. Search licensing data and navigate to the detail page of a licensee
            1. Copy the URL of the staff-authenticated licensee detail page
            1. Logout
            1. Attempt to visit the copied URL (the staff-authenticated licensee detail page)
            1. Since you are not logged in, you should be taken to the public dashboard page
            1. Login as the same staff user
            1. After successful login, you should be automatically redirected to the licensee detail page you tried to request while logged out
            1. Logout
            1. Login as the same Staff user
            1. After successful login, you should be automatically redirected to the default staff home page - licensee search page - because you didn't request any specific pages while logged out (as expected)
        - Example 2 (Licensee)
            1. Keep the same staff-authenticated licensee detail page URL copied from Example 1 (Staff) above
            1. While logged out, attempt to visit the copied URL (the staff-authenticated licensee detail page)
            1. Since you are not logged in, you should be taken to the public dashboard page
            1. Login as a _Licensee_ user
            1. After successful login, you should be automatically redirected to the default licensee home page - the licensee dashboard - because the staff URL you attempted while logged out is not a valid licensee page
            1. Visit the Generate-verification-doc page as a licensee
            1. Copy the URL of the licensee-authenticate verification-doc page
            1. Logout
            1. Attempt to visit the copied URL (the licensee-authenticate verification-doc page)
            1. Since you are not logged in, you should be taken to the public dashboard page
            1. Login as the same licensee user
            1. After successful login, you should be automatically redirected to the licensee detail page you tried to request while logged out
            1. Logout
            1. Login as the same Licensee user
            1. After successful login, you should be automatically redirected to the default licensee home page - licensee dashboard - because you didn't request any specific pages while logged out (as expected)

Closes #866 

## Summary by CodeRabbit

* **New Features**
  * Improved user redirection after authentication and logout by preserving both the intended destination and authentication type.
  * Users are now redirected back to their original path after login if the authentication type matches.

* **Bug Fixes**
  * Ensured the logout checklist runs consistently, even when the user is not logged in.

* **Enhancements**
  * Enhanced navigation logic to better handle redirect paths and authentication types across login, logout, and mock login scenarios.

### DIFF
--- a/webroot/src/pages/PublicDashboard/PublicDashboard.ts
+++ b/webroot/src/pages/PublicDashboard/PublicDashboard.ts
@@ -6,7 +6,12 @@
 //
 
 import { Component, Vue } from 'vue-facing-decorator';
-import { AuthTypes } from '@/app.config';
+import {
+    authStorage,
+    AuthTypes,
+    AUTH_LOGIN_GOTO_PATH,
+    AUTH_LOGIN_GOTO_PATH_AUTH_TYPE
+} from '@/app.config';
 import SearchIcon from '@components/Icons/Search/Search.vue';
 import RegisterIcon from '@components/Icons/RegisterAlt/RegisterAlt.vue';
 import StaffUserIcon from '@components/Icons/StaffUser/StaffUser.vue';
@@ -81,6 +86,8 @@ export default class DashboardPublic extends Vue {
     }
 
     async mockStaffLogin(): Promise<void> {
+        const goto = authStorage.getItem(AUTH_LOGIN_GOTO_PATH);
+        const gotoAuthType = authStorage.getItem(AUTH_LOGIN_GOTO_PATH_AUTH_TYPE);
         const data = {
             access_token: 'mock_access_token',
             token_type: 'Bearer',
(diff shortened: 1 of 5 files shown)


jsandoval81:
@jlkravitz This is ready for your review

webroot/src/pages/PublicDashboard/PublicDashboard.ts
```diff
@@ -89,13 +96,21 @@ export default class DashboardPublic extends Vue {
             refresh_token: 'mock_refresh_token'
         };
 
+        authStorage.removeItem(AUTH_LOGIN_GOTO_PATH);
+        authStorage.removeItem(AUTH_LOGIN_GOTO_PATH_AUTH_TYPE);
         await this.$store.dispatch('user/updateAuthTokens', { tokenResponse: data, authType: AuthTypes.STAFF });
         this.$store.dispatch('user/loginSuccess', AuthTypes.STAFF);
 
-        this.$router.push({ name: 'Home' });
+        if (goto && (!gotoAuthType || gotoAuthType === AuthTypes.STAFF)) {
```
jlkravitz, line 104: Can you explain what this is doing in layman's terms?

jlkravitz (commented):
@isabeleliassen good to merge

jlkravitz (approved):
@isabeleliassen good to merge

PR #904, opened 2025-07-02
Frontend/compact state live control

### LINKED ISSUE #853: Move authroize.net form below compact config form
### Notes
Move the authorize.net credentials form below the compact config form in the Settings UI.

The reasoning is that the compact config form has to be completed (initially) before the authorize.net form can be completed. Ongoing the order doesn't matter, but initially it would be best to align the form in process order.

Image

### LINKED ISSUE #880: Compacts control last enablement of state configs FE
Compacts need an added UI control to enable states as "fully ready". This is in addition to the current ability for Compacts to invite states and for states to mark themselves as ready.

Dependent on:
- #868 
- #871

jsandoval81:
### Requirements List
- _None_

### Description List
- Updated the `CompactConfig` type with the new `configuredStates` prop
- Updated the existing `CompactSettingsConfig` component to reflect the required `configuredStates` prop back to the server
- Created the `StateSettingsList` component
- Updated the `CompactSettings` page to:
    - Use the new `StateSettingsList` component
    - Move the `PaymentProcessorConfig` component below the `CompactSettingsConfig` component in the UI (#853)
- Updated the mock data `staffAccount` permissions to have some variety in state-admin permissions

### Testing List
- Code review
- Testing
    - Ensure that a staff user sees the appropriate capabilities in the compact Settings page **state list**:
        - _For devs, it's easiest to use the mock API here and adjust the mock data `staffAccount` permissions as needed_
        - If the staff user is a compact-admin only, the state list should only show the "Status" column, with either "Live" or "Enable" for each state
        - If the staff user is a state-admin only of multiple states, the state list should only show the "Configuration" column, with and "Edit" button for each state
        - If the staff user is a state-admin only of only 1 state, they should be auto redirected to the state config form for that state (same as previous)
        - If the staff user is a compact-admin **and** a state-admin of 1 or more states, the state list should show the "Status" and "Configuration columns, with the appropriate "Live / Enable" and "Edit" options for each state based on their permissions
    - For compact admins, ensure that for any non-live states, the enable process works as designed
    - For compact admins, ensure the existing compact configuration form (with pricing, emails, etc) still works as expected

Closes #880 
Closes #853

### DIFF
--- a/webroot/src/components/CompactSettingsConfig/CompactSettingsConfig.ts
+++ b/webroot/src/components/CompactSettingsConfig/CompactSettingsConfig.ts
@@ -256,6 +256,7 @@ class CompactSettingsConfig extends mixins(MixinForm) {
                 },
             },
             licenseeRegistrationEnabled: isRegistrationEnabled,
+            configuredStates: this.initialCompactConfig?.configuredStates || [],
         };
 
         // Call the server API to update
--- a/webroot/src/components/StateSettingsList/StateSettingsList.ts
+++ b/webroot/src/components/StateSettingsList/StateSettingsList.ts
@@ -1,0 +1,329 @@
+//
+//  StateSettingsList.ts
+//  CompactConnect
+//
+//  Created by InspiringApps on 7/1/2025.
+//
+
+import {
+    Component,
+    mixins,
+    Watch,
+    toNative
+} from 'vue-facing-decorator';
+import { reactive, nextTick } from 'vue';
+import { AuthTypes } from '@/app.config';
+import MixinForm from '@components/Forms/_mixins/form.mixin';
+import InputButton from '@components/Forms/InputButton/InputButton.vue';
+import InputSubmit from '@components/Forms/InputSubmit/InputSubmit.vue';
+import LoadingSpinner from '@components/LoadingSpinner/LoadingSpinner.vue';
+import Modal from '@components/Modal/Modal.vue';
+import { Compact, CompactType } from '@models/Compact/Compact.model';
+import { CompactPermission, StatePermission } from '@models/StaffUser/StaffUser.model';
+import { State } from '@/models/State/State.model';
+import { FormInput } from '@/models/FormInput/FormInput.model';
+import { dataApi } from '@network/data.api';
+
+interface StateConfigRowPermission {
+    state?: State;
+    isLiveForCompact?: boolean;
+    isCompactAdmin?: boolean;
+    isStateAdmin?: boolean;
+}
+
+@Component({
+    name: 'StateSettingsList',
+    components: {
+        LoadingSpinner,
+        Modal,
+        InputButton,
+        InputSubmit,
+    },
+})
+class StateSettingsList extends mixins(MixinForm) {
+    //
+    // Data
+    //
+    isLoading = false;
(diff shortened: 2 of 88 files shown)


webroot/src/components/StateSettingsList/StateSettingsList.ts
```diff
@@ -0,0 +1,337 @@
...
+        InputSubmit,
+    },
+})
+class StateSettingsList extends mixins(MixinForm) {
+    //
+    // Data
+    //
+    isLoading = false;
+    loadingErrorMessage = '';
+    initialCompactConfig: any = {};
```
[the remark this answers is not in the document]
jsandoval81: In general, we definitely try to avoid `any` types. Our exception to this is server responses due to keeping overly verbose typescript syntax from cluttering things that get reorganized as typed models, etc. `initialCompactConfig` here is a server response, and typing as `any` prevents us from having to double cast as `unknown` when we update its value from the server.

rmolinares (approved):

jsandoval81:

webroot/src/components/CompactSettingsConfig/CompactSettingsConfig.ts
```diff
@@ -256,6 +256,7 @@ class CompactSettingsConfig extends mixins(MixinForm) {
                 },
```
jlkravitz, line 256: image

1. I don't feel strongly but just flagging the "Back" button shows up even if you are auto-redirected to the state admin config page (because you only are an admin of one state).
2. Relatedly, on both Safari and Chrome, hovering over the Back link does not always put it in a hover state (such that it is clickable). It's acting oddly. Is that something you can reproduce or is that just for me?

webroot/src/components/CompactSettingsConfig/CompactSettingsConfig.ts
```diff
@@ -256,6 +256,7 @@ class CompactSettingsConfig extends mixins(MixinForm) {
...
             },
             licenseeRegistrationEnabled: isRegistrationEnabled,
+            configuredStates: this.initialCompactConfig?.configuredStates || [],
```
jlkravitz, line 259: Question: Should a compact admin be able to "Enable" a state, even if the state has not enabled itself for registration? What's the permissions flow supposed to look like here?

jlkravitz (commented):
two small items!

webroot/src/components/CompactSettingsConfig/CompactSettingsConfig.ts
jsandoval81: @jlkravitz 

1. The "Back" button is in the Figma designs either way. It just takes the user back to the previous page, which serves all cases.
2. Good catch. I just pushed a fix for that.

webroot/src/components/CompactSettingsConfig/CompactSettingsConfig.ts
jsandoval81: @jlkravitz I believe the backend would only return a state as enable-able if it were at that stage of the workflow. Otherwise the UI won't show an "Enable" button for that state.

isabeleliassen (approved):

PR #914, opened 2025-07-07
Fix/sprint 26 hotfixes

landonshumway-ia:
Hot fixes related to missing NPDB categories and checking for payment processor fields in compact config before going live

## Summary by CodeRabbit

* **New Features**
  * Added new categories for clinical privilege actions, including Non-compliance, Criminal Conviction, Confidentiality Violation, and Misconduct or Abuse.
  * Introduced a property to expose payment processor public fields in compact configuration data.

* **Bug Fixes**
  * Updating compact configuration now preserves existing payment processor fields unless explicitly changed.

* **Refactor**
  * Improved validation for enabling or disabling licensee registration, ensuring payment processor credentials are present when required.

* **Tests**
  * Added and updated tests to verify configuration updates, field preservation, and validation of licensee registration changes.

### DIFF
--- a/backend/compact-connect/lambdas/python/common/cc_common/data_model/compact_configuration_client.py
+++ b/backend/compact-connect/lambdas/python/common/cc_common/data_model/compact_configuration_client.py
@@ -99,15 +99,48 @@ class CompactConfigurationClient:
 
     def save_compact_configuration(self, compact_configuration: CompactConfigurationData) -> None:
         """
-        Save the compact configuration.
+        Save the compact configuration, preserving existing fields like paymentProcessorPublicFields.
+        If a record exists, it merges the new values with the existing record to preserve all fields.
 
         :param compact_configuration: The compact configuration data
         """
         logger.info('Saving compact configuration', compactAbbr=compact_configuration.compactAbbr)
 
-        serialized_compact = compact_configuration.serialize_to_database_record()
+        try:
+            existing_compact_config = self.get_compact_configuration(compact_configuration.compactAbbr)
+        except CCNotFoundException:
+            logger.info('Existing compact configuration not found.', compact=compact_configuration.compactAbbr)
+            existing_compact_config = None
 
-        self.config.compact_configuration_table.put_item(Item=serialized_compact)
+        if existing_compact_config:
+            # Record exists - merge with existing data to preserve fields like paymentProcessorPublicFields
+            logger.info('Updating existing compact configuration record', compactAbbr=compact_configuration.compactAbbr)
+
+            # Load the existing record into a data class to get the existing data
+            existing_data = existing_compact_config.to_dict()
+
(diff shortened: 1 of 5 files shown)


backend/compact-connect/lambdas/python/common/cc_common/data_model/compact_configuration_client.py
```diff
@@ -99,15 +99,48 @@ def get_compact_configuration(self, compact: str) -> CompactConfigurationData:
...
+
+            # Get the new data
+            new_data = compact_configuration.to_dict()
+
+            # Merge the data - new values override existing ones, but existing fields not in new_data are preserved
+            merged_data = existing_data.copy()
+            merged_data.update(new_data)
+
+            # Handle the special case where transactionFeeConfiguration should be removed
+            # If the new configuration doesn't have transactionFeeConfiguration, remove it from merged data
```
jlkravitz, line 130: Hmm, why?

landonshumway-ia: In the case of the `transactionFeeConfiguration` field, if the compact previously set this but then removed it by setting it to 0, our current behavior is to remove the field from the configuration object entirely ([see here where we manage that in the handler logic](https://github.com/csg-org/CompactConnect/blob/5a8e5a73812c86c95aa9390c8c9de6b331086269/backend/compact-connect/lambdas/python/compact-configuration/handlers/compact_configuration.py#L229-L233)) Because we are performing a merge here by calling `update` this will not account for fields that need to be removed. Currently this is the only field that can be removed after previously being set, so we need to account for that here.

backend/compact-connect/lambdas/python/common/cc_common/data_model/compact_configuration_client.py
```diff
@@ -99,15 +99,48 @@ def get_compact_configuration(self, compact: str) -> CompactConfigurationData:
...
-        serialized_compact = compact_configuration.serialize_to_database_record()
-
-        self.config.compact_configuration_table.put_item(Item=serialized_compact)
+        try:
+            existing_compact_config = self.get_compact_configuration(compact_configuration.compactAbbr)
+        except CCNotFoundException:
+            logger.info('Existing compact configuration not found.', compact=compact_configuration.compactAbbr)
+            existing_compact_config = None
+
+        if existing_compact_config:
```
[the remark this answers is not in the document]
landonshumway-ia: When you brought up the issues of the 502s you were encountering from the privilege options endpoint, I realized that this endpoint was discarding the payment related fields set by the other credentials endpoint whenever it is called, since we were calling `put_item` which will replace whatever data is not specified in the object you send in.

jlkravitz (approved):
@isabeleliassen good to merge1

Issue #944, opened 2025-07-21
Return encumbrance details in timeline if user has admin privileges


ChiefStief:
### Notes

- Staff users and licensee users (their own) will have access to encumbrance details
- Modify privilege encumbrance updates in (license and privilege encumbrances) to include adverse action id and / or details if significant performance difference.
- Tweak backend to include deactivation details note (previous timeline work did not serialize note in correctly and does not send it)
- Confirm front end is receiving and displaying all notes:

1. Encumbrance details when applicable
2. Deactivation note
3. Encumbrance and encumbrance_lifted events will have "uploaded to system: <createDate>" in their notes

### Estimate
Backend 5
Frontend 1

### Tasks
- [ ] TODO
- [ ] Automated tests
- [ ] API Docs
- [ ] Postman collection
- [ ] PR opened with labels / reviewers / assignee / linked-issue

PR #947, opened 2025-07-21
Feat/backend timeline

### LINKED ISSUE #770: Adverse Actions - backend generates 'synthetic' history events in provider detail response
As part of showing the history of privilege records, currently the frontend is handling filling in history events that are not stored in the database, but are useful for displaying visual events on the timeline. To simplify the frontend logic and keep event data changes from breaking the UI. This ticket involves moving that history event generation logic into the backend.

As part of this, we need to update our 'dateOfUpdate' logic to be able to specify specific date of update timestamps when generating data events, as there are some cases where we do not want to automatically set the 'dateOfUpdate' field to the current timestamp.

### Tasks
- [ ] TODO
- [ ] Automated tests
- [ ] API Docs
- [ ] Postman collection

### Implementation Notes
This logic was previously defined in this draft PR: https://github.com/csg-org/CompactConnect/pull/764 but the update broke many of our tests that were dependent on static json files and the needed test update was too complex to manage within the same PR it was originally introduced in. This logic can be re-used when we implement this change.

ChiefStief:
### Requirements List
-

### Description List
- Added methods in provider_record_util to help create generate clean history timeline with synthetic expiration and issuance events injected
- Added 3 endpoints(public, staff, provider users) that all route to the same get privilege history lambda with correct auth handling
- Added `get_privilege_data` method to get privilege record and optionally all associated adverse action and updates
- created response schemas
- added and updated tests as needed

### Testing List
- run tests
- Code review

Closes #770

## Summary by CodeRabbit

* **New Features**
  * Introduced new API endpoints to retrieve provider privilege history for authenticated users, public, and staff access.
  * Added a Lambda function and integrated it with API Gateway to serve privilege history data.
  * Implemented public-facing privilege history response models and schemas for consistent API responses.
  * Enhanced privilege history data with synthetic events for clearer event timelines.
  * Improved privilege data querying with detailed and summary retrieval options.

* **Tests**
  * Added comprehensive unit and integration tests to validate privilege history enrichment and API endpoint responses.

* **Documentation**
  * Updated API models to include privilege history response structure.

### DIFF
--- a/backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
+++ b/backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
@@ -937,11 +937,11 @@ class DataClient:
         *,
         compact: str,
         provider_id: str,
-        detail: bool = False,
         jurisdiction: str,
         license_type_abbr: str,
         consistent_read: bool = False,
-    ) -> list [dict]:
+        detail: bool = False,
+    ) -> list[dict]:
         """
         Get a privilege for a provider in a jurisdiction of the license type
 
@@ -949,6 +949,7 @@ class DataClient:
         :param str provider_id: The provider of the privilege
         :param str jurisdiction: The jurisdiction of the privilege
         :param str license_type_abbr: The license type abbreviation of the privilege
+        :param bool detail: Boolean determining whether we include associated records or just privilege record itself
         :raises CCNotFoundException: If the privilege record is not found
         :return If detail = False list of length one containing privilege item, if detail = True list containing,
         privilege record, privilege update records and privilege adverse action records
--- a/backend/compact-connect/lambdas/python/common/cc_common/data_model/provider_record_util.py
+++ b/backend/compact-connect/lambdas/python/common/cc_common/data_model/provider_record_util.py
@@ -240,10 +240,10 @@ class ProviderRecordUtility:
     ) -> list[dict]:
         """
         Enrich the privilege history with 'synthetic updates'.
-        Synthetic updates are what we're calling critical pieces of history that are not explicitly recorded in the data
-        system, because they occur passively, such as when a privilege expires or that we do not participate
-        in, like when a privilege is issued. These 'synthetic updates' do not have a corresponding record in the
-        database, but we can deduce their existence based on the privilege's other data. Because these events are
+        Synthetic updates are pieces of history that are not explicitly recorded in the data
+        system, because they occur passively, such as when a privilege expires or because they are redundant.
+        These 'synthetic updates' do not have a corresponding record in the database, but we can deduce their
+        existence based on the privilege's other data. Because these events are
         'synthetic', they have no actual changes in record values associated with them.
         Example issuance event:
         {
@@ -341,7 +341,8 @@ class ProviderRecordUtility:
     def construct_simplified_privilege_history_object(privilege_data: list[dict]) -> dict:
         """
         Construct a simplified list of history events to be easily consumed by the front end
-        :param privilege_data: All of the records relative to the privilege: the privilege, updates, and adverse actions
+        :param privilege_data: All of the records associated with the privilege:
+        the privilege, updates, and adverse actions
         :return: The simplified and enriched privilege history
         """
         privilege = list(filter(lambda x: x['type'] == 'privilege', privilege_data))[0]
--- a/backend/compact-connect/lambdas/python/common/tests/unit/test_provider_record_util.py
+++ b/backend/compact-connect/lambdas/python/common/tests/unit/test_provider_record_util.py
@@ -619,8 +619,7 @@ class TestProviderRecordUtility(TstLambdas):
     @patch('cc_common.config._Config.current_standard_datetime', datetime.fromisoformat('2026-03-15T00:00:00+00:00'))
     def test_enrich_privilege_history_does_not_inject_expiration_if_renewed_in_last_minute(self):
         """Test that get_enriched_history_with_synthetic_updates_from_privilege does not injection expiration event if
-        renewed last
-        minute
+        renewed last minute
         """
         from cc_common.data_model.provider_record_util import ProviderRecordUtility
 


backend/compact-connect/lambdas/python/common/cc_common/data_model/provider_record_util.py
```diff
@@ -220,6 +230,116 @@ def assemble_provider_records_into_api_response_object(provider_records: list[di
         return provider
 
 
+    @classmethod
+    def get_enriched_history_with_synthetic_updates_from_privilege(cls, privilege: dict, history: list[dict]) -> list[dict]:
+        """
```
landonshumway-ia, line 235: The parameter here says `privilege: dict` but the comment says it can handle both privileges and licenses, and gives an example of a license issuance event. The code looks like its hardcoded to deal with privilege updates, but we could have this handle both with a simple check to see if the 'type' of the passed in record is either a license or privilege right? If we don't want to make that change here, we should update the comment to reflect its intended use.

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
```diff
@@ -932,6 +931,45 @@ def process_registration_values(
...
+        self,
+        *,
+        compact: str,
+        provider_id: str,
+        detail: bool = False,
+        dynamo_pagination: dict,
+        jurisdiction: str,
+        license_type_abbr: str,
+        consistent_read: bool = False,
+    ):
```
landonshumway-ia, line 946: Please add a return type annotation here 
```suggestion
    ) -> dict:
```

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
```diff
@@ -932,6 +931,45 @@ def process_registration_values(
...
+        """
+        Get a privilege for a provider in a jurisdiction of the license type
+
+        :param str compact: The compact of the privilege
+        :param str provider_id: The provider of the privilege
+        :param str jurisdiction: The jurisdiction of the privilege
+        :param str license_type_abbr: The license type abbreviation of the privilege
```
landonshumway-ia, line 953: please add a `:return` line to this doc explaining exactly what is intended to be returned by this endpoint

backend/compact-connect/lambdas/python/provider-data-v1/handlers/privilege_history.py
```diff
@@ -0,0 +1,92 @@
...
+        # the two values for compact and providerId are stored as custom attributes in the user's cognito claims
+        # so we can access them directly from the event object
+        compact = event['requestContext']['authorizer']['claims']['custom:compact']
+        provider_id = event['requestContext']['authorizer']['claims']['custom:providerId']
+    except (KeyError, TypeError) as e:
+        # This shouldn't happen unless a provider user was created without these custom attributes.
+        logger.error(f'Missing custom provider attribute: {e}')
+        raise CCInvalidRequestException('Missing required user profile attribute') from e
+
+    return compact, provider_id
```
landonshumway-ia, line 91: We should extract this into a common method, since I assume this was copied over from another file that has the same need. We should aim for the DRY principle (Don't Repeat Yourself) and extract common logic into a place where both files can reference it, this will reduce duplicate code which will make the project easier to maintain over time.

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_privilege_history.py
```diff
@@ -0,0 +1,231 @@
...
+             'privilegeId': 'SLP-NE-1',
+             'providerId': '89a6377e-c3a5-40e5-bca5-317ec854c570'
+        }
+
+        self.assertEqual(expected_history, history_data)
+
+    def test_get_privilege_history_public_returns_expected_history(self):
+        from handlers.privilege_history import privilege_history_handler
+
+        self._load_provider_data()
```
landonshumway-ia, line 137: We want to avoid using `self._load_provider_data()` for future tests and move over towards using the test_data_generator wherever possible for generating test data. This will make the tests easier to modify in the future.

Obviously, you've already put in the effort using the old pattern, so we don't need to refactor it now as long as the tests are valid (I'm still reviewing them). We should probably leave a comment over that _load_provider_data() method to denote that it is deprecated and recommend devs to use the test data generator. I'll take care of that in my current ticket.

No action item for this, just something to be aware of.

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_privilege_history.py
```diff
@@ -0,0 +1,231 @@
...
+            event['httpMethod'] = 'GET'
+            event['resource'] = (
+                '/v1/compacts/{compact}/providers/{providerId}/privileges/jurisdiction/{jurisdiction}/licenseType/{licenseType}/history'
+            )
+            event['pathParameters'] = {
+                'jurisdiction': 'ma',
+                'licenseType': 'slp',
+                'compact': test_provider.compact,
+                'providerId': test_provider.providerId,
+            }
```
landonshumway-ia, line 92: This test setup is copied and pasted into almost every test that isn't calling the provider users endpoint, so it should be extracted into a common private method and parameterized for passing in the URL we want to call and the path parameters, to follow the DRY principle. That would reduce this file by at least 25%

backend/compact-connect/lambdas/python/common/cc_common/data_model/provider_record_util.py
```diff
@@ -224,6 +230,131 @@ def assemble_provider_records_into_api_response_object(provider_records: list[di
...
+        }] + create_date_sorted_original_history
+
+        renewal_updates = list(filter(lambda x: x['updateType'] == UpdateCategory.RENEWAL, enriched_history))
+
+        now = config.current_standard_datetime
+
+        # Inject expiration events that occurred between events
+        for update in renewal_updates:
+            date_of_expiration = update['previous']['dateOfExpiration']
+            datetime_of_expiration = datetime.fromisoformat(date_of_expiration.isoformat() + 'T00:00:00+00:00')
```
landonshumway-ia, line 288: I'm not sure this is quite accurate timezone-wise. According to the requirements we were given, a privilege is considered expired once the day has lapsed in UTC-4 time (we have a [config property specifically for getting this value](https://github.com/csg-org/CompactConnect/blob/6853418bc0d03d2990676c79f5c5af9bc013bbc1/backend/compact-connect/lambdas/python/common/cc_common/config.py#L216C1-L224C101)) so you should be able to compare the date of this property to the date of the renewal, rather than trying to compare datetimes.

backend/compact-connect/lambdas/python/common/cc_common/data_model/provider_record_util.py
```diff
@@ -224,6 +230,131 @@ def assemble_provider_records_into_api_response_object(provider_records: list[di
...
+            if date_of_expiration < update['createDate'].date():
```
landonshumway-ia, line 289: do we want to compare the `createDate` or the `effectiveDate`? `createDate`  always matches `effectiveDate` for now, but this may not always be true. We can imagine some case in the future where the compacts allow practitioners to purchase renewals in advance, for example. If we compare this against the `effectiveDate` of the renewal event, it should hopefully be more robust to future changes.

backend/compact-connect/lambdas/python/common/cc_common/data_model/provider_record_util.py
```diff
@@ -224,6 +230,131 @@ def assemble_provider_records_into_api_response_object(provider_records: list[di
...
+                    'previous': {},
+                    'updatedValues': {},
+                    'dateOfUpdate': privilege_datetime_of_expiration,
+                }
+            )
+
+        return sorted(enriched_history, key=lambda x: x['effectiveDate'])
+
+    @classmethod
+    def construct_simplified_public_privilege_history_object(cls, privilege_data: list[dict]) -> dict:
```
landonshumway-ia, line 331: The method name includes `public_privilege_history_object` but this is used by more than just the public endpoint, correct?

backend/compact-connect/lambdas/python/common/cc_common/data_model/provider_record_util.py
landonshumway-ia: Actually, the privilege wouldn't be considered expired until the day **after** the value of the `dateOfExpiration` field (ie 2025-10-10 is still active on that day, and does not expire until 2025-10-11) so you would need to add one day to the date_of_expiration date object to compare accurately. You can do this using the `timedelta`
```suggestion
            day_after_expiration = date_of_expiration + timedelta(days=1)
            datetime_of_expiration = datetime.fromisoformat(day_after_expiration, datetime.min.time(), tzinfo=config.expiration_resolution_timezone)
```

backend/compact-connect/lambdas/python/provider-data-v1/handlers/privilege_history.py
```diff
@@ -0,0 +1,92 @@
...
+        case (
+            'GET',
+            '/v1/public/compacts/{compact}/providers/{providerId}/jurisdiction/{jurisdiction}/licenseType/{licenseType}/history'
+        ):
+            return _get_privilege_history(event)
+        case (
+            'GET',
+            '/v1/compacts/{compact}/providers/{providerId}/privileges/jurisdiction/{jurisdiction}/licenseType/{licenseType}/history'
+        ):
+            return _get_privilege_history(event)
```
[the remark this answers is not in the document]
ChiefStief: We broke that out into its own ticket because there are still a few requirement details to nail down and including the encumbrance details will either require another db call or saving that info into the update as currently that data is not part of the update.

Here is the ticket: https://github.com/csg-org/CompactConnect/issues/944

backend/compact-connect/lambdas/python/common/cc_common/data_model/provider_record_util.py
landonshumway-ia: Also, we should probably compare the full datetimes here, since we know the timezone by which these should be compared by 
```suggestion
            if datetime_of_expiration < update['effectiveDate'].astimezone(config.expiration_resolution_timezone):
```

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
```diff
@@ -932,6 +931,45 @@ def process_registration_values(
...
+        :raises CCNotFoundException: If the privilege record is not found
```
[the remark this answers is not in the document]
ChiefStief: This can return either just the item or the item and all updates and adverse actions depending if you pass the detail = True prop in. I modeled it after `data_client.get_provider`.

The idea was to only make one db call to get all the data I needed, theoretically I could make the single item and the detail version just different methods, but I took cues from `data_client.get_provider` for the best way to do this.

For the @paginated_query I thought this was necessary for paginating dynamo requests specifically. Because we could theoretically be returning many updates and adverse actions that may be paginated, I kept it in there.

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
```diff
@@ -973,13 +1011,12 @@ def deactivate_privilege(
...
-        privilege_record = self.get_privilege(
+
+        privilege_data = self.get_privilege_data(
             compact=compact, provider_id=provider_id, jurisdiction=jurisdiction, license_type_abbr=license_type_abbr
         )
 
-        # Find the main privilege record (not history records)
-        privilege_record_schema = PrivilegeRecordSchema()
-        privilege_record = privilege_record_schema.load(privilege_record)
+        privilege_record = privilege_data['items'][0]
```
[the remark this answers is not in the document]
ChiefStief: Here I am using not using `detail = True` so it is just the item itself, for the feature I added this time I am using `detail = True` to get the history and privilege item in one db call.

backend/compact-connect/lambdas/python/common/cc_common/data_model/provider_record_util.py
```diff
@@ -224,6 +230,131 @@ def assemble_provider_records_into_api_response_object(provider_records: list[di
...
+                    'createDate': privilege_datetime_of_expiration,
+                    'previous': {},
+                    'updatedValues': {},
+                    'dateOfUpdate': privilege_datetime_of_expiration,
+                }
+            )
+
+        return sorted(enriched_history, key=lambda x: x['effectiveDate'])
+
+    @classmethod
```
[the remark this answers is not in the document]
ChiefStief: this one does use cls

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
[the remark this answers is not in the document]
landonshumway-ia: There are several examples in this file that provide an example for getting all pages from a query which you should be able to reference.

backend/compact-connect/lambdas/python/common/cc_common/data_model/provider_record_util.py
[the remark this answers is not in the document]
landonshumway-ia: I see, ok if you make the previous method static, then you should be able to make this method static as well and call the other method as a static method `ProviderRecordUtility.get_enriched_history_with_synthetic_updates_from_privilege(privilege, history)`

backend/compact-connect/lambdas/python/common/tests/unit/test_provider_record_util.py
```diff
@@ -257,7 +257,63 @@ def test_enrich_privilege_history_with_expiration_event_if_expired(self):
...
+        """Test that enrich_privilege_history_with_synthetic_updates adds an expiration if expired"""
+        from cc_common.data_model.provider_record_util import ProviderRecordUtility
+
+        # Create a privilege with no history
+        privilege = {
+            **self.base_privilege,
+            'providerId': 'test-provider-id',
+            'dateOfIssuance': datetime.fromisoformat('2024-01-01T00:00:00+00:00'),
+            'dateOfExpiration': date.fromisoformat('2024-02-01'),
+            'dateOfRenewal': date.fromisoformat('2024-01-01'),
```
landonshumway-ia, line 272: dateOfRenewal should be a datetime for privileges, if we want our test data to be consistent with the actual schema.

backend/compact-connect/lambdas/python/common/tests/unit/test_provider_record_util.py
```diff
@@ -257,7 +257,63 @@ def test_enrich_privilege_history_with_expiration_event_if_expired(self):
...
+                'licenseType': 'occupational therapy assistant',
+                'previous': {},
+                'providerId': 'test-provider-id',
+                'type': 'privilegeUpdate',
+                'updateType': 'expiration',
+                'updatedValues': {},
+            },
+        ]
+
+        # Check that the history contains exactly one update with the expected values
```
landonshumway-ia, line 313: This comment doesn't quite line up, since there are two updates in this case: one issuance and one expiration.

backend/compact-connect/lambdas/python/common/tests/unit/test_provider_record_util.py
```diff
@@ -300,6 +356,49 @@ def test_enrich_privilege_history_does_not_add_expiration_if_day_of_expiration(s
         self.maxDiff = None
         self.assertEqual(expected_updates, enriched_history)
 
+    @patch('cc_common.config._Config.current_standard_datetime', datetime.fromisoformat('2024-03-15T23:59:00+04:00'))
+    def test_enrich_privilege_history_does_not_add_expiration_if_minute_before_expiration(self):
```
landonshumway-ia, line 362: Good use of patching to test the expiration edge cases 👍

backend/compact-connect/lambdas/python/common/tests/unit/test_provider_record_util.py
```diff
@@ -300,6 +356,49 @@ def test_enrich_privilege_history_does_not_add_expiration_if_day_of_expiration(s
...
+        """Test that enrich_privilege_history_with_synthetic_updates does not add expiration on day of expiration."""
+        from cc_common.data_model.provider_record_util import ProviderRecordUtility
+
+        # Create a privilege with no history
+        privilege = {
+            **self.base_privilege,
+            'providerId': 'test-provider-id',
+            'dateOfIssuance': datetime.fromisoformat('2024-01-01T00:00:00+00:00'),
+            'dateOfExpiration': date.fromisoformat('2024-03-15'),
+            'dateOfRenewal': date.fromisoformat('2024-01-01'),
```
landonshumway-ia, line 373: dateOfRenewal should be a datetime to be consistent with the schema.

backend/compact-connect/lambdas/python/common/tests/unit/test_provider_record_util.py
```diff
@@ -512,3 +611,216 @@ def test_enrich_privilege_history_adds_expiration_events_in_correct_spots(self):
...
+        """Test that enrich_privilege_history_with_synthetic_updates adds expiration and issuance events correctly"""
+        from cc_common.data_model.provider_record_util import ProviderRecordUtility
+
+        # Create a privilege with no history
+        privilege = {
+            **self.base_privilege,
+            'providerId': 'test-provider-id',
+            'dateOfIssuance': datetime.fromisoformat('2024-01-01T00:00:00+00:00'),
+            'dateOfExpiration': date.fromisoformat('2028-03-15'),
+            'dateOfRenewal': date.fromisoformat('2024-01-01'),
```
landonshumway-ia, line 633: Same here

backend/compact-connect/lambdas/python/common/tests/unit/test_provider_record_util.py
```diff
@@ -512,3 +611,216 @@ def test_enrich_privilege_history_adds_expiration_events_in_correct_spots(self):
...
+                    'compactTransactionId': '120060004893',
+                    'dateOfExpiration': date.fromisoformat('2028-02-12'),
+                    'dateOfRenewal': datetime.fromisoformat('2025-03-25T19:03:56+00:00'),
+                    'privilegeId': 'OTA-NE-10',
+                },
+                'updateType': 'renewal',
+            },
+        ]
+
+        # Check that the history contains exactly one update with the expected values
```
landonshumway-ia, line 718: This comment does not line up with what the test actually expects.

backend/compact-connect/lambdas/python/common/tests/unit/test_provider_record_util.py
```diff
@@ -512,3 +611,216 @@ def test_enrich_privilege_history_adds_expiration_events_in_correct_spots(self):
...
+        """Test that enrich_privilege_history_with_synthetic_updates adds expiration and issuance events correctly"""
+        from cc_common.data_model.provider_record_util import ProviderRecordUtility
+
+        # Create a privilege with no history
+        privilege = {
+            **self.base_privilege,
+            'providerId': 'test-provider-id',
+            'dateOfIssuance': datetime.fromisoformat('2024-01-01T00:00:00+00:00'),
+            'dateOfExpiration': date.fromisoformat('2028-03-15'),
+            'dateOfRenewal': date.fromisoformat('2024-01-01'),
```
landonshumway-ia, line 734: set as datetime here as well

backend/compact-connect/lambdas/python/common/tests/unit/test_provider_record_util.py
```diff
@@ -512,3 +611,216 @@ def test_enrich_privilege_history_adds_expiration_events_in_correct_spots(self):
...
+                    'compactTransactionId': '120060004893',
+                    'dateOfExpiration': date.fromisoformat('2028-02-12'),
+                    'dateOfRenewal': datetime.fromisoformat('2025-03-25T19:03:56+00:00'),
+                    'privilegeId': 'OTA-NE-10',
+                },
+                'updateType': 'renewal',
+            },
+        ]
+
+        # Check that the history contains exactly one update with the expected values
```
landonshumway-ia, line 832: ditto on comment here

landonshumway-ia (changes requested):
Tests look good, just a couple nits

landonshumway-ia (approved):
Nice work

ChiefStief:
@jlkravitz This is ready for review.

backend/compact-connect/lambdas/python/common/cc_common/data_model/provider_record_util.py
```diff
@@ -224,6 +233,137 @@ def assemble_provider_records_into_api_response_object(provider_records: list[di
...
+                }
+            )
+
+        return sorted(enriched_history, key=lambda x: x['effectiveDate'])
+
+    @staticmethod
+    def construct_simplified_privilege_history_object(privilege_data: list[dict]) -> dict:
+        """
+        Construct a simplified list of history events to be easily consumed by the front end
+        :param privilege_data: All of the records relative to the privilege: the privilege, updates, and adverse actions
```
jlkravitz, line 344: maybe? is `privilege_data` the raw data or the augmented synthetic data?

```suggestion
        :param privilege_data: All of the records associated with the privilege: the privilege, updates, and adverse actions
```

backend/compact-connect/lambdas/python/common/cc_common/utils.py
```diff
@@ -696,3 +699,27 @@ def send_licenses_to_preprocessing_queue(licenses_data: list[dict], event_time:
 
     # Return success status and failure count
     return failed_license_numbers
+
+def load_records_into_schemas(records: list[dict]):
+    """Load records into their defined schema"""
+    try:
+        return [BaseRecordSchema.get_schema_by_type(item['type']).load(item) for item in records]
```
jlkravitz, line 706: What ensures 'type' exists as a key here?

backend/compact-connect/lambdas/python/common/tests/unit/test_provider_record_util.py
```diff
@@ -135,3 +159,676 @@ def test_find_best_license_complex_scenario(self):
...
+        """Test that get_enriched_history_with_synthetic_updates_from_privilege adds an expiration if first second
+        of expired"""
+        from cc_common.data_model.provider_record_util import ProviderRecordUtility
+
+        # Create a privilege with no history
+        privilege = {
+            **self.base_privilege,
+            'providerId': 'test-provider-id',
+            'dateOfIssuance': datetime.fromisoformat('2024-01-01T00:00:00+00:00'),
+            'dateOfExpiration': date.fromisoformat('2024-02-01'),
```
jlkravitz, line 271: hmm, aren't these two (`dateOfIssuance` and `dateOfExpiration`) not one second apart, as the test name implies?

backend/compact-connect/lambdas/python/common/tests/unit/test_provider_record_util.py
```diff
@@ -135,3 +159,676 @@ def test_find_best_license_complex_scenario(self):
...
+
+        # Check that the history contains updates with the expected values
+        self.maxDiff = None
+        self.assertEqual(expected_updates, enriched_history)
+
+    @patch('cc_common.config._Config.current_standard_datetime', datetime.fromisoformat('2026-03-15T00:00:00+00:00'))
+    def test_enrich_privilege_history_does_not_inject_expiration_if_renewed_in_last_minute(self):
+        """Test that get_enriched_history_with_synthetic_updates_from_privilege does not injection expiration event if
+        renewed last
+        minute
```
jlkravitz, line 623:
```suggestion
        renewed last minute
```

backend/compact-connect/lambdas/python/common/cc_common/utils.py
ChiefStief: All of the db records have the type param, this is only to be used on db records.

If type is absent in one of the db records the try, excepts should handle this.

ChiefStief:
@jlkravitz This is ready for another look

backend/compact-connect/lambdas/python/common/cc_common/data_model/provider_record_util.py
```diff
@@ -224,6 +233,137 @@ def assemble_provider_records_into_api_response_object(provider_records: list[di
...
 
+    @staticmethod
+    def get_enriched_history_with_synthetic_updates_from_privilege(
+        privilege: dict,
+        history: list[dict]
+    ) -> list[dict]:
+        """
+        Enrich the privilege history with 'synthetic updates'.
+        Synthetic updates are what we're calling critical pieces of history that are not explicitly recorded in the data
+        system, because they occur passively, such as when a privilege expires or that we do not participate
```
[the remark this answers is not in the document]
jlkravitz: Got it- thanks for the explanation! Makes sense.

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
```diff
@@ -932,6 +931,44 @@ def process_registration_values(
...
+    @logger_inject_kwargs(logger, 'compact', 'provider_id', 'detail', 'jurisdiction', 'license_type')
+    def get_privilege_data(
+        self,
+        *,
+        compact: str,
+        provider_id: str,
+        detail: bool = False,
+        jurisdiction: str,
+        license_type_abbr: str,
+        consistent_read: bool = False,
```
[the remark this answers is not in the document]
ChiefStief: oh sorry you're right. Because this is a "building block" method there could conceivably be cases where a consistent_read is important enough to warrant the costs and performance tradeoffs.

At the moment, this is not needed with the uses of this method but I think it conceivably could be so I retained having the option while defaulting to false.

Issue #924, opened 2025-07-11
Adjust Cognito security policy for blocking suspicious login attempts


landonshumway-ia:
We had a staff user that was attempting to login for the first time to the beta environment and failed multiple times over the course of several days while traveling. This combination caused the Cognito threat protection system to block the user from logging in as it was flagged as a risk of account takeover. We have some configuration we can set to determine when a user should be blocked, or when they should just be required to use MFA (which we already do by default for prod and beta)

The purpose of this ticket is to determine what configuration options we have available and configured those desired changes in Cognito, if any, as determined by management.

This may be part of MVP launch, or we may decide to address it if more issues like this pop up, as it should only require minor configuration changes to implement.

### Estimate
3

### Implementation Notes

<img width="953" height="502" alt="Image" src="https://github.com/user-attachments/assets/012c6b5a-0232-4ed7-bcd8-22cab97576e2" />

jusdino:
After doing some research, it appears that email-based MFA is not a viable option for either staff or practitioner users, as it would require significant changes to account recovery configuration, including enabling phone-based account recovery and making collection of user phone numbers mandatory.

PR #990, opened 2025-08-06
Add risk notification emails

jusdino:
### Requirements List
- Adding very basic user security notification emails for account takeover risk events

### Testing List
- Code review

Part of #924

### DIFF
--- a/backend/compact-connect/common_constructs/user_pool.py
+++ b/backend/compact-connect/common_constructs/user_pool.py
@@ -1,4 +1,5 @@
 import base64
+import os
 from collections.abc import Mapping
 
 from aws_cdk import CfnOutput, Duration, RemovalPolicy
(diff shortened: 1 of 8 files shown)


backend/compact-connect/common_constructs/user_pool.py
```diff
@@ -208,6 +215,10 @@ def add_ui_client(
         )
 
     def _add_risk_configuration(self, security_profile: SecurityProfile):
+        with open(os.path.join('resources', 'cognito-blocked-notification.txt')) as f:
+            blocked_notify_text = f.read()
+        with open(os.path.join('resources', 'cognito-no-action-notification.txt')) as f:
+            no_action_notify_text = f.read()
```
landonshumway-ia, line 221: Does Cognito not provide a default message for these notifications? If so was there a reason we chose to roll our own rather than rely on their messaging?

landonshumway-ia (approved):
Looks reasonable for a MVP solution. Just had one question but is not intended to hold up the PR.

backend/compact-connect/common_constructs/user_pool.py
jusdino: They do not provide a default, so I added some very basic messages.

PR #996, opened 2025-08-06
Turn off txn reporting for pre-live compacts

### LINKED ISSUE #766: Bug - do not create weekly reports for inactive compacts
Currently, we create transaction reports even for compacts that are not active in a given environment, which causes alerts to fire since there is no compact configuration for the reporting system to reference. We need to only create weekly/monthly reports for compacts that are registered as active within the system.

### Implementation Notes

This line needs to be updated: https://github.com/csg-org/CompactConnect/blob/089a8945c21385502a5688a448a704cfde65292a/backend/compact-connect/stacks/reporting_stack.py#L185

jusdino:
### Description List
- Update transaction history lambda to check compact config and exit gracefully if not live
- Update transaction reporting lambda to exit gracefully if compact is not live instead of raising an exception
- Re-enable alarm for transaction history step function, now that it should exit gracefully in all cases

### Testing List
- Code review

Closes #766

### DIFF
--- a/backend/compact-connect/lambdas/python/purchases/handlers/transaction_reporting.py
+++ b/backend/compact-connect/lambdas/python/purchases/handlers/transaction_reporting.py
@@ -11,7 +11,7 @@ from cc_common.config import config, logger
 from cc_common.data_model.schema.compact import Compact
 from cc_common.data_model.schema.compact.common import COMPACT_TYPE
 from cc_common.data_model.schema.jurisdiction.common import JURISDICTION_TYPE
-from cc_common.exceptions import CCInternalException, CCNotFoundException
+from cc_common.exceptions import CCInternalException
 
 SETTLEMENT_ERROR_STATE = 'settlementError'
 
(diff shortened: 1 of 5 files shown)


backend/compact-connect/lambdas/python/purchases/handlers/transaction_reporting.py
```diff
@@ -217,22 +231,6 @@ def generate_transaction_reports(event: dict, context: LambdaContext) -> dict:
     # for more information on how settlement errors are reprocessed.
     transactions = [t for t in transactions if t.get('transactionStatus') != SETTLEMENT_ERROR_STATE]
```
landonshumway-ia, line 233: So I know this has nothing to do with the current ticket description, but in our beta environment I found one of the transactions from the actual OT authorize.net account was processed with an unexpected `transactionStatus` of `generalError`. Would you mind adding a check for that as part of the filter here so those types of transactions do not get included in the report? Thanks

landonshumway-ia (changes requested):
The code changes look good. Since we are currently in this space, I found from our testing with the OT authorize.net account that there is another `transactionStatus` value, `generalError`, that did not show up in the sandbox environment which we will want to account for. I've marked the place in the reporting where these will need to be filtered, and we will also need to add a check for this status type in the conditional here: https://github.com/csg-org/CompactConnect/blob/94c24ffd1e8bb89d02b9e0f330ffcdf2a589dc9a/backend/compact-connect/lambdas/python/purchases/purchase_client.py#L790

landonshumway-ia (approved):
Thanks for adding that 👍

backend/compact-connect/lambdas/python/purchases/purchase_client.py
```diff
@@ -609,7 +618,7 @@ def _get_settled_batch_list(self, start_time: str, end_time: str) -> apicontract
             batch_ids = [str(batch.batchId) for batch in batch_response.batchList.batch]
             logger.info('Found settled batches', batch_ids=batch_ids)
             for batch in batch_response.batchList.batch:
-                if batch.settlementState == 'settlementError':
+                if batch.settlementState == AuthorizeNetTransactionErrorStates.SettlementError:
```
jlkravitz, line 621: Just confirming- this should specifically check `SettlementError` and not `in AuthorizeNetTransactionErrorStates`?

jlkravitz (commented):
one small question but otherwise looks good!

backend/compact-connect/lambdas/python/purchases/purchase_client.py
jusdino: Yes, my understanding is that this status is for the batch as a whole, instead of an individual transaction. For batches, we only want to handle the actual settlement error state.

PR #1005, opened 2025-08-09
Add ecdsa auth to state api

### LINKED ISSUE #973: Add secondary auth for state api private data out
In order to provide extra protection of PII leaving CompactConnect to a state system, we should add a second authentication factor, such as requiring a client HMAC request signature with public/private keys, or mutual TLS. Some sort of factor that would require an attacker to compromise a second independent secret from the state system, before they can successfully extract private data. We should work with state stakeholders to agree on a single mechanism that is suitable for all states that wish to sync private data to their systems.

jusdino:
### Requirements List
- Add a second authentication factor for m2m clients that retrieve private data
- Second factor must be distinct enough that a single credential compromise is not enough for an attacker to gain access

### Description List
- Added draft doc describing an signature-based authentication method
- Split auth for state-api into separate Cognito User Pool, with no users, only clients
- Added ECDSA-based secondary auth mechanism
- Instrumented required and optional signature auth on state-api endpoints
- Updated docs and scripts to reference new API, user pool for endpoints

### Testing List
- Unit tests of new signature decorators
- Function tests for public key data retrieval
- Function tests for instrumented API endpoints
- CDK tests for infrastructure
- Smoke test for ETE

Closes #973 

## Summary by CodeRabbit

* **New Features**
  * Dedicated State Auth service for machine-to-machine access and a bulk-upload URL endpoint; CLI to manage signature keys and a reference request-signing helper.

* **Security**
  * ECDSA‑SHA256 request signing (required for some read paths, optional for uploads), nonce replay protection, 1‑minute clock skew tolerance, and improved 401 responses with CORS support.

* **Documentation**
  * Signature-auth guide, updated API/OpenAPI and Postman samples, and revised onboarding/email templates.

* **Tests**
  * Expanded unit, functional, and smoke tests for signature flows.

* **Chores**
  * Dependency bumps, CI/workflow formatting tweaks, and added ignore entry for repository root.


landonshumway-ia (commented):
Looks like a good start. Have some questions/comments for discussion.

backend/compact-connect/docs/client_hmac_auth.md
```diff
@@ -0,0 +1,190 @@
...
+- Implement retry logic with exponential backoff
+
+## Client Registration
+
+To implement HMAC authentication:
+
+1. **Generate your RSA key pair** as described above
+2. **Contact CompactConnect support** to register your client
+3. **Provide your public key** (not the private key) during registration
+4. **Receive your client ID** for use in signatures
```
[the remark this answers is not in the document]
jusdino: Yeah, that could benefit from some rewording. I'll tighten it up.

backend/compact-connect/docs/client_hmac_auth.md
```diff
@@ -0,0 +1,190 @@
...
+
+Generate an ECDSA key pair using the P-256 curve for your client:
+
+```bash
+# Generate private key
+openssl ecparam -genkey -name prime256v1 -noout -out client_private_key.pem
+
+# Extract public key
+openssl ec -in client_private_key.pem -pubout -out client_public_key.pem
+```
```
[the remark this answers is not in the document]
jusdino: Theoretically, if they install OpenSSL, I think so, but I can't readily verify that myself.

backend/compact-connect/lambdas/python/compact-configuration/requirements-dev.txt
```diff
@@ -4,9 +4,9 @@
 #
 #    pip-compile --no-emit-index-url compact-connect/lambdas/python/compact-configuration/requirements-dev.in
 #
-boto3==1.40.8
+boto3==1.40.15
     # via moto
-botocore==1.40.8
+botocore==1.40.15
     # via
```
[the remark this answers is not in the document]
jusdino: moto 5.1.10 and 5.1.11 are listed on pypi.

backend/compact-connect/requirements.txt
```diff
@@ -12,18 +12,18 @@ aws-cdk-asset-awscli-v1==2.2.242
...
+aws-cdk-lib==2.212.0
     # via
     #   -r compact-connect/requirements.in
     #   aws-cdk-aws-lambda-python-alpha
     #   cdk-nag
 cattrs==25.1.1
     # via jsii
-cdk-nag==2.37.1
+cdk-nag==2.37.8
     # via -r compact-connect/requirements.in
```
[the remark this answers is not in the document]
jusdino: aws-cdk-asset-awscli-v1 and aws-cdk-asset-node-proxy-agent-v6 are maintained and versioned independently. They are also indirect dependencies, their versions resolved based on our direct dependencies, listed in the corresponding `requirements.in` file.

backend/compact-connect/tests/smoke/hmac_auth_smoke_tests.py
```diff
@@ -0,0 +1,482 @@
...
+    load_smoke_test_env,
+)
+
+# Add the common library path to import the sign_request function
+common_lib_path = os.path.join('lambdas', 'python', 'common')
+sys.path.append(common_lib_path)
+
+# Import the sign_request function from the common library
+from common_test.sign_request import sign_request  # noqa: E402
+
```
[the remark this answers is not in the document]
jusdino: All of our CLI scripts are written assuming a specific cwd, by convention, so this is a non-issue.

backend/compact-connect/tests/smoke/dsa_auth_smoke_tests.py
```diff
@@ -0,0 +1,479 @@
...
+
+def get_state_api_base_url():
+    """Get the state API base URL from config."""
+    return config.state_api_base_url
+
+
+def get_state_auth_url():
+    """Get the state auth URL from config."""
+    return config.state_auth_url
+
```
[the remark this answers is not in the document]
jusdino: No, the config references _are_ properties and this works as written.

backend/compact-connect/stacks/state_api_stack/v1_api/bulk_upload_url.py
```diff
@@ -21,6 +24,8 @@ def __init__(
         method_options: MethodOptions,
         bulk_uploads_bucket: IBucket,
         license_upload_role: IRole,
+        compact_configuration_table: ITable,
+        persistent_stack: ps.PersistentStack,
```
landonshumway-ia, line 28: curious why we put two parameters here, when the compact config table is already nested inside of the persistent stack. It would keep the signature cleaner to just pass in the persistent stack and reference the table when needed.

backend/compact-connect/stacks/state_api_stack/v1_api/post_licenses.py
```diff
@@ -35,15 +36,24 @@ def __init__(
...
         self.api.log_groups.extend(self.log_groups)
 
     def _add_post_license(
-        self, method_options: MethodOptions, license_preprocessing_queue: IQueue, license_upload_role: IRole
+        self,
+        method_options: MethodOptions,
+        license_preprocessing_queue: IQueue,
+        license_upload_role: IRole,
+        compact_configuration_table: ITable,
+        persistent_stack: ps.PersistentStack,
```
landonshumway-ia, line 50: same here, if you're passing in the persistent stack, can't we just reference the config table when needed?

backend/compact-connect/tests/smoke/smoke_tests_env_example.json
```diff
@@ -1,5 +1,8 @@
 {
   "CC_TEST_API_BASE_URL": "https://api.test.compactconnect.org",
+  "CC_TEST_STATE_API_BASE_URL": "https://state-api.test.compactconnect.org",
+  "CC_TEST_STATE_AUTH_URL": "https://compact-connect-state-auth-test.auth.us-east-1.amazoncognito.com/oauth2/token",
+  "CC_TEST_COGNITO_STATE_AUTH_USER_POOL_ID": "us-east-1_12345",
```
landonshumway-ia, line 5: Thanks for keeping this up to date

landonshumway-ia (commented):
The code for this looks really clean. Nice work! Most of my comments are around documentation and testing.

backend/compact-connect/lambdas/python/common/cc_common/signature_auth.py
```diff
@@ -0,0 +1,412 @@
...
+from datetime import UTC, datetime
+from functools import wraps
+from typing import Any
+from urllib.parse import quote
+
+from boto3.dynamodb.conditions import Attr, Key
+from botocore.exceptions import ClientError
+from cryptography.exceptions import InvalidSignature
+from cryptography.hazmat.primitives import hashes, serialization
+from cryptography.hazmat.primitives.asymmetric import ec
```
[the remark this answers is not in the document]
jusdino: Just did some spot checking - looks like my version of the common lambda layer with `cryptography` included is ~57 MB. Most of our lambdas that use the layer are <1 MB in size themselves. The biggest lambda I see, with `Authorize.net` as a direct dependency, is the Post Purchase handler, at 11 MB, so we've still got a lot of head room, I think.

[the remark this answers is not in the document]
landonshumway-ia: Yeah that puts us at around 23 percent, which is fine.

backend/compact-connect/lambdas/python/common/cc_common/signature_auth.py
```diff
@@ -0,0 +1,412 @@
...
+def optional_signature_auth(fn: Callable) -> Callable:
+    """
+    Decorator for optional signature validation.
+
+    This decorator checks if signature keys are configured for the compact/state combination.
+    If keys are configured and X-Key-Id is provided, it enforces signature validation.
+    If no keys are configured, it allows the request to proceed without signature validation.
+    If keys are configured but no X-Key-Id is provided, access is denied.
+
+    This is useful for endpoints that support both authenticated and unauthenticated access,
```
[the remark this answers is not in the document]
jusdino: Yeah, that's worded poorly. I'll revise.

backend/compact-connect/lambdas/python/common/cc_common/utils.py
```diff
@@ -98,19 +99,19 @@ def api_handler(fn: Callable):
...
         # Determine the appropriate CORS origin header value
-        origin = headers.get('Origin')
+        origin = event['headers'].get('Origin')
         if origin in config.allowed_origins:
             cors_origin = origin
         else:
             cors_origin = config.allowed_origins[0]
 
-        content_type = headers.get('Content-Type')
+        content_type = event['headers'].get('Content-Type')
```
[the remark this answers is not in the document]
jusdino: The difference is that we're actually replacing the `event` body's `header` field with  CaseInsensitiveDict, so all downstream consumers then benefit from the case-insensitivity and no-longer have to handle it themselves. This is also closing a bug in how I was dropping the Auth headers - I meant to be dropping them from the event before it was passed along, but instead was dropping them from the (temporary) `headers` variable, which works but is less robust than removing them from the actual event structure.

backend/compact-connect/lambdas/python/common/tests/function/test_signature_auth.py
```diff
@@ -0,0 +1,358 @@
...
+
+from common_test.sign_request import sign_request
+from moto import mock_aws
+
+from tests.function import TstFunction
+
+
+@mock_aws
+class TestSignatureAuthFunctional(TstFunction):
+    """Functional tests for signature authentication using real database interactions."""
```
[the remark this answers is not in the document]
landonshumway-ia: Makes sense. Thanks for clarifying.

backend/compact-connect/stacks/state_api_stack/v1_api/bulk_upload_url.py
```diff
@@ -31,34 +32,48 @@ def __init__(
...
     ) -> PythonFunction:
         stack: Stack = Stack.of(self.resource)
         handler = PythonFunction(
             self.api,
             'V1BulkUrlHandler',
             description='Get upload url handler',
             lambda_dir='provider-data-v1',
-            index=os.path.join('handlers', 'bulk_upload.py'),
+            index=os.path.join('handlers', 'state_api.py'),
```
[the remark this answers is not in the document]
jusdino: That is not duplicate - the internal API uses the handler defined in `bulk_upload.py`, which comes with slightly different auth requirements.

backend/compact-connect/stacks/state_api_stack/v1_api/bulk_upload_url.py
```diff
@@ -31,34 +32,48 @@ def __init__(
...
             handler='bulk_upload_url_handler',
             role=license_upload_role,
-            environment={'BULK_BUCKET_NAME': bulk_uploads_bucket.bucket_name, **stack.common_env_vars},
+            environment={
+                'BULK_BUCKET_NAME': persistent_stack.bulk_uploads_bucket.bucket_name,
+                'COMPACT_CONFIGURATION_TABLE_NAME': persistent_stack.compact_configuration_table.table_name,
+                'RATE_LIMITING_TABLE_NAME': persistent_stack.rate_limiting_table.table_name,
+                **stack.common_env_vars,
+            },
```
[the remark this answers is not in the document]
jusdino: Those are read via the `lambdas/python/common/cc_common/config.py` module.

backend/compact-connect/stacks/state_api_stack/v1_api/post_licenses.py
```diff
@@ -92,14 +98,18 @@ def _post_licenses_handler(self, license_preprocessing_queue: IQueue, license_up
...
                 **stack.common_env_vars,
             },
             alarm_topic=self.api.alarm_topic,
         )
 
         # Grant permissions to put messages on the preprocessing queue
-        license_preprocessing_queue.grant_send_messages(handler)
+        persistent_stack.ssn_table.preprocessor_queue.queue.grant_send_messages(handler)
+        persistent_stack.compact_configuration_table.grant_read_data(handler)
+        persistent_stack.rate_limiting_table.grant_read_write_data(handler)
```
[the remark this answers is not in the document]
jusdino: We can consider that as a potential future enhancement, if we desire to focus access to only nonce-related records. The policy is already pretty tailored, given that the rate limiting table is pretty use-specific and not especially sensitive.

landonshumway-ia (approved):
Pulled this down, deployed the endpoints, and successfully ran the smoke tests. Nice work 👍

jusdino:
Ready for you, @jlkravitz !

backend/compact-connect/app_clients/README.md
```diff
@@ -1,104 +1,271 @@
...
+Examples:
+- `prod-key-001`
+- `beta-key-2024-01`
+
+#### 3. Create SIGNATURE Public Key Using Interactive Python Script
+
+**Use the provided Python script in the bin directory for streamlined SIGNATURE key management:**
+
+```bash
+python3 bin/manage_signature_keys.py create -t <table_name>
```
jlkravitz, line 160: Can you clarify which table should be specified here under `<table_name>`?

backend/compact-connect/docs/client_signature_auth.md
```diff
@@ -0,0 +1,178 @@
...
+### Error Handling
+- Authentication failures will return HTTP 401 or 403
+- Check both OAuth2 token validity and signature
+- Implement retry logic with exponential backoff
+
+## Client Registration
+
+To implement signature authentication:
+
+1. **Receive client credentials and test your client credentials grant flow**
```
jlkravitz, line 131: The linked file seems to have moved.

backend/compact-connect/docs/client_signature_auth.md
```diff
@@ -0,0 +1,178 @@
...
+
+We maintain an example implementation, which we use to test and validate our own authentication mechanism
+[here](../lambdas/python/common/common_test/sign_request.py). You can use this as a reference for your own
+implementation.
+
+
+## Security Considerations
+
+### Timestamp Validation
+- Requests must be made within 1 minute of the timestamp
```
jlkravitz, line 105: Does this mean the timestamp in the request (as recorded by the client) is compared to the timestamp on our servers?

backend/compact-connect/docs/client_signature_auth.md
```diff
@@ -0,0 +1,178 @@
...
+### Private Key Security
+- Store private keys securely (HSM, key vault, encrypted storage)
+- Implement key rotation procedures
+- Never log or transmit private keys
+- Use appropriate file permissions (600) on key files
+
+### Error Handling
+- Authentication failures will return HTTP 401 or 403
+- Check both OAuth2 token validity and signature
+- Implement retry logic with exponential backoff
```
jlkravitz, line 125: I wonder if this might be better off combined with the "Common Issues" section below, making that section something that resembles more of an FAQ. I think the most useful section in this README for state IT staff will be the steps listed below (#1-4). Putting myself in their shoes, I think this section might be comparatively overwhelming and introduce a lot of additional questions.

backend/compact-connect/docs/client_signature_auth.md
```diff
@@ -0,0 +1,178 @@
...
+X-Signature: <base64_encoded_signature>
+```
+
+### 4. Example Implementation
+
+#### Python Example
+
+We maintain an example implementation, which we use to test and validate our own authentication mechanism
+[here](../lambdas/python/common/common_test/sign_request.py). You can use this as a reference for your own
+implementation.
```
jlkravitz, line 99: Thanks for including an example- really helpful, and I think will be a great starting point for state IT staff.

backend/compact-connect/lambdas/python/common/cc_common/signature_auth.py
```diff
@@ -0,0 +1,412 @@
...
+        timestamp = datetime.fromisoformat(timestamp_str)
+        if timestamp.tzinfo is None:
+            # Treat naive timestamps as UTC to avoid mismatched aware vs naive comparisons
+            timestamp = timestamp.replace(tzinfo=UTC)
+        now = config.current_standard_datetime
+        time_diff = abs((timestamp - now).total_seconds())
+
+        if time_diff > config.signature_max_clock_skew_seconds:
+            logger.warning(
+                'Request timestamp too old or in future',
```
jlkravitz, line 253: Technically, this logic allows `timestamp` to be up to 60 seconds in the future, right?

backend/compact-connect/docs/client_signature_auth.md
```diff
@@ -0,0 +1,178 @@
...
+   NONCE
+   KEY_ID
+   ```
+
+#### Canonical query string
+
+- Percent-encode keys and values per RFC 3986 (space as %20, not +; do not encode unreserved characters).
+- Sort first by key, then by value, using byte-order of the percent-encoded strings.
+- Join as `key=value` pairs with `&` as the separator.
+- Omit the entire line if there are no query parameters (leave it empty in the signature string).
```
jlkravitz, line 74: Reviewing the code, it seems like this should be an empty line, not a deleted line (which is how I initially interpreted this sentence). Can we clarify? e.g.

```suggestion
- If there are no query parameters, include an empty line in the signature string (e.g., not `''` but `'\n'`)
```

backend/compact-connect/lambdas/python/common/tests/unit/test_signature_auth_integration.py
```diff
@@ -0,0 +1,236 @@
...
+                {'message': 'Public key not found for this compact/jurisdiction/key-id'}, json.loads(resp['body'])
+            )
+
+    def test_decorator_order_matters(self):
+        """Test that decorator order affects behavior (api_handler should be outermost)."""
+        from cc_common.signature_auth import required_signature_auth
+        from cc_common.utils import api_handler
+
+        # This test demonstrates that api_handler should be the outermost decorator
+        # so it can properly handle exceptions from signature_auth_required
```
jlkravitz, line 163: Hmm, can you clarify here? Given `base_event` does not include a key ID header, the exception seems correct - despite the fact that the decorator order is wrong here.

backend/compact-connect/tests/smoke/signature_auth_smoke_tests.py
```diff
@@ -0,0 +1,479 @@
...
+        configure_signature_public_key(public_key_pem)
+
+        # Step 3: Test bulk-upload endpoint without SIGNATURE (keys configured - should fail)
+        test_bulk_upload_endpoint_without_signature_after_key_configuration(client_id, client_secret)
+
+        # Step 4: Test bulk-upload endpoint with valid SIGNATURE authentication
+        test_bulk_upload_endpoint_with_signature(client_id, client_secret, private_key_pem)
+
+        # Step 5: Test providers/query endpoint with valid SIGNATURE authentication
+        test_providers_query_endpoint_with_signature(client_id, client_secret, private_key_pem)
```
jlkravitz, line 460: If easy, can we add a quick step to test that this fails without signature auth? (i.e., between steps 1 and 2 above)

jlkravitz (commented):
Really great docs and testing- made it very clear what this is/does and how it works. Few questions/comments/notes.

backend/compact-connect/docs/client_signature_auth.md
jusdino: Yes, that's part of the authentication mechanism which, in combination with the nonce, makes replay attacks more difficult, since the signature is only good for a minute after it is generated.

backend/compact-connect/docs/client_signature_auth.md
jusdino: I'll give the whole doc a bit of a clean-up review, now that we've settled on the final details.

backend/compact-connect/lambdas/python/common/tests/unit/test_signature_auth_integration.py
jusdino: If the `api_handler` were the outer decorator, the exception would be caught and translated to a 4XX error, so this test is verifying that the exception escapes the `api_handler`, when the decorators are backwards.

jusdino:
@jlkravitz , I believe I've covered all of your comments ✔️

jlkravitz (approved):

PR #1037, opened 2025-08-26
Set cloudtrail/access logs buckets object lock retention for 90 days

### LINKED ISSUE #669: Enable access log object locks
Just before we first ingest any real provider data, we should enable Object Lock on the CloudTrail and Access Log s3 buckets, so that it is impossible to delete security logs, short of terminating the AWS logs account and waiting 90 days. We don't want to do this earlier than we have to, though, because, once that is set, _noone_ can delete any logs before the expiry age we set, which could get expensive or risky, should we log more data or more sensitive data than we intend, while development work is still in flux.

landonshumway-ia:
Now that we are getting closer to launch, we are setting a retention policy on our production access logs, as well as our cloud trail logs. This updates the CDK config to set those retention policies.

Closes #669


landonshumway-ia:
@jusdino I created a calendar event for us to meet in January 2026 to review the policies.

landonshumway-ia:

backend/multi-account/backups/stacks/backup_account_stack.py
```diff
@@ -219,6 +219,15 @@ def _create_backup_vault(self) -> None:
...
             removal_policy=RemovalPolicy.RETAIN,  # Prevents deletion - resources are retained when stack is deleted
+            lock_configuration=LockConfiguration(
+                min_retention=Duration.days(90),
+                # By setting this, the retention period can NEVER be changed after 6 months
+                # from the time this is set (it will be locked in compliance mode approximately March 2026)
+                # After that point in time, if the period needs to be changed, a new
+                # vault will need to be created.
+                # see https://docs.aws.amazon.com/aws-backup/latest/devguide/vault-lock.html
+                changeable_for=Duration.days(180),
+            ),
```
jlkravitz, line 230: Briefly- why does this vault require `changeable_for` but the others (e.g. backup_infrastructure_stack) doesn't?

jlkravitz (commented):
one small question to make sure I fully understand the implications of this!

backend/multi-account/backups/stacks/backup_account_stack.py
landonshumway-ia: It doesn't require it, we are setting it as a security measure to lock down our backups in the backup account under compliance mode, meaning that no one, not even AWS support, can delete these backups in the backup account without closing the account and waiting for 90 days. In the primary account, we are locking the backup vault under governance mode, meaning it can still be changed by an admin or someone with sufficient permissions to remove vault locks, so it's more of an additional layer to prevent accidental deletion of recovery points. The link I provided in that comment there gives more details on this. https://docs.aws.amazon.com/aws-backup/latest/devguide/vault-lock.html

PR #1080, opened 2025-09-11
Restrict deployment roles

### LINKED ISSUE #1066: Tighten cross-account, deployment roles
IAM Roles are currently used for cross-account deployment and ControlTower account management activities. To tighten the security risk of these integrations, we should review the IAM roles involved and ensure that each:
- Trusts a specific IAM role or roles in the source account or
- Requires an ExternalId for role assumption, as a second auth factor

We also should assess the scope of AWS features in our app footprint at this point and put a permissions boundary in place around the CloudFormation execution role to scope down its permissions to what it requires.

jusdino:
### Requirements List
- Scope cross-account trusts to specific source-account roles
- Add permissions boundary to CloudFormation execution role
- Deployment pipeline remains functional
- Docs reflect workflow changes

### Description List
- Updated pipelines to use single predictably-named roles for trusts
- Updated CDK bootstrap process to deploy custom stack, trusting only named pipeline roles
- Updated docs to describe deployment process in dependency order (pipelines before application accounts)
- Added permissions boundary policy to CloudFormation execution role

### Testing List
- cdk tests inspect new pipeline roles
- manually tested cross account deployment
- manually inspected resulting bootstrap role trust policies
- Code review

Closes #1066 

## Summary by CodeRabbit

- New Features
  - Predictable, per-environment pipeline roles; pipelines now use dedicated cross-account roles.
  - Environment-specific bootstrap stacks for Test, Beta, Prod with secure KMS, S3, ECR, IAM, and exports.

- Bug Fixes
  - Removed legacy cross-stack exports to simplify deployment boundaries.

- Documentation
  - Multi-account guide updated: deploy pipeline stacks first; added custom bootstrap stack guidance and explicit environment-specific bootstrap commands.

- Tests
  - Added tests validating predictable role names, trust policies, and pipeline role usage.

- Chores
  - Ignored Kiro artifacts.
  - Default web ACL rate limit increased (100 → 1000 requests per 5 minutes per IP).

### DIFF
--- a/backend/compact-connect/README.md
+++ b/backend/compact-connect/README.md
@@ -187,8 +187,9 @@ The pipelined environments leverage a custom bootstrap stack, which includes cro
 as well as a permissions boundary around the CloudFormation execution role. If new AWS services are added to the app
 architecture, that permissions boundary will need to be updated to allow access to the new service. See the
 [multi-account documentation](../multi-account/README.md#bootstrap-the-application-accounts) for details on how to
-deploy the custom bootstrap stack. The custom stack can also be deployed to a sandbox account for testing, using the
-same steps.
+deploy the custom bootstrap stack. If you want to test the bootstrap stack customizations in your sandbox, for example,
+to make sure the new resources you are creating in your sandbox won't be blocked by the CloudFormation execution role's
+permission boundary, you can deploy the custom stack to a sandbox account for testing, using the same steps.
 
 ### Verifying SES configuration for Cognito User Notifications
 If your account is in the SES sandbox, the simplest way to verify that SES is integrated with your cognito user pool is
--- a/backend/compact-connect/pipeline/__init__.py
+++ b/backend/compact-connect/pipeline/__init__.py
@@ -166,7 +166,6 @@ class BasePipelineStack(Stack):
         """Create a predictable cross-account role that will be trusted by bootstrap roles.
 
         :param pipeline_type: 'Backend' or 'Frontend'
-        :param role_type: 'Deploy' (for construct naming)
         :return: The cross-account role with predictable name for bootstrap trust policies
         """
         # Create environment and pipeline-type specific cross-account roles


backend/compact-connect/pipeline/__init__.py
```diff
@@ -150,6 +150,37 @@ def __init__(
...
+        if self.environment_name not in ALLOWED_ENVIRONMENT_NAMES:
+            raise ValueError(f'Environment name must be one of {ALLOWED_ENVIRONMENT_NAMES}')
+
+        return f'CompactConnect-{self.environment_name}-{pipeline_type}-{role_type}Role'
+
+    def create_predictable_pipeline_role(self, pipeline_type: str) -> Role:
+        """Create a predictable cross-account role that will be trusted by bootstrap roles.
+
+        :param pipeline_type: 'Backend' or 'Frontend'
+        :param role_type: 'Deploy' (for construct naming)
```
jlkravitz, line 169: looks like this isn't a parameter to this method?

backend/compact-connect/pipeline/frontend_pipeline.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
jlkravitz, line 1: Is this necessary for our version of python?

backend/compact-connect/pipeline/backend_pipeline.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
jlkravitz, line 1: Is this necessary for our version of python?

backend/compact-connect/stacks/api_stack/v1_api/api.py
```diff
@@ -10,17 +10,17 @@
 
 from stacks import persistent_stack as ps
 from stacks.api_lambda_stack import ApiLambdaStack
```
jlkravitz, line 12: nit: It seems like we still import these via absolute imports - is that intentional? Just out of curiosity, what's our convention for using relative vs absolute imports?

backend/compact-connect/README.md
```diff
@@ -180,6 +181,15 @@ to reflect the changes in your code. Full deployment steps are:
...
 4) Run `cdk deploy 'Sandbox/*'` to deploy the app to your AWS account.
 
+### Custom bootstrap stack
+
+The pipelined environments leverage a custom bootstrap stack, which includes cross-account trusts to the deploy account
+as well as a permissions boundary around the CloudFormation execution role. If new AWS services are added to the app
+architecture, that permissions boundary will need to be updated to allow access to the new service. See the
+[multi-account documentation](../multi-account/README.md#bootstrap-the-application-accounts) for details on how to
+deploy the custom bootstrap stack. The custom stack can also be deployed to a sandbox account for testing, using the
+same steps.
```
jlkravitz, line 191: Perhaps my naivety, but can we clarify the implications of this for local development? My initial read is that everything will work as-is without this custom bootstrap stack, and that this stack basically provides the necessary permissions for auto-deploy (via github), backups, and other CI/CD related things to work --- which aren't necessarily required for getting a basic version of the app running?

jlkravitz (commented):
looks good- few very small questions & comments. (mostly for my understanding!)

backend/compact-connect/pipeline/frontend_pipeline.py
jusdino: They'll be necessary until we upgrade to Python 3.14

backend/compact-connect/pipeline/backend_pipeline.py
jusdino: until 3.14

backend/compact-connect/stacks/api_stack/v1_api/api.py
jusdino: It isn't a hard convention, so much as a rough guess:
- If a refactor is likely to have the importer and the imported module moved together, so their relative location is likely to stay the same, relative import
- If a refactor is likely to have the importer and the imported module moved separately, so their relative location is different, absolute import

In this case, the API lambda stack's location isn't tightly tied to the `v1_api1 or `api_stack` package, so I think an absolute import makes sense.

backend/compact-connect/README.md
jusdino: Your initial read is correct. The custom stack is unnecessary for sandboxes, since cross-account deploys aren't at hing. But if you _want_ to test the permissions boundary, for example, you _can_ deploy the custom bootstrap stack to your sandbox to make sure CloudFormation can still deploy your changes. That gives you the option to fail earlier than when your code is merged and headed down the pipeline.

jlkravitz (approved):

jlkravitz:
oops, sorry! @isabeleliassen done

PR #1110, opened 2025-09-24
Set up skeleton for feature flag directory

landonshumway-ia:
### Requirements List
-

### Description List
-

### Testing List
- Code review
- ???

Closes #


backend/compact-connect/lambdas/python/feature-flag/handlers/check_feature_flag.py
```diff
@@ -0,0 +1,47 @@
...
+            logger.warning('Feature flag validation failed', error=str(e))
+            raise CCInvalidRequestException(str(e)) from e
+
+        # Create request object for flag evaluation
+        flag_request = FeatureFlagRequest(**validated_body)
+
+        # Check the feature flag
+        result = feature_flag_client.check_flag(flag_request)
+
+        logger.debug('Feature flag checked', flag_name=validated_body['flagName'], enabled=result.enabled)
```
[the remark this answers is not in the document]
landonshumway-ia: I have verified through extensive API calls that this is not true. This handles the case where the caller does not provide a context object it parses the body and forwards the request without error.

backend/compact-connect/lambdas/python/feature-flag/feature_flag_client.py
```diff
@@ -0,0 +1,629 @@
...
+            # Create new gate with environment-specific rule
+            return self._create_new_gate(flag_name, auto_enable, custom_attributes)
+
+        # Gate exists - check if environment rule exists
+        gate_id = existing_gate.get('id')
+        rule_name = f'{self.environment.lower()}-rule'
+        environment_rule = self._find_environment_rule(existing_gate, rule_name)
+
+        # we only set the environment rule if it doesn't already exist
+        # else we leave it alone to avoid overwriting manual changes
```
jusdino, line 380: Very helpful comment. Thank you.

backend/compact-connect/lambdas/python/feature-flag/feature_flag_client.py
```diff
@@ -0,0 +1,629 @@
...
+        if not environment_rule:
+            updated_gate = self._add_environment_rule(existing_gate, auto_enable, custom_attributes)
```
jusdino, line 382: It took me a minute to understand what this was doing here, adding an environment rule, then updating the gate. I assumed, based on the `_add_environment_rule` name, that the gate would have been updated before the `_update_gate` call. Maybe call it `_add_environment_rule_to_gate_data` or something, for clarity? Or maybe actually call `self._update_gate()` from `_add_environment_rule()`?

backend/compact-connect/stacks/api_lambda_stack/__init__.py
```diff
@@ -29,6 +30,17 @@ def __init__(
             **kwargs,
         )
 
+        # we only pass the API_BASE_URL env var if the API_DOMAIN_NAME is set
+        # this is because the API_BASE_URL is used by the feature flag client to call the flag check endpoint
```
jusdino, line 34: So, do the feature flags only work when we have a domain name defined, then? Sandboxes with no domain name will just use `fail_default`?

backend/compact-connect/stacks/api_lambda_stack/feature_flags.py
```diff
@@ -0,0 +1,64 @@
...
+        environment_name = self.stack.common_env_vars['ENVIRONMENT_NAME']
+        self.statsig_secret = Secret.from_secret_name_v2(
+            self.scope,
+            'StatsigSecret',
+            f'compact-connect/env/{environment_name}/statsig/credentials',
+        )
+
+        self.check_feature_flag_function = self._create_check_feature_flag_function(lambda_environment)
+
+    def _create_check_feature_flag_function(self, lambda_environment: dict) -> PythonFunction:
```
jusdino, line 38: The function's log group should be added to our saved query, right?

backend/compact-connect/stacks/feature_flag_stack/feature_flag_resource.py
```diff
@@ -0,0 +1,168 @@
...
+            lambda_dir='feature-flag',
+            handler='on_event',
+            log_retention=RetentionDays.ONE_MONTH,
+            environment={'ENVIRONMENT_NAME': environment_name},
+            timeout=Duration.minutes(5),
+            memory_size=256,
+        )
+
+        # Grant permissions to read secrets
+        secret_name = f'compact-connect/env/{environment_name}/statsig/credentials'
```
jusdino, line 75: Are we manually creating this or did I miss its declaration, somewhere?

backend/compact-connect/stacks/feature_flag_stack/feature_flag_resource.py
```diff
@@ -0,0 +1,168 @@
...
+        self.manage_function.add_to_role_policy(
```
jusdino, line 76: Did [Secret.from_secret_name_v1()](https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_secretsmanager/Secret.html#aws_cdk.aws_secretsmanager.Secret.from_secret_name_v2) not work for this?

backend/compact-connect/stacks/feature_flag_stack/feature_flag_resource.py
```diff
@@ -0,0 +1,168 @@
...
+            PolicyStatement(
+                actions=['secretsmanager:GetSecretValue'],
+                resources=[
+                    f'arn:aws:secretsmanager:{Stack.of(self).region}:{Stack.of(self).account}:secret:{secret_name}-*'
+                ],
+            )
+        )
+
+        # Create the custom resource provider
+        self.provider = Provider(
```
jusdino, line 86: This layout of one flag == one lambda == one provider isn't necessary for this stack. It's going to raise resource counts and deploy times. The function logic is not resource-specific, so you can instead declare one provider and one function for the whole stack, then reference those for each flag. You can still even declare the function and provider within this construct, if you just set its scope to the stack and use `try_find_child` to see if the resource already exists on the stack before constructing a new one.

backend/compact-connect/stacks/feature_flag_stack/feature_flag_resource.py
landonshumway-ia: Our secrets are manually created, since we can't check the values into the project

backend/compact-connect/lambdas/python/common/cc_common/feature_flag_client.py
```diff
@@ -0,0 +1,124 @@
...
+    try:
+        api_base_url = _get_api_base_url()
+        endpoint_url = f'{api_base_url}/v1/flags/check'
+
+        # Build request payload
+        payload = {'flagName': flag_name}
+        if context:
+            payload['context'] = context.to_dict()
+
+        response = requests.post(
```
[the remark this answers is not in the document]
landonshumway-ia: I debated on that, however, this is a query endpoint with a nested request body to include the custom attributes, which doesn't play nicely with query params, so I opted for a POST like we're doing with the query providers endpoint.

backend/compact-connect/stacks/api_lambda_stack/feature_flags.py
landonshumway-ia: I've now added it in the api stack where we define the API endpoint.

backend/compact-connect/lambdas/python/feature-flag/feature_flag_client.py
```diff
@@ -0,0 +1,629 @@
...
+
+        url = f'{STATSIG_API_BASE_URL}{endpoint}'
+        headers = {
+            'STATSIG-API-KEY': self._console_api_key,
+            'STATSIG-API-VERSION': STATSIG_API_VERSION,
+            'Content-Type': 'application/json',
+        }
+
+        try:
+            if method == 'GET':
```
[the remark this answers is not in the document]
landonshumway-ia: I have updated the methods to use the `json` parameter instead of `data`. As far as using `requests.request`, at this point the tests are all based on checking the mocks of specific method calls, so if there is not a strong preference I would prefer keeping that aspect as is. But we can change it if you feel it needs to be updated.

backend/compact-connect/stacks/api_stack/v1_api/feature_flags.py
```diff
@@ -0,0 +1,58 @@
...
+            'POST',
+            integration=LambdaIntegration(api_lambda_stack.feature_flags_lambdas.check_feature_flag_function),
+            request_models={'application/json': api_model.check_feature_flag_request_model},
+            method_responses=[
+                {
+                    'statusCode': '200',
+                    'responseModels': {'application/json': api_model.check_feature_flag_response_model},
+                },
+            ],
+        )
```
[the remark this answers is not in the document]
landonshumway-ia: Please check again as this is the pattern we use in our other API Gateway resource method definitions

backend/compact-connect/README.md
```diff
@@ -286,6 +288,41 @@ Once the pipelines are established with the above steps, deployments will be aut
...
+     ```
+     compact-connect/env/{environment_name}/statsig/credentials
+     ```
+   - The secret value should be a JSON object with the following structure:
+     ```json
+     {
+       "serverKey": "<your_server_secret_key>",
+       "consoleKey": "<your_console_api_key>"
+     }
+     ```
```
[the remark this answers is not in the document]
landonshumway-ia: this is opened with ```json, is it not?

backend/compact-connect/lambdas/nodejs/tests/lib/email/ingest-event-email-service.test.ts
```diff
@@ -97,19 +97,21 @@ describe('IngestEventEmailService', () => {
...
+                        Body: {
+                            Html: {
+                                Charset: 'UTF-8',
+                                Data: expect.stringContaining('<!DOCTYPE html>')
+                            }
+                        },
+                        Subject: {
                             Charset: 'UTF-8',
-                            Data: expect.stringContaining('<!DOCTYPE html>')
+                            Data: 'License Data Error Summary: aslp / Ohio'
```
jusdino, line 110: We should update our mock data to reflect that this now shows the full compact name instead of the slug.

backend/compact-connect/lambdas/nodejs/package.json
```diff
@@ -13,6 +13,10 @@
   },
   "author": "Inspiring Apps",
   "license": "UNLICENSED",
+  "resolutions": {
```
jusdino, line 16: What are these for?

jusdino (changes requested):
Thanks for handling this. Two quick comments:

backend/compact-connect/lambdas/nodejs/tests/lib/email/ingest-event-email-service.test.ts
landonshumway-ia: done

backend/compact-connect/lambdas/nodejs/package.json
landonshumway-ia: This was an attempt to resolve some type issues I was having with introducing the new SESv2 client. It turns out the issue was my AWS SDK deps were using different versions which was causing issues. I've since synced those and removed this section now.

jusdino (approved):
Great!

backend/compact-connect/lambdas/python/feature-flag/feature_flag_client.py
```diff
@@ -0,0 +1,618 @@
...
+            self.statsig_client.initialize().wait()
+            self._is_initialized = True
+
+        except Exception as e:
+            raise FeatureFlagException(f'Failed to initialize StatSig client: {e}') from e
+
+    def _create_statsig_user(self, context: dict[str, Any]) -> StatsigUser:
+        """Convert context dictionary to StatsigUser"""
+        user_data = {
+            'user_id': context.get('userId') or 'default_cc_user',
```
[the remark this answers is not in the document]
jlkravitz: Update: A little research indicates that this is used to provide a consistent user experience (e.g., if 50% of users have the flag on, the same user ID will always get the same result). I guess I'm a little less sure of the impact of this `default_cc_user` user_id - perhaps a short comment can explain? Or a link to relevant statsig docs?

backend/compact-connect/stacks/feature_flag_stack/__init__.py
```diff
@@ -0,0 +1,206 @@
...
+- beta → staging (StatSig tier)
+- prod → production (StatSig tier)
+- sandbox/other → development (StatSig tier, default)
+
+
+Checking Flags in Lambda:
+-------------------------
+There is a common feature flag client python module that can be used to check if a flag is enabled in StatSig
+
+```python
```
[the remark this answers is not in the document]
landonshumway-ia: The core benefit to putting the feature flag check behind an API endpoint is it becomes callable from outside our python lambdas. There may be cases where we want to check a flag using this API endpoint from the frontend, or from another AWS service that supports HTTP hooks, or even our email service lambda (which is node based). Making it an REST endpoint allows a flag to be checked anywhere in the system regardless of the underlying language/service.

Could we enhance the python common library to create a wrapper to call StatSig directly? Absolutely, but we wanted to build out the generic common solution first that would work for any case that support HTTP calls. And given how large this PR is already, adding a common wrapper for the StatSig Client would fit better as a future enhancement 😄

PR #1157, opened 2025-10-14
Toggle srp auth for zap

jusdino:
Toggle SRP auth back on for the SecurityProfile.VULNERABLE, to enable our ZAP scan automation in test

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the UI sign-in flow to align with the selected security profile. The Recommended profile keeps the simpler password-based sign-in, while stricter profiles now enable an enhanced, challenge-based authentication method. This ensures the login experience and security level are consistent with your chosen profile without impacting other app behaviors.

### DIFF
--- a/backend/compact-connect/common_constructs/user_pool.py
+++ b/backend/compact-connect/common_constructs/user_pool.py
@@ -196,7 +196,7 @@ class UserPool(CdkUserPool):
                 # (and automated testing in test environments)
                 admin_user_password=True,
                 custom=False,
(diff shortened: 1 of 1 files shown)


landonshumway-ia (approved):
Approved, just note that if we ever perform ZAP scans against beta this won't work but that's not the case for now.

jlkravitz (approved):
Looks good, thanks for handling. @isabeleliassen This is good to merge!

PR #1168, opened 2025-10-21
Frontend/registration states update

### LINKED ISSUE #1054: only show live states in registration FE
Just realized that the frontend implementation for this item may be a little trickier than we initially thought. On the registration page, the list of "live" states will depend on which profession the licensee actually picks. For this reason, we need to swap the license type and home state drop downs on the registration page. We also may need to somehow restrict the home state drop down to be something they can't click until they select the profession.

Or, a different (and potentially better?) solution to this would be to allow all states to remain in the list, but if they select a state that isn't live, we show them an error message stating that that state is not yet live (exact wording tbd) and we do not allow them to proceed with the registration

jsandoval81:
### Requirements List
- _None_

### Description List
- Updated the Card component to allow overflow as a config setting
- Created a new InputLabel form component to handle the growing complexity of the form label UI
- Included optional info icon capability in the new InputLabel component
- Updated the FormInput model with a new `labelInfo` property
- Use the new InputLabel within the InputSelect component
- Add the new info icon to the State input on the registration page
- Move the License Type and State inputs to the top of the registration form
- Update the models / store / network layers to fetch the compact live states from the server
- Make the State options list dependent on the selected License Type
- Show which State options are live / not-live
- Update registration Next button to be full width to match updated designs

### Testing List
- Code review
- Testing
    - Navigate to the registration page
    - Confirm the License Type & State inputs are now the first 2 fields
    - Confirm the State input is not enabled until a License Type is selected
    - Confirm the State input values are updated when the License Type is changed
    - Confirm the State input now has an info icon that can be toggled & matches the designs
    - Confirm the registration flow can be successfully completed as usual
    - Generally smoke test around the app that select dropdown inputs do not have any regressions

Closes #1054 

## Summary by CodeRabbit

* **New Features**
  * New InputLabel with expandable info panels (keyboard/ARIA accessible).
  * Licensee registration shows live/disabled state availability with dynamic state selection.
  * Card components can opt into allowing overflow for layout flexibility.

* **Localization**
  * Added translations for state availability and info-toggle labels/messages.

* **Tests**
  * Added unit tests for the new label component and registration compact-state flows.

### DIFF
--- a/webroot/src/components/Forms/InputLabel/InputLabel.less
+++ b/webroot/src/components/Forms/InputLabel/InputLabel.less
@@ -89,9 +89,9 @@
                 width: 26rem;
                 padding: 1.4rem;
                 border-radius: 20px;
-                color: white;
+                color: @white;
                 font-size: @fontSizeSmaller;
-                background-color: @darkBlue;
+                background-color: @fontColor;
 
                 @media @tabletWidth {
                     width: 30rem;
--- a/webroot/src/pages/RegisterLicensee/RegisterLicensee.less
+++ b/webroot/src/pages/RegisterLicensee/RegisterLicensee.less
@@ -93,9 +93,13 @@
     .register-licensee-form-container {
         width: 100%;
 
-        &:deep(.register-licensee-continue-to-confirmation-button) {
+        .register-licensee-continue-to-confirmation-button {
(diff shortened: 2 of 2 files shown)


jsandoval81:

PR #1187, opened 2025-10-30
Feat/prevent repeated ssns

### LINKED ISSUE #1181: Prevent license CSV batch uploads that repeat SSNs
We had an issue in beta where a state uploaded many license records, all under the same SSN. This created thousands of license update records under the same licensee, which broken several parts of the system as it was not designed to handle this case of the same licensee being updated thousands of time with a single license upload.

We can prevent this from occurring in our CSV uploads by checking every SSN in the file and not processing any of the batch if duplicates are detected. We will also update our technical documentation to notify states that we do not accept repeated SSNs within the same CSV file.

landonshumway-ia:
We had an issue in beta where a state uploaded many license records, all under the same SSN. This created thousands of license update records under the same licensee, which broken several parts of the system as it was not designed to handle this case of the same licensee being updated thousands of time with a single license upload.

To prevent this from occurring in the future, this adds additional logic in our CSV uploads by checking for duplicates SSN in the file as we process them and not processing any records that duplicate SSN that were already processed. In addition, the JSON API has been updated to check all the records in the request for any duplicates, and rejecting the entire request.

We will also update our technical documentation to notify states that we do not accept repeated SSNs within the same CSV file.

Closes #1181

## Summary by CodeRabbit

* **Documentation**
  * Clarified upload guidance: SSNs must be unique within a single CSV file upload (duplicates in-file are rejected after the first occurrence) and within a single JSON request payload.

* **New Features**
  * Duplicate-SSN detection: uploads containing the same SSN more than once in the same file or request are rejected with a validation error/400 response. The check can be toggled via a feature flag.

### DIFF
--- a/backend/compact-connect/docs/it_staff_onboarding_instructions.md
+++ b/backend/compact-connect/docs/it_staff_onboarding_instructions.md
@@ -240,7 +240,7 @@ For your convenience, use of this feature is included in the [Postman Collection
 - `licenseType` must match exactly with one of the valid types for the specified compact
 - All date fields must use the `YYYY-MM-DD` format
 - The API does not accept `null` values. For optional fields with no value, omit the field or leave it empty in CSV.
-- For CSV uploads, SSNs must be unique within a single file. Do not include multiple rows with the same `ssn` in one upload. If duplicate SSNs are sent within the same file, the first row to be processed, but all other duplicate rows will be rejected.
+- For CSV uploads, SSNs must be unique within a single file. Do not include multiple rows with the same `ssn` in one upload. If duplicate SSNs are sent within the same file, the first row will be processed, but all other duplicate rows will be rejected.
 - For JSON uploads, SSNs must be unique within a single request payload (array). Do not include duplicate `ssn` values in the same batch. Attempting to do so will cause the entire request to be rejected.
 
 ## Common Upload Strategies: JSON vs CSV
--- a/backend/compact-connect/docs/README.md
+++ b/backend/compact-connect/docs/README.md
@@ -31,7 +31,7 @@ Export your license data to a CSV file, formatted as follows:
    - String lengths are enforced - exceeding them will cause validation errors
    - Some fields have a set list of allowed values. For those fields, make sure to enter the value exactly, including
      spacing and capitalization
-   - SSNs must be unique within a single CSV upload file. Do not include multiple rows with the same `ssn` in one file. If duplicate SSNs are sent within the same file, the first row to be processed, but all other duplicate rows will be rejected.
+   - SSNs must be unique within a single CSV upload file. Do not include multiple rows with the same `ssn` in one file. If duplicate SSNs are sent within the same file, the first row will be processed, but all other duplicate rows will be rejected.
 
 #### Field Descriptions
 


backend/compact-connect/lambdas/python/provider-data-v1/handlers/bulk_upload.py
```diff
@@ -13,6 +13,9 @@
 )
 from cc_common.event_batch_writer import EventBatchWriter
 from cc_common.exceptions import CCInternalException
+
+# initialize flag outside of handler so the flag is cached for the lifecycle of the container
```
jlkravitz: execution environment?

backend/compact-connect/lambdas/python/provider-data-v1/handlers/bulk_upload.py
```diff
@@ -148,6 +158,17 @@ def process_bulk_upload_file(
...
+                    # verify that this ssn has not been used previously in the same batch
+                    license_ssn = validated_license['ssn']
+                    if duplicate_ssn_check_flag_enabled:
+                        matched_ssn_index = ssns_in_file_upload.get(license_ssn)
+                        if matched_ssn_index:
+                                raise ValidationError(
+                                    message=f'Duplicate License SSN detected. SSN matches with record '
+                                            f'{matched_ssn_index}. Every record must have a unique SSN within the same '
+                                            f'file.'
+                                )
```
jlkravitz: Did we decide to not go with the "first record wins" approach?

backend/compact-connect/docs/it_staff_onboarding_instructions.md
```diff
@@ -240,6 +240,8 @@ For your convenience, use of this feature is included in the [Postman Collection
 - `licenseType` must match exactly with one of the valid types for the specified compact
 - All date fields must use the `YYYY-MM-DD` format
 - The API does not accept `null` values. For optional fields with no value, omit the field or leave it empty in CSV.
+- For CSV uploads, SSNs must be unique within a single file. Do not include multiple rows with the same `ssn` in one upload
+- For JSON uploads, SSNs must be unique within a single request payload (array). Do not include duplicate `ssn` values in the same batch
```
jlkravitz: Should we specify what happens when this is the case? Does first row win, or is the request/file rejected?

jlkravitz (approved):

backend/compact-connect/lambdas/python/provider-data-v1/handlers/bulk_upload.py
landonshumway-ia: @jlkravitz We are propagating the error messages. The validation error messages get placed in [our ingest event messages here](https://github.com/csg-org/CompactConnect/blob/b51f243e891aeccf7e18a3babc92be965ec1be58/backend/compact-connect/lambdas/python/provider-data-v1/handlers/bulk_upload.py#L190) which then get sent out in the email address we send to states.

But I may be misunderstanding your suggestion here. Are you suggesting that instead of a try/catch we somehow validate the record in a way that doesn't throw an exception, detect if there are errors in the validation result, and place it on the event bus directly that way? I'm always up for making the code easier to understand and there may be something we can do there, but I agree that would be a little out of scope for this ticket.

PR #1188, opened 2025-10-31
Support license upload reverts

### LINKED ISSUE #1175: undo state upload in case of errors (security audit issue)

landonshumway-ia:
In the event that a state inadvertently uploads invalid data into the system for a compact, we need a way to effectively 'undo' that upload so that license data within the system can be reverted to its state before the upload. This PR includes the following enhancements to the system:

## Move license/privilege update records under separate sort key to decouple updates from other main records

To reduce risk that massive invalid updates will cause the system to crash when loading provider data, we have migrated the sort keys of our update records to follow a tier based pattern, which will allow us to query for update records only as needed. This requires some more complexity in the query patterns used to grab provider records, but most of this complexity is abstracted by a small number of methods.

## Add licenseUploadDate GSI
We've added a GSI to the provider table to track when uploads occurred for every license update record, and we will begin tracking the first time a license record was uploaded into the system. The PK for this new GSI includes the compact, jurisdiction, and upload month. The SK includes the epoch timestamp of the time the record was uploaded/updated, the license type, and the provider id.
```
licenseUploadDateGSIPK = f'C#{in_data["compact"].lower()}#J#{in_data["jurisdiction"].lower()}#D#{YYYY-MM}'
licenseUploadDateGSISK = f'TIME#{upload_epoch_time}#LT#{in_data["licenseType"]}#PID#{in_data["providerId"]}'
```
These **OPTIONAL** GSI fields will be added to all license and license update records moving forward for every ingest event.

## Add Step Function for reverting license uploads
We've added a step function that takes in the compact, jurisdiction, and time window (start date time and end date time).

The step function looks up all licenses that were uploaded for that compact/jurisdiction between the timestamps using the new GSI, and reverts the license record to the latest state prior to the rollback time if there were not other updates made to that provider since the upload time period. When completed it generates a JSON report stored in S3 which includes details about which licenses were reverted and which providers will require manual review due to detected updates not related to the license upload.

These enhancements will protect the system from accidental uploads and provide a process for reverting such incidents.

### Requirements List
- A migration script has been added to convert all existing update records to the new sort key pattern. This migration path is **fully** backwards compatible, meaning the system will continue to function before, during, and after the cutover without downtime.

Closes #1175 

## Summary by CodeRabbit

* **New Features**
  * Disaster recovery: license-upload rollback workflow with persisted results and revert events.
  * Tiered update model and migration tooling for richer historical queries.
  * License-upload date tracking and index to speed provider queries.

* **Bug Fixes**
  * SSN uniqueness enforcement now scoped per license type in batch uploads.

* **Documentation**
  * Added comprehensive disaster-recovery guides and rollback usage docs.

### DIFF
--- a/backend/compact-connect/lambdas/python/disaster-recovery/tests/function/test_rollback_license_upload.py
+++ b/backend/compact-connect/lambdas/python/disaster-recovery/tests/function/test_rollback_license_upload.py
@@ -193,7 +193,7 @@ class TestRollbackLicenseUpload(TstFunction):
             }
         )
 
-        # old update record before upload window (e.g., RENEWAL)
+        # old update record before upload window
         existing_update = self.test_data_generator.put_default_license_update_record_in_provider_table(
             {
                 'providerId': provider_id,
@@ -1146,7 +1146,6 @@ class TestRollbackLicenseUpload(TstFunction):
         final_results_data = self._perform_rollback_and_get_s3_object()
 
         # Verify: All existing data is preserved and new data is appended
-        # Note: All keys should now be camelCase for consistency
         self.assertEqual(
             {
                 'executionName': MOCK_EXECUTION_NAME,
--- a/backend/compact-connect/disaster_recovery/LICENSE_UPLOAD_ROLLBACK.md
+++ b/backend/compact-connect/disaster_recovery/LICENSE_UPLOAD_ROLLBACK.md
@@ -15,19 +15,20 @@ Before starting the rollback:
 1. ✅ **Verify the Problem**: Confirm which jurisdiction uploaded bad data for which compact(s)
 2. ✅ **Disable automated access for Jurisdiction**: If jurisdiction has API credentials for automated uploads, disable those credentials to prevent further data changes until system has been recovered. To do this, determine which Cognito app client(s) the jurisdiction is using for the compact(s) and delete the appropriate app client(s) from the State Auth Cognito user pool.
 3. ✅ **Determine Time Window**: Identify the exact start and end times (UTC) of the problematic uploads
-4. ✅ **Stakeholder Notification**: Coordinate with relevant state administrators and other stakeholders
+4. ✅ **Determine When Rollback Should be Performed**: Depending on the severity of the issue and scale of records that need to be rolled back, determine if the rollback needs to be performed as soon as possible or if it can be performed outside of peak traffic hours. When possible, it is recommended to perform rollbacks during periods of low traffic. While the risk is low, there is a narrow race condition (.2 second window based on load testing) where a license record may be modified by another part of the system after the rollback system checked for updates and the modification could be removed by the rollback. Running the rollback when traffic is low reduces this risk even further.  
+5. ✅ **Stakeholder Notification**: Coordinate with relevant state administrators and other stakeholders. Ensure jurisdiction is aware they should not attempt to upload any more license data until the rollback has been completed.
 
 ### Step 1: Gather Required Information
 
 You'll need the following information for the execution:
 
-| Parameter | Description | Example |
-|-----------|-------------|---------|
-| `compact` | The compact abbreviation (lowercase) | `"aslp"`, `"octp"`, `"counseling"` |
-| `jurisdiction` | The state/jurisdiction code (lowercase) | `"oh"`, `"ky"`, `"ne"` |
-| `startDateTime` | UTC timestamp when problematic uploads began | `"2020-01-15T08:00:00Z"` |
-| `endDateTime` | UTC timestamp when problematic uploads ended | `"2020-01-15T17:59:59Z"` |
-| `rollbackReason` | Description for audit trail | `"Invalid license data uploaded by OH staff"` |
+| Parameter | Description                                              | Example |
+|-----------|----------------------------------------------------------|---------|
+| `compact` | The compact abbreviation (lowercase)                     | `"aslp"`, `"octp"`, `"counseling"` |
+| `jurisdiction` | The state/jurisdiction code (lowercase)                  | `"oh"`, `"ky"`, `"ne"` |
+| `startDateTime` | UTC timestamp when problematic uploads began (inclusive) | `"2020-01-15T08:00:00Z"` |
+| `endDateTime` | UTC timestamp when problematic uploads ended (inclusive) | `"2020-01-15T17:59:59Z"` |
+| `rollbackReason` | Description for audit trail                              | `"Invalid license data uploaded by OH staff"` |
 
 **Important Notes:**
 - All timestamps must be in UTC
@@ -188,3 +189,5 @@ These events include:
 - The rollback reason
 - Time window information
 - Revision IDs for tracking
+
+These events purely for auditing purposes. They are not currently referenced by any downstream processes.
--- a/backend/compact-connect/lambdas/python/common/cc_common/data_model/update_tier_enum.py
+++ b/backend/compact-connect/lambdas/python/common/cc_common/data_model/update_tier_enum.py
@@ -5,8 +5,19 @@ class UpdateTierEnum(StrEnum):
     """
     Enum for update record tiers in the sort key hierarchy.
 
+    DynamoDB sort keys are treated as numeric values, even if the key is a string.
+    This means we can perform comparison operations on string sort keys, such as less than (lt)
+    and grab records within a certain range.
+
+    To reduce risk that massive invalid updates from a jurisdiction will cause the system to crash
+    when loading provider data, we migrated the sort keys of our update records to follow this
+    tier based pattern, which will allow us to query for update records only as needed.
+
     Update records are organized into tiers to enable efficient range queries.
-    Using lt (less than) conditions, we can fetch multiple tiers in a single query.
+    Because all the primary provider records are prefixed under a common `{compact}#PROVIDER` prefix,
+    which is lexicographically less than the `{compact}#UPDATE` prefix, using the lt condition with the
+    UPDATE prefix will grab all the update records up to the specified tier and all primary records under
+    the PROVIDER prefix.
 
     Tier structure in sort keys:
     - Tier 1: {compact}#UPDATE#1#privilege/... (Privilege updates)
@@ -22,9 +33,6 @@ class UpdateTierEnum(StrEnum):
 
     - TIER_THREE: Fetches all updates (privilege + provider + license)
         Query: Key('sk').lt('{compact}#UPDATE#4')
-
-    This tiered approach prevents bulk invalid license updates from breaking
-    queries that only need privilege and provider data.
     """
 
     TIER_ONE = '1'  # Privilege updates only


backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
```diff
@@ -179,9 +180,22 @@ def get_provider_user_records(
...
+        # When include_update_tier is set, use lt (less than) to get main records plus updates up to that tier
+        if include_update_tier is None:
+            # Get only main records: {compact}#PROVIDER prefix
+            sk_condition = Key('sk').begins_with(f'{compact}#PROVIDER')
+        else:
+            # Get main records and updates up to specified tier using lt (less than)
+            # This fetches all SKs less than {compact}#UPDATE#{next_tier}
+            next_tier = int(include_update_tier) + 1
+            sk_condition = Key('sk').lt(f'{compact}#UPDATE#{next_tier}')
+
```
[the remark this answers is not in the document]
landonshumway-ia: This is a string enum, which supports operations like this. The code currently works as designed

backend/compact-connect/tests/smoke/rollback_license_upload_smoke_tests.py
```diff
@@ -0,0 +1,796 @@
...
+        'type': 'privilege',
+        'providerId': provider_id,
+        'compact': compact,
+        'jurisdiction': privilege_jurisdiction,
+        'licenseJurisdiction': JURISDICTION,
+        'licenseType': LICENSE_TYPE,
+        'dateOfIssuance': datetime.now(tz=UTC).isoformat(),
+        'dateOfRenewal': datetime.now(tz=UTC).isoformat(),
+        'dateOfExpiration': date(2050, 12, 10).isoformat(),
+        'dateOfUpdate': datetime.now(tz=UTC).isoformat(),
```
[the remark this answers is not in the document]
landonshumway-ia: privileges are different than licenses. They use datetimes for their time fields. This is correct.

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
```diff
@@ -179,9 +180,22 @@ def get_provider_user_records(
...
         provider_id: UUID,
         consistent_read: bool = True,
+        include_update_tier: UpdateTierEnum | None = None,
     ) -> ProviderUserRecords:
         logger.info('Getting provider')
 
+        # Determine SK condition based on include_update_tier parameter
+        # When include_update_tier=None, use begins_with to get only main records (provider, licenses, privileges)
+        # When include_update_tier is set, use lt (less than) to get main records plus updates up to that tier
+        if include_update_tier is None:
```
jusdino: Nicely done

backend/compact-connect/lambdas/python/common/cc_common/data_model/schema/provider/record.py
```diff
@@ -225,19 +225,30 @@ class ProviderUpdateRecordSchema(BaseRecordSchema, ChangeHashMixin):
...
-        # This needs to include a POSIX timestamp (seconds) and a hash of the changes
-        # to the record. We'll use the current time and the hash of the updatedValues
+        # This needs to include a iso formatted datetime string and a hash of the changes
+        # to the record. We'll use the createDate and the hash of the updatedValues
         # field for this.
         change_hash = self.hash_changes(in_data)
-        in_data['sk'] = (
-            f'{in_data["compact"]}#PROVIDER#UPDATE#{int(config.current_standard_datetime.timestamp())}/{change_hash}'
-        )
+        in_data['sk'] = f'{in_data["compact"]}#UPDATE#2#provider/{in_data["createDate"]}/{change_hash}'
```
jusdino: Should we be using your update tiers enum in these `@post_dump` hooks?

backend/compact-connect/lambdas/python/disaster-recovery/handlers/rollback_license_upload.py
```diff
@@ -0,0 +1,1051 @@
...
+    # Validate time window
+    if start_datetime >= end_datetime:
+        logger.error('Start time must be before end time')
+        return {
+            'rollbackStatus': 'FAILED',
+            'error': 'Start time must be before end time',
+        }
+
+    time_window_seconds = (end_datetime - start_datetime).total_seconds()
+    if time_window_seconds > MAX_ROLLBACK_WINDOW_SECONDS:
```
jusdino: This is a lot of data loading and validation. Should we be using a data class and/or Schema here?

backend/compact-connect/lambdas/python/disaster-recovery/handlers/rollback_license_upload.py
```diff
@@ -0,0 +1,1051 @@
...
+    Build and execute DynamoDB transactions to revert provider records.
+
+    This function processes all records in a single pass:
+    - Checks eligibility (returns ProviderSkippedDetails if ineligible)
+    - Builds transaction items
+    - Executes transactions
+
+    Returns either a summary of what was reverted or details about why the provider was skipped.
+    """
+    from cc_common.data_model.schema.license import LicenseData
```
jusdino: Why local imports here?

backend/compact-connect/lambdas/python/disaster-recovery/handlers/rollback_license_upload.py
```diff
@@ -0,0 +1,1051 @@
...
+    reverted_privileges = []
+    updates_deleted_sks = []  # List of SKs for deleted update records
+    ineligible_updates: list[IneligibleUpdate] = []
+
+    # Helper functions for cleaner item building
+    def add_put(item: dict, update_record: bool):
+        """
+        Add a Put operation to the appropriate list.
+
+        Args:
```
jusdino: Got some google-style doc strings sneaking in here

backend/compact-connect/lambdas/python/disaster-recovery/handlers/rollback_license_upload.py
```diff
@@ -0,0 +1,1051 @@
...
+                updates_deleted_sks.append(serialized_license_update['sk'])
+                logger.info(
+                    'Will delete license update record if provider is eligible for rollback',
+                    update_type=license_update.updateType,
+                    license_type=license_update.licenseType,
+                )
+
+        # if license record was created during the window, delete it
+        if (
+            license_record.firstUploadDate is not None
```
jusdino: This means that we will need to be sure to never even _try_ to run a reversion for a time window that starts before this feature is deployed to prod, right? Otherwise we risk deleting update records but leaving the new license in place?

backend/compact-connect/lambdas/python/disaster-recovery/handlers/rollback_license_upload.py
```diff
@@ -0,0 +1,1051 @@
...
+        return ProviderRevertedSummary(
+            provider_id=provider_id,
+            licenses_reverted=reverted_licenses,
+            privileges_reverted=reverted_privileges,
+            updates_deleted=updates_deleted_sks,
+        )
+
+    _perform_transaction(transaction_items, provider_id)
+
+    # Now read all the license records for the provider and update the provider record
```
jusdino: Why do we wait until after all the license, privilege, and update record revisions to update the provider records?

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_data_model/test_provider_transformations.py
```diff
@@ -2,16 +2,18 @@
...
 from .. import TstFunction
 
+MOCK_CURRENT_DATETIME_STRING = '2024-11-08T23:59:59+00:00'
+
 
 @mock_aws
 class TestTransformations(TstFunction):
     # Yes, this is an excessively long method. We're going with it for sake of a single illustrative test.
-    @patch('cc_common.config._Config.current_standard_datetime', datetime.fromisoformat('2024-11-08T23:59:59+00:00'))
+    @patch('cc_common.config._Config.current_standard_datetime', datetime.fromisoformat(MOCK_CURRENT_DATETIME_STRING))
```
jusdino: Good idea.

backend/compact-connect/lambdas/python/disaster-recovery/handlers/rollback_license_upload.py
```diff
@@ -0,0 +1,1051 @@
...
+            and upload_window_start_datetime <= license_record.firstUploadDate <= upload_window_end_datetime
+        ):
+            serialized_license_record = license_record.serialize_to_database_record()
+            add_delete(serialized_license_record['pk'], serialized_license_record['sk'], update_record=False)
+            logger.info('Will delete license record (created during upload) if provider is eligible for rollback')
+            reverted_licenses.append(
+                RevertedLicense(
+                    jurisdiction=license_record.jurisdiction,
+                    license_type=license_record.licenseType,
+                    revision_id=uuid4(),
```
jusdino: Can you talk me through the utility of a unique id for each record revision? I'm a little fuzzy on what that buys us.

backend/compact-connect/lambdas/python/disaster-recovery/handlers/rollback_license_upload.py
landonshumway-ia: Yes that is true. We really can't perform a rollback for any upload that occurs before this feature is rolled out.

backend/compact-connect/lambdas/python/disaster-recovery/handlers/rollback_license_upload.py
landonshumway-ia: It is much, much simpler that way, rather than trying to determine what the final outcome should be based on the rollback. Right now, the find_best_license logic assumes the data is CURRENTLY in a valid state, which during a rollback is not the case. Once we've rolled back all the other records, then we can leverage the existing logic to determine the best license for the provider record and roll them back.

backend/compact-connect/lambdas/python/disaster-recovery/handlers/rollback_license_upload.py
landonshumway-ia: Good point, we could set the revision id once per provider, so that the revision id is then treated somewhat as a type of transaction id, since we rollback these records one provider at a time.

We could also set a revision id per execution, but at that point it may make more sense to use the execution id, but we already define that in the bucket key where the file is stored, so that seems redundant.

Do you have a preference?

backend/compact-connect/lambdas/python/disaster-recovery/handlers/rollback_license_upload.py
landonshumway-ia: good catch, those should be at the top

backend/compact-connect/lambdas/python/disaster-recovery/handlers/rollback_license_upload.py
jusdino: I think having an execution id in the data events would be most valuable. A per-provider id could be helpful but maybe less so if we already tag all changes for the entire upload-revert with a single id. If the execution id is already in the s3 object key, that's probably sufficient for s3-report purposes.

backend/compact-connect/lambdas/python/common/cc_common/event_bus_client.py
```diff
@@ -437,3 +437,105 @@ def publish_investigation_closed_event(
...
+        license_revert_detail_schema = LicenseRevertDetailSchema()
+        deserialized_detail = license_revert_detail_schema.dump(event_detail)
+
+        self._publish_event(
+            source=source,
+            detail_type='license.revert',
+            detail=deserialized_detail,
+            event_batch_writer=event_batch_writer,
+        )
+
```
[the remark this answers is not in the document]
landonshumway-ia: I've added validation on the dumped record

backend/compact-connect/lambdas/python/migration/migrate_update_sort_keys/main.py
```diff
@@ -0,0 +1,187 @@
...
+    while True:
+        response = config.provider_table.scan(
+            FilterExpression=Attr('type').eq(ProviderRecordType.LICENSE_UPDATE)
+            | Attr('type').eq(ProviderRecordType.PROVIDER_UPDATE)
+            | Attr('type').eq(ProviderRecordType.PRIVILEGE_UPDATE),
+            **scan_pagination,
+        )
+
+        items = response.get('Items', [])
+        update_records.extend(items)
```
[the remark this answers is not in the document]
landonshumway-ia: I normally would agree, but I just ran the numbers for this:

in the test env, we currently have less than 2k update records. The average size of these records is 800 bytes, so having all those records in memory would be around 1.6 MB. The data migration lambda has a memory size of 3 GB, so we have more than enough headroom for our current load.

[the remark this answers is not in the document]
landonshumway-ia: That being said, if you really want to update the algorithm to grab them in batches, I can update the current script, but this seems unnecessary.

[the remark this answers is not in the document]
jusdino: Be sure we make a point to do that purge _before_ this is deployed, then 👍

landonshumway-ia:

backend/compact-connect/lambdas/python/disaster-recovery/tests/function/test_rollback_license_upload.py
```diff
@@ -0,0 +1,1532 @@
...
+                    'dateOfExpiration': (upload_datetime - timedelta(days=365)).date(),
+                    'licenseStatus': 'inactive',
+                    'familyName': MOCK_UPDATED_FAMILY_NAME,
+                    'givenName': MOCK_UPDATED_GIVEN_NAME,
+                },
+            }
+        )
+
+        # Update the license record to reflect the new expiration and status
+        updated_license = self.test_data_generator.put_default_license_record_in_provider_table(
```
jlkravitz: Will this overwrite the first one created above?

backend/compact-connect/lambdas/python/disaster-recovery/tests/function/test_rollback_license_upload.py
```diff
@@ -0,0 +1,1532 @@
...
+                'jurisdiction': self.license_jurisdiction,
+                'familyName': MOCK_ORIGINAL_FAMILY_NAME,
+                'givenName': MOCK_ORIGINAL_GIVEN_NAME,
+                'dateOfExpiration': (self.default_start_datetime + timedelta(days=30)).date(),
+                'firstUploadDate': license_upload_datetime,
+                'licenseStatus': 'active',
+            }
+        )
+
+        # old update record before upload window (e.g., RENEWAL)
```
jlkravitz: This isn't a renewal, right?

backend/compact-connect/lambdas/python/disaster-recovery/tests/function/test_rollback_license_upload.py
```diff
@@ -0,0 +1,1532 @@
...
+                        'occurred after rollback end time. '
+                        'Manual review required.',
+                    }
+                ],
+            },
+            results_data,
+        )
+
+    def test_expected_s3_object_stored_when_provider_schema_validation_fails_during_rollback(self):
+        """Test that failed provider details are correctly stored in S3 results when a validation exception occurs."""
```
jlkravitz: Is this just a stress test that should be impossible? Bad data gets into database -> rollback attempts to read that bad data?

backend/compact-connect/lambdas/python/disaster-recovery/tests/function/test_rollback_license_upload.py
```diff
@@ -0,0 +1,1532 @@
...
+            Bucket=self.config.disaster_recovery_results_bucket_name,
+            Key=s3_key,
+            Body=json.dumps(existing_results_data, indent=2),
+            ContentType='application/json',
+        )
+
+        final_results_data = self._perform_rollback_and_get_s3_object()
+
+        # Verify: All existing data is preserved and new data is appended
+        # Note: All keys should now be camelCase for consistency
```
jlkravitz: What do you mean should _now_ be camelCase?

backend/compact-connect/stacks/disaster_recovery_stack/license_upload_rollback_step_function.py
```diff
@@ -0,0 +1,245 @@
...
+                },
+            ],
+        )
+
+        # Create state machine
+        self.state_machine = StateMachine(
+            self,
+            'LicenseUploadRollbackStateMachine',
+            definition_body=DefinitionBody.from_chainable(definition),
+            timeout=Duration.hours(8),  # Long timeout for processing many providers
```
jlkravitz: I assume this is handled, but what happens if privileges/licenses/providers are updated during this long-running step function? e.g., if a license is encumbered during a time-consuming rollback and should be skipped. Is there any way for these records to get into partial states?

jlkravitz (commented):
Mostly questions arising from my lack of knowledge, plus a few suggestions!

backend/compact-connect/lambdas/python/disaster-recovery/tests/function/test_rollback_license_upload.py
landonshumway-ia: Ah, that's some LLM cruft that I missed while designing the rollback results schema. I'll remove the comment.

backend/compact-connect/stacks/disaster_recovery_stack/license_upload_rollback_step_function.py
landonshumway-ia: Correct, the check is a read operation, and the transaction is a write operation, and these cannot be performed in the same api call. @jusdino has discussed the potential of eventually adding revision numbers to our provider records, which would allow us to detect if another process has modified the record since our last check before writing records using a condition expression (https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.ConditionExpressions.html)

For now, I have documented this race condition in the `LICENSE_UPLOAD_ROLLBACK.md` documentation so that future devs will be aware of it and hopefully run the process during low traffic to reduce the risk of this ever occurring.

jlkravitz: Got it. That sounds like a plan to me. Thanks for explaining.

jlkravitz (approved):

PR #1238, opened 2025-12-26
Fix/update lambda logger dependency to be compatible with Node.js 24

landonshumway-ia:
While smoke testing a new feature, it was discovered that the lambda responsible for sending out Cognito emails was no longer functioning as a result of the recent update to the Node.js 24 runtime with the following error message:
```
[UserLambdaValidationException] Errors were encountered during user creation. Please review the errors and retry.

message: CustomMessage failed with error ERROR: AWS Lambda has removed support for callback-based function handlers starting with Node.js 24. You need to modify this function to use a supported handler signature to use Node.js 24 or later. For more information see https://docs.aws.amazon.com/lambda/latest/dg/nodejs-handler.html..
```

Although we confirmed that the lambda handler definition was using the proper async/await syntax, further investigation showed that the lambda is wrapped with a powertools logger decorator which was altering the way the function was called which was incompatible with the Node.js runtime upgrade. The solution was to upgrade the lambda powertools logger dependency to the latest version.

## Release Notes

* **Chores**
  * Updated internal logging dependency to a newer version.

### DIFF
--- a/backend/compact-connect/lambdas/nodejs/package.json
+++ b/backend/compact-connect/lambdas/nodejs/package.json
@@ -40,7 +40,7 @@
     "typescript": "5.6.3"
   },
   "dependencies": {
-    "@aws-lambda-powertools/logger": "^2.10.0",
+    "@aws-lambda-powertools/logger": "^2.30.0",
     "@aws-sdk/client-dynamodb": "^3.901.0",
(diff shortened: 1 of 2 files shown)


jusdino (approved):

jlkravitz (approved):
LGTM

PR #1242, opened 2025-12-30
increase memory size of authorize.net related lambdas

landonshumway-ia:
While smoke testing the privilege purchasing flow to verify another feature, it was discovered that lambdas using the authorize.net SDK set with a memory size of 256MB were failing due `Runtime.OutOfMemory` failures. It appears that the recent library updates we've made to the project have put these lambdas over their memory limit. This increases the memory size of those related lambdas to be over 1GB to avoid this from happening in the future.

This also includes some formatting changes from the Rust formatter so the lint checks will pass.

## Summary by CodeRabbit

* **Improvements**
  * Increased memory allocation for payment processing and purchase-privilege operations to improve performance and reliability under load.

* **Chores**
  * Added runtime logging for payment request handling to improve observability.
  * Minor import and test formatting tidy-ups with no behavioral changes.


jlkravitz (approved):
@isabeleliassen  good to merge!

PR #1244, opened 2026-01-06
hotfix/do not include dynamic keys in removedValues diff

landonshumway-ia:
We have identified a case where the diff that checks for license updates was including the first upload date field in the list of removed values. This is not a field set by states, so it should not be included in the diff.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved provider data processing to avoid treating calculated metadata (including firstUploadDate) as removed values, preventing false removal detections.

* **Tests**
  * Expanded ingestion tests to verify update-record creation behavior and confirm removed-value reporting for actual data removals.


jusdino (approved):
Great!

PR #1317, opened 2026-03-27
Add support for Cosmetology public search

### LINKED ISSUE #1295: cosmetology public search - BE
To support the needs of the Cosmetology search API, we will need to add the ability to query license records by license number, which will require a new GSI for license records which include the license number in the partition key. 

This ticket involves designing the GSI and API contract to support searching by license number in addition to name and jurisdiction. It also includes implementing the API for the frontend to use.

### Tasks
- [ ] TODO
- [ ] Automated tests
- [ ] API Docs
- [ ] Postman collection

landonshumway-ia:
Cosmetology has slightly different requirements for public searching given the expected volume of practitioners with similar names. We need to support additional search parameters, such as license number. In order to accommodate these various patterns, we have determined to leverage the OpenSearch Domain to support public search, while keeping the request schema similar in shape to the public search request body used by JCC.

Closes #1295 

## Summary by CodeRabbit

* **New Features**
  * Public license search endpoint added (query by licenseNumber, jurisdiction, givenName, familyName) with opaque cursor pagination and sorting.

* **Updates**
  * Public provider responses now include a licenses array; public privilege items simplified (history/adverseActions removed).
  * Request validation and API schemas tightened (licenseNumber filter, pageSize limits) and public response shapes revised.
  * Public query routed to a dedicated search service and given monitoring; index-reset/config options added.

* **Tests**
  * Added comprehensive public-search tests; removed legacy public-query tests.


backend/cosmetology-app/lambdas/python/search/opensearch_client.py
```diff
@@ -80,6 +83,282 @@ def create_alias(self, index_name: str, alias_name: str) -> None:
...
+
+        :param index_name: The index to delete
+        :raises CCInternalException: If all retry attempts fail
+        """
+        self._execute_with_retry(
+            operation=lambda: self._client.indices.delete(index=index_name),
+            operation_name=f'delete_index({index_name})',
+        )
+
+    def create_provider_index_with_alias(
```
ChiefStief: Why are these methods being added in conjunction with the licenses being returned in search?

backend/cosmetology-app/stacks/api_stack/v1_api/api_model.py
```diff
@@ -819,55 +813,36 @@ def _update_type_schema(self) -> JsonSchema:
...
                     pattern=cc_api.YMD_FORMAT,
                 ),
-                'encumbranceType': self._encumbrance_type_schema,
-                'clinicalPrivilegeActionCategories': JsonSchema(
-                    type=JsonSchemaType.ARRAY,
-                    description='The categories of clinical privilege action',
-                    items=JsonSchema(type=JsonSchemaType.STRING),
-                ),
+                'clinicalPrivilegeActionCategories': self._clinical_privilege_action_categories_schema,
             },
```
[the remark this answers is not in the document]
landonshumway-ia: That commit is for another PR and has been removed here.

backend/cosmetology-app/lambdas/python/common/cc_common/data_model/schema/provider/api.py
```diff
@@ -202,17 +203,31 @@ class ProviderPublicResponseSchema(ForgivingSchema):
...
+    Jurisdiction is renamed to licenseJurisdiction for parity with JCC implementation.
+    """
+
+    providerId = Raw(required=True, allow_none=False)
+    givenName = String(required=True, allow_none=False, validate=Length(1, 100))
+    familyName = String(required=True, allow_none=False, validate=Length(1, 100))
+    licenseJurisdiction = String(required=True, allow_none=False)
+    compact = Compact(required=True, allow_none=False)
+    licenseType = String(required=True, allow_none=False)
+    licenseNumber = String(required=True, allow_none=False, validate=Length(1, 100))
```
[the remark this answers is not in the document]
landonshumway-ia: Yes, Cosmetology requires license numbers be included for every license upload.

ChiefStief (approved):

backend/cosmetology-app/lambdas/python/common/cc_common/data_model/schema/provider/api.py
```diff
@@ -202,17 +203,31 @@ class ProviderPublicResponseSchema(ForgivingSchema):
...
+    # licenses and privileges and does not return any adverse action data.
+    licenses = List(Nested(LicensePublicResponseSchema(), required=False, allow_none=False))
     privileges = List(Nested(PrivilegePublicResponseSchema(), required=False, allow_none=False))
-    # Note the lack of `licenses` here: we do not return license data for public endpoints
+
+
+class PublicLicenseSearchResponseSchema(ForgivingSchema):
+    """
+    License object fields returned by the public query providers endpoint (OpenSearch-backed).
+    Jurisdiction is renamed to licenseJurisdiction for parity with JCC implementation.
```
jlkravitz: Renamed from what? Can you just add the original name and the source of that name (OpenSearch?) for clarity?

backend/cosmetology-app/lambdas/python/search/tests/function/test_public_search_providers.py
```diff
@@ -0,0 +1,553 @@
...
+        )
+        response = public_search_api_handler(event, self.mock_context)
+        self.assertEqual(400, response['statusCode'])
+        body = json.loads(response['body'])
+        self.assertIn('Invalid sort key', body['message'])
+        mock_opensearch_client.search.assert_not_called()
+
+    @patch('handlers.public_search.opensearch_client')
+    def test_no_search_criteria_returns_200(self, mock_opensearch_client):
+        """Test that at least one of licenseNumber, jurisdiction, or familyName is required."""
```
jlkravitz: Isn't this comment contradictory with the test name, which seems to allow a search request with none of these three fields?

backend/cosmetology-app/lambdas/python/search/tests/function/test_public_search_providers.py
```diff
@@ -0,0 +1,553 @@
...
+        from handlers.public_search import public_search_api_handler
+
+        event = self._create_public_api_event('cosm', body={'query': {'familyName': 'x'}})
+        event['resource'] = '/v1/public/compacts/{compact}/providers/other'
+        response = public_search_api_handler(event, self.mock_context)
+        self.assertEqual(400, response['statusCode'])
+        self.assertIn('Unsupported method or resource', json.loads(response['body'])['message'])
+
+    @patch('handlers.public_search.opensearch_client')
+    def test_terminal_page_returns_last_key_null(self, mock_opensearch_client):
```
jlkravitz: Not really problematic, but I think this covers basically the same thing as `test_response_last_key_null_when_fewer_hits_than_page_size`?

backend/cosmetology-app/lambdas/python/search/tests/function/test_manage_opensearch_indices.py
```diff
@@ -67,6 +67,24 @@ def _when_testing_mock_opensearch_client(
         else:
             mock_client_instance.index_exists.return_value = index_exists_return_value
 
+        # We want to make assertions on the opensearch client calls that are made during index creation
+        # to ensure the alias/indices are created as expected. We also need to ensure we are checking against
+        # the actual index mapping that is used in the runtime logic. This points the mock to the actual
+        # function pointers in the opensearch_client client.
```
jlkravitz: Where are we actually making these assertions/using this functionality in tests?

backend/cosmetology-app/lambdas/python/search/handlers/public_search.py
```diff
@@ -0,0 +1,206 @@
...
+        raw_body = event.get('body') or '{}'
+        body = schema.loads(raw_body)
+    except ValidationError as e:
+        logger.warning('Invalid public query request body', errors=e.messages)
+        raise CCInvalidRequestException(f'Invalid request: {e.messages}') from e
+
+    return body
+
+
+def _decode_public_cursor(last_key: str | None) -> dict | None:
```
jlkravitz, line 139: Just for my understanding, isn't a lot of this opensearch stuff already implemented via the staff search? e.g., cursors

backend/cosmetology-app/stacks/search_persistent_stack/populate_provider_documents_handler.py
```diff
@@ -81,8 +81,9 @@ def __init__(
             alarm_topic=alarm_topic,
         )
 
-        # Grant the handler write access to the OpenSearch domain
-        opensearch_domain.grant_write(self.handler)
+        # Grant read/write HTTP to the domain (same as index manager). resetIndexes and normal indexing need
+        # HEAD/GET for alias and index existence checks, plus PUT/POST for bulk index and DELETE for index reset.
```
jlkravitz, line 85: Tell me if I'm misunderstanding, but since 'index reset' is only for local development, is it worth thinking about removing the 'DELETE' capability?

jlkravitz (commented):
Looks good! Few questions/comments.

backend/cosmetology-app/lambdas/python/search/handlers/public_search.py
landonshumway-ia: The staff user search API more closely follows the native OpenSearch DSL when it comes to pagination because the frontend was updated to use the OpenSearch DSL as part of the API contract. This public search endpoint was originally implemented to work with DynamoDB pagination, which is much different than OpenSearch. Because the JCC public search implementation is still relying on the DynamoDB solution, we have to maintain parity with that API request/response schema, which means the way we are handling pagination has to be transformed to match that pattern so we don't break the frontend.

backend/cosmetology-app/stacks/search_persistent_stack/populate_provider_documents_handler.py
landonshumway-ia: We need it in our test environments as well, not just sandbox. I'm willing to consider alternatives, but we needed the ability to reset those indices if the test environment gets into a bad state, which our test environments are prone to do given that they only have a couple data nodes.

backend/cosmetology-app/lambdas/python/search/tests/function/test_public_search_providers.py
landonshumway-ia: Yep, the comment is how it was originally implemented. I changed the title after talking to John about supporting empty queries, but missed the comment.

backend/cosmetology-app/lambdas/python/common/cc_common/data_model/schema/provider/api.py
landonshumway-ia: That logic was moved out of the schema and into [the lambda handler here](https://github.com/csg-org/CompactConnect/blob/096f2e58f1a96447bfcd3fbcf17a51b09a988c7e/backend/cosmetology-app/lambdas/python/search/handlers/public_search.py#L75-L78) to reduce confusion. I'll remove this line since we have a comment to explain the renaming there.

backend/cosmetology-app/lambdas/python/search/handlers/public_search.py
```diff
@@ -0,0 +1,206 @@
...
+                {'providerId': os_dir},
+                {'_id': 'asc'},
+            ]
+        case 'dateOfUpdate':
+            return [
+                {'dateOfUpdate': os_dir},
+                {'_id': 'asc'},
+            ]
+        case _:
+            raise CCInvalidRequestException(f"Invalid sort key: '{sort_key}'")
```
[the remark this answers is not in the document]
landonshumway-ia: Could we also update the sort fields to use the nested license fields?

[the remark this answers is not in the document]
landonshumway-ia: This has been addressed. Please verify

backend/cosmetology-app/lambdas/python/search/handlers/public_search.py
jlkravitz: Ah, I see. Makes total sense. Thanks for the explanation.

backend/cosmetology-app/stacks/search_persistent_stack/populate_provider_documents_handler.py
jlkravitz: Got it. I more worry about those permissions being used on production unnecessarily, but more just a flag that I leave up to your judgement.

jlkravitz (approved):

backend/cosmetology-app/stacks/search_persistent_stack/populate_provider_documents_handler.py
landonshumway-ia: Agreed. I added a gate to this script that doesn't allow developers to call the reset index flow in production for that reason. Given that the OpenSearch records are replicated data from DynamoDB that can be regenerated on demand, I believe this is sufficient.

PR #1466, opened 2026-04-14
fix/ zap scan

jlkravitz:
## Summary

The ZAP scan has been failing on every PR since we switched from a two-branch model (`development` + `main`) to a single `main` branch with tag-based deployments. Two root causes:

- **Secrets inaccessible**: Since IA opens PRs from their fork of CompactConnect, the GitHub Actions `pull_request` trigger gets `Secret source: None` — the Cognito auth step fails immediately.
- **Scanning the wrong state**: The scan targets the deployed test environment, but PR code isn't deployed to test yet, so results are meaningless for the PR anyway.

This PR moves the ZAP scan to a **weekly schedule + manual dispatch**, which matches the original sprint-cadence intent and runs in a context where secrets are available.

### Scan improvements

- **Enriched OAS specs with real test data** — ZAP was using literal parameter names as URL values (e.g. `/compacts/compact/` instead of `/compacts/aslp/`), causing every request to fail validation. A new enrichment script (`owasp-zap/enrich-oas-for-zap.py`) patches the CDK-generated specs with real provider IDs, jurisdiction codes, etc. before ZAP ingests them. New endpoints added to the spec are automatically covered.
- **Removed state API** — it uses a separate Cognito pool (`StateAuthUsers`) from the staff pool, so every request returned 401. Excluded until a state API test credential is provisioned.
- **Removed frontend** (`app.test.compactconnect.org`) — it's an S3-hosted SPA where all findings were CloudFront header configuration issues, not application vulnerabilities.
- **Added search API** (`search.test.compactconnect.org`) to scan coverage.
- **Suppressed known false positives** — SQL injection alerts on `/provider-users/initiateRecovery` (DynamoDB, not SQL) and `/credentials/payment-processor` (error from Authorize.net, not injection).
- **Stripped DELETE endpoints** from the spec to prevent ZAP from deleting staff users during scans.
- **Capped active scan at 45 minutes** to stay under Cognito token expiry (~60m).

### Test user changes

Updated the ZAP scan staff user to have `aslp/admin` compact-level permissions plus full `oh` and `ky` jurisdiction permissions, expanding scope coverage from 4/100 to covering most ASLP endpoints.
 
 
 ### Follow-up
  - Add provider user scanning. Requires a test provider user credential + separate token.
  - Add state API scanning (state-api.test.compactconnect.org)

### DIFF
--- a/.github/workflows/zap-scan-test.yml
+++ b/.github/workflows/zap-scan-test.yml
@@ -1,9 +1,6 @@
 name: ZAP-Scan-Test
 
 on:
-  pull_request:
-    branches:
-      - main
   # Weekly scheduled scan of the deployed test environment
   schedule:
     - cron: '0 6 * * 1'  # Every Monday at 06:00 UTC
--- a/.github/workflows/zap-scan-test.yml
+++ b/.github/workflows/zap-scan-test.yml
@@ -15,6 +15,7 @@ jobs:
     permissions:
       id-token: write
       contents: read
+      security-events: write
 
     steps:
       - name: Log trigger info
@@ -59,6 +60,12 @@ jobs:
         uses: zaproxy/action-af@v0.1.0
         with:
           plan: 'owasp-zap/data/test-automation.yml'
+
+      - name: Upload SARIF to GitHub Security
+        if: always()
+        uses: github/codeql-action/upload-sarif@v3
+        with:
+          sarif_file: report/zap-results.sarif
 
       - name: Upload Report
         if: always()
--- a/.github/workflows/zap-scan-test.yml
+++ b/.github/workflows/zap-scan-test.yml
@@ -1,44 +1,35 @@
-# Compact Connect - Web Frontend Deployment - Development
-
 name: ZAP-Scan-Test
 
-# Controls when the action will run.
 on:
-  pull_request:
-    branches:
-      - main
+  # Weekly scheduled scan of the deployed test environment
+  schedule:
+    - cron: '0 6 * * 1'  # Every Monday at 06:00 UTC
 
-  # Allows you to run this workflow manually from the Actions tab
+  # Manual trigger for ad-hoc runs
   workflow_dispatch:
 
-# A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
   ZapScan:
-    # Runner OS
     runs-on: ubuntu-latest
 
-    # Job needs id-token access to work with GitHub OIDC to AWS IAM Role
     permissions:
       id-token: write
       contents: read
 
     steps:
-      - run: echo "🎉 The job was automatically triggered by a ${{ github.event_name }} event."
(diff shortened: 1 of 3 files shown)


owasp-zap/enrich-oas-for-zap.py
```diff
@@ -0,0 +1,99 @@
...
+
+def _fix_missing_array_items(obj):
+    """Recursively add items: {type: string} to arrays missing an items field."""
+    if not isinstance(obj, dict):
+        return
+    if obj.get("type") == "array" and "items" not in obj:
+        obj["items"] = {"type": "string"}
+    for value in obj.values():
+        if isinstance(value, dict):
+            _fix_missing_array_items(value)
```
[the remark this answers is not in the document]
jlkravitz: ZAP did not reject the OAS so we will ignore this.

owasp-zap/data/test-automation.yml
```diff
@@ -2,13 +2,11 @@ env:
...
     urls:
-    - https://app.test.compactconnect.org
     - https://api.test.compactconnect.org
-    - https://state-api.test.compactconnect.org
+    - https://search.test.compactconnect.org
     includePaths:
-    - https://app.test.compactconnect.org.*
     - https://api.test.compactconnect.org.*
-    - https://state-api.test.compactconnect.org.*
+    - https://search.test.compactconnect.org.*
```
landonshumway-ia, line 9: Do we not also want to keep scanning the state-api?

.github/workflows/zap-scan-test.yml
```diff
@@ -56,10 +47,23 @@ jobs:
           COGNITO_PASSWORD: ${{ secrets.TEST_ZAP_PASSWORD_STAFF }}
```
landonshumway-ia, line 47: Does the zap scan authenticate just as a staff user? Or do we also have a test provider user for calling the provider authenticated endpoints?

landonshumway-ia (commented):
Thanks for adding these improvements. A couple questions:

owasp-zap/data/test-automation.yml
jlkravitz: That was listed as a follow-up in the PR description. Our old scans were not doing anything meaningful there since it didn't have proper credentials and just hit 401s.

.github/workflows/zap-scan-test.yml
jlkravitz: Same thing here– That was listed as a follow-up in the PR description. Our original zap setup that Justin helped with only fetched a token for one type of user- the staff user. So it'll require a little more work to get the other accounts and tokens setup.

I'm open to putting that all into this PR, but figure step (1) was just to get it back working again.

landonshumway-ia (approved):
Approved. The following tickets will track future enhancements:

https://github.com/csg-org/CompactConnect/issues/1467
https://github.com/csg-org/CompactConnect/issues/1468

jlkravitz:
@isabeleliassen This is ready to merge.

PR #1470, opened 2026-04-17
Extend ZAP scan to provider users and state API M2M (#1467, #1468)

### LINKED ISSUE #1467: Add provider user scanning to ZAP
Follow-up from #1466.

The License API has 10 endpoints using a separate Cognito user pool (`ProviderUsersPoolAuthorizer`) for provider/licensee users. The ZAP scan currently uses a staff user credential, so all provider endpoints return 401.

**Endpoints not being scanned:**
- `POST /v1/purchases/privileges`
- `GET /v1/provider-users/me`
- `GET /v1/compacts/{compact}/attestations/{attestationId}`
- `GET /v1/purchases/privileges/options`
- `GET /v1/provider-users/me/jurisdiction/{jurisdiction}/licenseType/{licenseType}/history`
- `PUT /v1/provider-users/me/home-jurisdiction`
- `POST /v1/provider-users/me/military-affiliation`
- `PATCH /v1/provider-users/me/military-affiliation`
- `PATCH /v1/provider-users/me/email`
- `POST /v1/provider-users/me/email/verify`

**What's needed:**
- Provision a test provider user in the test environment's provider Cognito pool
- Add provider-specific secrets to the GitHub repo (`TEST_COGNITO_PROVIDER_POOL_ID`, etc.)
- Update the workflow to obtain a second token and either run a second scan or configure a second ZAP context

### LINKED ISSUE #1468: Add state API scanning to ZAP
Follow-up from #1466.

The state API (`state-api.test.compactconnect.org`) has 4 endpoints and uses its own Cognito user pool (`StateAuthUsers`) for machine-to-machine authentication. It was excluded from the ZAP scan because no test credential exists for that pool.

**Endpoints not being scanned:**
- `POST /v1/compacts/{compact}/jurisdictions/{jurisdiction}/licenses`
- `GET /v1/compacts/{compact}/jurisdictions/{jurisdiction}/licenses/bulk-upload`
- `POST /v1/compacts/{compact}/jurisdictions/{jurisdiction}/providers/query`
- `GET /v1/compacts/{compact}/jurisdictions/{jurisdiction}/providers/{providerId}`

**What's needed:**
- Provision a test client credential in the StateAuthUsers pool
- Add state-specific secrets to the GitHub repo
- Update `owasp-zap/data/test-automation.yml` to re-include the state API context and OpenAPI import
- May need separate authentication flow (machine-to-machine, not SRP)

jlkravitz:
Extends the ZAP scan from staff-only to **all three auth tiers** (Staff, Provider, State M2M) and hardens the scan's configuration and result-handling.

Implements #1467 (provider-user scanning) and #1468 (state-api scanning).

## Multi-tier authentication
- `bearer-token.js` routes each request to the correct token by host + path: the state-api host gets the M2M token **plus a per-request ECDSA-SHA256 signature** (per `docs/client_signature_auth.md`); `/v1/provider-users/*`, `/v1/purchases/*`, and `GET attestations-by-id` get the provider **ID** token; everything else gets the staff token.
- The workflow runs three auth steps, masks each token, and forwards them (and the signing key) into the ZAP container via `docker_env_vars` (`zaproxy/action-af@v0.2.0`).
- `main.js` takes `--mode=staff|provider`; `get-m2m-token.sh` performs the client-credentials grant.
- `enrich-oas-for-zap.py` injects valid path/body examples so ZAP gets 2xx baselines to fuzz around instead of 400-ing on every request.

## Scan hardening
- **Active-scan cap = 75 min.** A full scan measured ~57 min; the prior 330 needlessly held a runner slot.
- **Global SQL-injection (ruleId 40018) false-positive filter.** CompactConnect is DynamoDB-only, so 40018 is a known FP class; this replaces the brittle per-URL list while leaving the rule enabled for any future non-DynamoDB sink.
- **`exitStatus` fails the build on any High alert** (Medium is reported but non-fatal) instead of silently passing via `warnExitValue: 0`.
- **JSON report** emitted alongside the HTML and uploaded as the `ZAP Report` artifact.
- State M2M token expiry (60 min vs ~57-min scan) noted in `bearer-token.js` so a future cap increase doesn't silently 401.

## Result review
Findings are triaged via a `security`-labelled GitHub issue rather than a chat notification. Latest validated run: [27373665498](https://github.com/csg-org/CompactConnect/actions/runs/27373665498) — **0 High / 6 Medium / 20 Low / 13 Info**, with all three auth tiers producing clean 2xx baselines. Tracked for review in #1644.

## Testing
- Two green end-to-end runs against the test environment; the active scan finished **naturally at 57m14s**, under the 75-min cap.
- Confirmed staff / provider / state auth all succeed: provider `GET /v1/provider-users/me` → 200, state-api `POST .../providers/query` → 200.

## Notes
- The workflow runs on the **weekly schedule + manual dispatch**; the temporary `pull_request` trigger used to validate the scan during development has been removed.
- Intermediate `TEMP:` commits will be collapsed on squash-merge.

Closes #1467
Closes #1468

## Release Notes

* **Chores**
  * Enhanced security scanning workflow to support multiple authentication methods: staff users, provider users, and state API M2M authentication.
  * Upgraded OWASP ZAP scanning action and added automatic ECDSA-SHA256 signature generation for state API requests.
  * Improved documentation with comprehensive setup and configuration guidance for all authentication scenarios and local testing.


jlkravitz:
@landonshumway-ia This is ready for your review!

.github/workflows/zap-scan-test.yml
```diff
@@ -63,11 +97,25 @@ jobs:
...
+            backend/compact-connect/docs/api-specification/latest-oas30-zap.json
 
       - name: ZAP Scan
-        uses: zaproxy/action-af@v0.1.0
+        uses: zaproxy/action-af@v0.2.0
+        env:
+          # State API requires ECDSA signatures on top of the M2M bearer token.
+          # See backend/compact-connect/docs/client_signature_auth.md.
+          ZAP_STATE_SIGNATURE_PRIVATE_KEY: ${{ secrets.TEST_ZAP_STATE_SIGNATURE_PRIVATE_KEY }}
+          ZAP_STATE_SIGNATURE_KEY_ID: ${{ secrets.TEST_ZAP_STATE_SIGNATURE_KEY_ID }}
```
landonshumway-ia, line 110: Did you insert the public key into the test environment's compact configuration table as documented in: https://github.com/csg-org/CompactConnect/blob/main/backend/compact-connect/app_clients/bin/manage_signature_keys.py ?

owasp-zap/data/bearer-token.js
```diff
@@ -1,26 +1,148 @@
 /*
  * Adapted from https://github.com/zaproxy/community-scripts/blob/main/httpsender/AddBearerTokenHeader.js
- * This script adds a bearer token to all requests in scope except the authorization request itself
- * The token is retrieved from the environment variable TOKEN, which can be acquired separately from ZAP.
+ *
+ * Selects an Authorization bearer token based on the target URL. CompactConnect fronts three
+ * distinct Cognito pools behind the scan targets:
+ *   - Staff pool for most endpoints on api.test.compactconnect.org
+ *   - Provider users pool for /v1/provider-users/*, /v1/purchases/*, and the GET on
+ *     /v1/compacts/{compact}/attestations/{attestationId}
```
landonshumway-ia, line 8: Seems a little brittle to branch based off a specific section of the path in the sense that new endpoints might not be picked up, but I can't think of a better solution for this.

landonshumway-ia (approved):
Looks good, just one clarifying question to make sure the public key is set up correctly.

owasp-zap/data/bearer-token.js
jlkravitz: I agree, but that was sort of my thought here. Unless this is somehow better specified in the OAS or postman spec?

.github/workflows/zap-scan-test.yml
jlkravitz: I did this a while ago, but I confirmed that the compact configuration table in the test environment contains two public keys for KY and OH for ZAP. The app client registry has also been updated: https://docs.google.com/spreadsheets/d/1HzVoDJubwNlUm0vHK7_W9igkQG4gq2afNiukZwL0n8Y/edit?gid=0#gid=0

I assume that's sufficient but let me know if not.

owasp-zap/data/bearer-token.js
landonshumway-ia: The only way I can think of parsing the specific endpoints would be to look in the OpenAPI spec, and for every endpoint that has the 'security' field, match any endpoints that have a `ProviderUsersPoolAuthorizer` substring for the authorizer name. For example in the `/v1/purchases/privileges` [endpoint definition:](https://github.com/csg-org/CompactConnect/blob/6195c0226e0b009db2561363137fcaedea911853/backend/compact-connect/docs/internal/api-specification/latest-oas30.json#L549-L554)
```
"security": [
          {
            "SandboxAPIStackLicenseApiProviderUsersPoolAuthorizerEB7523BA": []
          }
        ]
```

The `ProviderUsersPoolAuthorizer` is specifically used by all of our authenticated endpoints that require an token from a provider account, rather than a staff user account.

.github/workflows/zap-scan-test.yml
landonshumway-ia: Yes as long as the public keys have been uploaded with the expected key id then that should be ready to use.

PR #1572, opened 2026-05-20
Adding initial SocialWork compact codebase

landonshumway-ia:
This change includes the initial addition of the Social Work Compact to the CompactConnect codebase.

The codebase from the Cosmetology compact was replicated into the `social-work-app` directory as a starting point to build off of, since both the Social Work and Cosmetology compacts follow the multi-state licensing model. The compact abbreviation has been updated across the configuration and tests, but the license types are still to be determined, so the tests still show references to the cosmetology license types in some areas. These will be updated once the license types are finalized in a future PR.

As we are adding more compacts to the project, it is necessary for us to consolidate most of our common CDK code that will be referenced across the various compact CDK apps into a common location. We initially set up the `common-cdk` for this purpose, and this PR includes pulling in all the common CDK constructs that we have found to be used consistently across both the JCC and Cosmetology projects. Test coverage has also been added for these constructs to bring our overall coverage to 90%

### Requirements List
- The AWS accounts for this new compact have been provisioned and bootstrapped for CDK deployments. This PR is ready to be merged once approved.

Closes #1573


backend/common-cdk/tests/test_compact_connect_api.py
```diff
@@ -0,0 +1,275 @@
...
+
+
+class TestCompactConnectApi(TestCase):
+    def setUp(self):
+        self.app = _make_app()
+        self.stack = _make_stack(self.app)
+
+    # --- stage name ---------------------------------------------------------
+
+    def test_stage_name_includes_environment_and_blue_suffix(self):
```
ChiefStief: Do we still want to do this blue-green thing?

backend/common-cdk/tests/test_bucket.py
```diff
@@ -0,0 +1,120 @@
...
+                    'PublicAccessBlockConfiguration': {
+                        'BlockPublicAcls': True,
+                        'BlockPublicPolicy': True,
+                        'IgnorePublicAcls': True,
+                        'RestrictPublicBuckets': True,
+                    }
+                }
+            },
+        )
+        # Both the bucket and the access logs bucket must block all public access
```
ChiefStief, line 36: This comment / assertion dont really line up.

This could pass if only the access logs bucket matched or only this bucket matched

backend/common-cdk/tests/test_bucket.py
```diff
@@ -0,0 +1,120 @@
...
+        # SSL is enforced via a bucket policy requiring aws:SecureTransport
+        policies = template.find_resources('AWS::S3::BucketPolicy')
+        policy_docs = [p['Properties']['PolicyDocument'] for p in policies.values()]
+        ssl_deny_found = any(
+            stmt.get('Condition', {}).get('Bool', {}).get('aws:SecureTransport') == 'false'
+            for doc in policy_docs
+            for stmt in doc.get('Statement', [])
+            if stmt.get('Effect') == 'Deny'
+        )
+        self.assertTrue(ssl_deny_found, 'No SSL-enforcing Deny policy statement found on bucket')
```
ChiefStief: Could the access-logs bucket actually trigger this?

backend/common-cdk/tests/test_bucket.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def test_bucket_owner_enforced_object_ownership(self):
+        Bucket(self.stack, 'Bucket', server_access_logs_bucket=self.access_logs_bucket)
+
+        template = Template.from_stack(self.stack)
+        buckets = template.find_resources(
+            CfnBucket.CFN_RESOURCE_TYPE_NAME,
+            props={'Properties': {'OwnershipControls': {'Rules': [{'ObjectOwnership': 'BucketOwnerEnforced'}]}}},
+        )
+        self.assertGreaterEqual(len(buckets), 1)
```
ChiefStief: Again, I feel like having both the access log bucket and the bucket in question present somewhat invalidates the self.assertGreaterEqual(len(buckets), 1) because it could pass by just the access log bucket an have nothing to do with this bucket

backend/common-cdk/tests/test_nodejs_function.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def test_throttle_alarm_wired_to_topic_when_alarm_topic_provided(self):
+        topic = Topic(self.stack, 'AlarmTopic')
+        _make_fn(self.stack, alarm_topic=topic)
+
+        template = Template.from_stack(self.stack)
+        alarms = template.find_resources(CfnAlarm.CFN_RESOURCE_TYPE_NAME)
+        self.assertEqual(1, len(alarms))
+        (alarm,) = alarms.values()
+        self.assertGreaterEqual(len(alarm['Properties'].get('AlarmActions', [])), 1)
```
ChiefStief: This just tests that there is an action not that it targets the topic arn

ChiefStief (changes requested):
Ok I had some minor test validity nitpicks. Other than that looks good to me

backend/common-cdk/tests/test_compact_connect_api.py
landonshumway-ia: Good call out. We don't really need it, but because JCC is already setting the stage using this pattern we will likely need to keep it for backwards compatibility since eventually all of the projects will consume this common API construct. I will extract the suffix out as a parameter so it can be changed as needed.

backend/common-cdk/tests/test_bucket.py
landonshumway-ia: Yes. Good catch. I'll update this into a snapshot test to verify the actual bucket policy.

backend/common-cdk/tests/test_bucket.py
landonshumway-ia: Great catch, that should be an assertEqual. I'll update that.

backend/common-cdk/tests/test_bucket.py
landonshumway-ia: Yeah, I'm not sure why the agent used that assertion, it's pretty weak in terms of checking CDK config, I'll change it to an assertEqual check.

backend/common-cdk/tests/test_nodejs_function.py
landonshumway-ia: I've updated it to check for the value of the alarm action.

ChiefStief (approved):
Ok, looks good!

landonshumway-ia:

backend/common-cdk/tests/test_security_profile.py
```diff
@@ -0,0 +1,14 @@
+from unittest import TestCase
```
jlkravitz, line 1: Are these tests useful? It seems to just test that the enums have particular values, but does our code actually depend on those particular values?

backend/common-cdk/tests/test_constants.py
```diff
@@ -0,0 +1,11 @@
+from unittest import TestCase
+
+from common_constructs.constants import BETA_ENV_NAME, PROD_ENV_NAME
+
+
+class TestConstants(TestCase):
```
jlkravitz, line 6: Will something actually break if the env names are changed? Just wondering what the utility of these tests are.

backend/cosmetology-app/README.md
```diff
@@ -25,14 +25,6 @@ To deploy this app, you will need:
 2) Python>=3.14 installed on your machine, preferably through a virtual environment management tool like
    [pyenv](https://github.com/pyenv/pyenv), for clean management of virtual environments across multiple Python
    versions.
-   > Note: The [purchases lambda](./lambdas/python/purchases) depends on the
```
jlkravitz, line 28: Confirming that we've now updated the lambda such that this comment is no longer relevant?

backend/multi-account/README.md
```diff
@@ -70,12 +70,15 @@ new AWS organization that we will set up here. Have them:
...
-- Create a new OU structure as follows:
+- Create a new OU structure as follows. Each compact gets its own OU under `Workflows`, with its own
+  `PreProd` and `Prod` sub-OUs inside it. Name the compact-level OU after the compact (e.g. `CompactConnect`
+  for the initial JCC compact):
 ```text
 └── Workflows
     ├── Deployment
-    ├── PreProd
-    └── Prod
+    └── CompactConnect
```
jlkravitz, line 79: Did we actually change the OU structure, or is this just updating the docs to reflect reality?

backend/multi-account/README.md
```diff
@@ -287,3 +291,140 @@ After setting up the multi-account architecture, you can deploy the log aggregat
 This will set up a CloudTrail organization trail that logs read operations on DynamoDB tables with the `-DataEventsLog` suffix across all accounts in the organization.
 
 The logs will be stored in an S3 bucket in the Logs account, and the trail itself will be managed from the Management account.
+
+## Adding a new compact
+
+Each compact runs in its own set of application accounts under the existing AWS Organization, in its own dedicated OU under `Workflows`. The Deploy account is **shared** across all compacts, so you do not provision a new Deploy account when onboarding a new compact. You need to create a new compact-level OU (with `PreProd` and `Prod` sub-OUs inside it), provision the compact's application accounts within those OUs, and stand up the compact's pipeline stacks inside the existing Deploy account.
```
jlkravitz, line 297: What's the justification for keeping deployment in one account?

backend/multi-account/README.md
```diff
@@ -287,3 +291,140 @@ After setting up the multi-account architecture, you can deploy the log aggregat
...
+
+The follow documentation mirrors the process described above for setting up the original JCC compact, but with details specific for new compacts.
```
jlkravitz:
```suggestion
The following documentation mirrors the process described above for setting up the original JCC compact, but with details specific for new compacts.
```

jlkravitz (commented):
few questions! generally looks great. I did not really review anything under `social-work` due to the magnitude and nature of the changes (i.e., mostly copies of other code). If there is any particular code that changed or could use a second look, please let me know.

backend/cosmetology-app/README.md
landonshumway-ia: The cosmetology app doesn't actually have any lambdas related to purchases. This comment was copied over from the JCC codebase and is irrelevant for the cosmetology or social work compacts.

backend/multi-account/README.md
landonshumway-ia: This is just updating the docs to reflect what was actually configured in the management account.

backend/multi-account/README.md
landonshumway-ia: This is the pattern that was established when we initially designed the strategy to support multiple compacts from the same GitHub repository. The deployment account manages the codestar connection with GitHub which allows us to trigger our CDK pipelines based on the tag prefix we set when releasing changes. Keeping all of the pipeline infrastructure in one deploy account keeps the configuration simple for all of the different environment/compact deployments, and each pipeline is restricted to a particular account based on the trust policies we set when we bootstrap each environment account to support CDK deployments.

landonshumway-ia: This is the same process we followed when setting up Cosmetology, BTW, we just didn't have it documented. I've added the documentation to make this easier for the next compact we have to set up.

backend/common-cdk/tests/test_security_profile.py
landonshumway-ia: In this case, yes. We reference the security profile enum across multiple CDK stacks. The tests would fail if somehow someone (or an agent) changed the values unexpectedly. So while they're simple, they do have a small utility.

backend/common-cdk/tests/test_constants.py
landonshumway-ia: I was going to say yes, but when I just dug into the file it looks like those constants are not actually being referenced in the codebase. It looks like that constant file was added with the intent to track string values used in multiple places, but then must have been lost in the shuffle since we are still using the hard coded strings. 

Given that the constants in that file aren't actually being used, I'll remove them and this test file for now.

landonshumway-ia: So it turns out they are being used throughout the search related stacks across the different compact folders. My editor wasn't tracking them because it couldn't resolve the common_constructs namespace package. So in that case, yes, these tests serve as a type of snapshot test to alert us if something gets changed unexpectedly, which would have major consequences in our deployment logic since we have different retention policies and OpenSearch domain configurations based on which environment we are deploying to.

backend/multi-account/README.md
jlkravitz: Helpful thanks!

jlkravitz (approved):

PR #1591, opened 2026-05-26
Chore/remove beta public endpoints

landonshumway-ia:
There is currently no valid use case for supporting public endpoints in the beta environment, as the beta environment is primarily used for state onboarding and searching records through the staff user search. This removes the public API endpoints from the beta environment for both the JCC and Cosmetology APIs

### Testing List
- Code review

Closes # 1574

## Summary by CodeRabbit

* **Bug Fixes**
  * Public provider lookup now returns "Provider not found" for providers without privileges.
  * Public provider lookup now executes with a short, deliberate delay.

* **Deployment**
  * Public lookup API is deployed only in non-beta environments.

* **Tests**
  * Added tests for provider privilege validation and environment-specific API exposure.

* **Documentation**
  * Updated API spec version timestamp.

[Review Change Stack](https://app.coderabbit.ai/change-stack/csg-org/CompactConnect/pull/1591?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/backend/compact-connect/docs/internal/api-specification/latest-oas30.json
+++ b/backend/compact-connect/docs/internal/api-specification/latest-oas30.json
@@ -2,7 +2,7 @@
   "openapi": "3.0.1",
   "info": {
     "title": "LicenseApi",
-    "version": "2026-01-14T16:16:32Z"
+    "version": "2026-05-26T15:59:12Z"
   },
   "servers": [
     {
--- a/backend/compact-connect/lambdas/python/provider-data-v1/handlers/public_lookup.py
+++ b/backend/compact-connect/lambdas/python/provider-data-v1/handlers/public_lookup.py
@@ -1,8 +1,8 @@
 from aws_lambda_powertools.utilities.typing import LambdaContext
 from cc_common.config import config, logger
 from cc_common.data_model.schema.provider.api import ProviderPublicResponseSchema, QueryProvidersRequestSchema
-from cc_common.exceptions import CCInvalidRequestException
-from cc_common.utils import api_handler
+from cc_common.exceptions import CCInvalidRequestException, CCNotFoundException
+from cc_common.utils import api_handler, delayed_function
 from marshmallow import ValidationError
 
 from . import get_provider_information
@@ -113,6 +113,7 @@ def public_query_providers(event: dict, context: LambdaContext):  # noqa: ARG001
 
 
 @api_handler
+@delayed_function(delay_seconds=1.0)
 def public_get_provider(event: dict, context: LambdaContext):  # noqa: ARG001 unused-argument
     """Return one provider's data
     :param event: Standard API Gateway event, API schema documented in the CDK ApiStack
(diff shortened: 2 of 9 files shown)


landonshumway-ia:
Dependency checks are currently failing. Those dependencies have been updated in the following PR: #1572.

backend/compact-connect/lambdas/python/provider-data-v1/handlers/public_lookup.py
```diff
@@ -113,6 +113,7 @@ def public_query_providers(event: dict, context: LambdaContext):  # noqa: ARG001
 
 
 @api_handler
+@delayed_function(delay_seconds=1.0)
```
ChiefStief, line 116: What does this do / why?

ChiefStief (approved):
Ok, approved. Just have the one question

backend/compact-connect/lambdas/python/provider-data-v1/handlers/public_lookup.py
landonshumway-ia: I'll post the description from the [method docs](https://github.com/csg-org/CompactConnect/blob/ec45a3b85c5b0d4525259b69b68fb7507b7f1e6e/backend/compact-connect/lambdas/python/common/cc_common/utils.py#L583-L587) since it explains it well:

```
Delay the result of the decorated function by the specified number of seconds.

This decorator ensures consistent response times for security-sensitive endpoints,
helping to prevent timing attacks by making all responses take the same amount of time
regardless of the execution path taken.
```
Essentially, we don't want users spamming this public endpoint trying to determine the provider ids for practitioners that do not have any privileges by tracking the response times based on if the DB returns a result or not. Just another measure to tighten security.

landonshumway-ia:

backend/compact-connect/docs/internal/api-specification/latest-oas30.json
```diff
@@ -2,7 +2,7 @@
   "openapi": "3.0.1",
```
jlkravitz, line 2: Looks good. My only suggestion is that we could maybe add some documentation as to _why_ we're deviating from the standard of "beta and prod should be identical." It felt simple enough to grok differences between test and beta/prod, but now we're adding deviations for beta specifically. I assume we have documentation somewhere about the different environments, so it might be nice to add some context there.

jlkravitz (commented):
one suggestion on documentation!

backend/compact-connect/docs/internal/api-specification/latest-oas30.json
landonshumway-ia: I've added a comment to each of the CDK API stacks explaining the purpose of this discrepancy. I scanned through our READMEs, and it looks like we currently don't have a section explicitly explaining the different environments. Let me know if you think we need to add additional documentation beyond what I've added.

jlkravitz (approved):

PR #1597, opened 2026-05-28
[Social Work] Support single and multi-state license types

landonshumway-ia:
The Social Work compact needs to support both single and multi-state license types. To support this, we are adding a new license field specific to their API, `licenseScope`, which will have a value of either `single-state` or `multi-state`. 

This change includes updating all of the relevant schemas and tests to account for this new field, and inserts the relevant license types specific for Social Work throughout the codebase.

As part of this, we needed to update the sort keys for our license related DynamoDB records to include the license scope, in order to distinguish the single state versus multi state records when storing these records in the provider DynamoDB table. For adverse actions/investigations, this means the frontend client will now need to specify the license scope as part of the request body when setting/lifting encumbrances and investigations (privilege related records will always have a 'licenseScope' of single-state, since a privilege record in the system is scoped to the state it is associated with).

In addition, the ingest flow now checks for additional conditions to trigger home state changes for providers. A home state change is not triggered until both the single and multi-state licenses have been uploaded for a jurisdiction, and the multi-state license is more recent than any other jurisdiction's multi-state license.

This logic also holds for generating privileges. We only generate privileges for multi-state licenses if the multi-state license is marked as compact eligible and there is an associated single-state license for the same jurisdiction/license type that is also compact eligible. Likewise, if the single-state license is marked as compact ineligible, than the associated multi-state license will also show as compact ineligible in the api response. 

### Testing List
- Code review

Closes #1585 

## Summary by CodeRabbit

* **New Features**
  * Introduced required licenseScope (single-state | multi-state) across APIs, events, storage and search; provider/license identities and public search now include scope and privilege rules respect paired single/multi-state logic. Removed prior "mostRecentLicenseForType" constraint from public searches.

* **Documentation**
  * CSV/JSON examples, onboarding, API schemas, design docs and Postman updated for licenseScope and Social Work license types.

* **Bug Fixes**
  * Bulk upload/ingest always enforce duplicate SSN + license type + licenseScope uniqueness.

* **Tests**
  * Fixtures, unit/integration tests and snapshots updated for scope and new license types.


backend/social-work-app/lambdas/python/provider-data-v1/handlers/investigation.py
```diff
@@ -56,6 +59,14 @@ def investigation_handler(event: dict, context: LambdaContext) -> dict:
         raise CCInvalidRequestException('Invalid endpoint requested')
 
 
+def _load_license_investigation_patch_body(event: dict) -> dict:
+    body = json.loads(event['body'])
+    try:
+        return LicenseInvestigationPatchRequestSchema().load(body)
+    except ValidationError as e:
+        raise CCInvalidRequestException(f'Invalid request body: {e.messages}') from e
```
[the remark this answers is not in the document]
landonshumway-ia: the api_handler decorator has a catch block specifically for this case, so we don't need to add a redundant check for it here.

backend/social-work-app/lambdas/python/search/handlers/public_search.py
```diff
@@ -261,9 +261,7 @@ def _build_public_license_search_body(*, compact: str, body: dict, cursor: dict
 
     search_after = cursor.get('search_after') if cursor else None
 
-    # For public search, we only match against licenses that are most recent
-    # for its license type. This value is set when the document is indexed into OpenSearch.
-    nested_must: list[dict] = [{'term': {'licenses.mostRecentLicenseForType': True}}]
+    nested_must: list[dict] = []
```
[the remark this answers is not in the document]
landonshumway-ia: The design for public searching is still under development. We will address public searching in a future PR once the requirements are well defined.

backend/cosmetology-app/lambdas/python/provider-data-v1/handlers/bulk_upload.py
```diff
@@ -162,21 +155,20 @@ def process_bulk_upload_file(
                     validated_license = schema.load(dict(compact=compact, jurisdiction=jurisdiction, **raw_license))
                     # verify that this ssn/licenseType combination has not been used previously in the same batch
```
ChiefStief, line 156: ~~Something definitely needs to be changed here, maybe just the comments and error message?~~

~~But duplicate ssn + licensetype shouldnt throw an error by itself now right?~~

Edit: Oh cosmetology

ChiefStief (changes requested):
Looks good to me. Well want to get those questions we spoke about earlier answered.

I believe we do actually have a bug with the rollback logic in that it can leave a provider with a different home state license than before the bad license data + rollback in this scenario and maybe a few others:

IN uploads single state expiring in 2022
OH uploads single state expiring in 2023

Home state is still IN.

OK uploads multi state and single state,
Ok gets rolled back.

Algorithm gets best license => OH 2023

So the home state would then be OH 2023 when it should be IN 2022.

The "best license" logic alone cant be applied here because its not how the home state is determined, date of upload is also relevant

ChiefStief (approved):
Ok looks good!

landonshumway-ia:

backend/social-work-app/docs/README.md
```diff
@@ -50,7 +50,8 @@ leave the field entirely empty. If some of your licenses are missing a required
 | homeAddressState* | State/province of provider's home address | String (max 100 chars) | IL |
 | homeAddressStreet1* | First line of provider's street address | String (max 100 chars) | 123 Main St |
 | licenseNumber* | License number | String (max 100 chars) | OT12345 |
-| licenseType* | Type of professional license. Types you provide must be associated with the compact you are uploading for. | One of: `cosmetologist`, `esthetician` | cosmetologist |
+| licenseType* | Type of professional license. Types you provide must be associated with the compact you are uploading for. | One of: `licensed clinical social worker`, `licensed master social worker`, `licensed bachelor social worker` | licensed clinical social worker |
+| licenseScope* | Whether the license is a single-state or multi-state license under the compact. Multi-state licenses in a provider's home jurisdiction can generate privileges to practice in other compact member states. | One of: `single-state`, `multi-state` | single-state |
```
jlkravitz, line 54: Can we add some docs (or a reference to an external source) that explains the mechanics behind these license scopes?

My sense is that a `multi-state` license is effectively a license that providers can get that gives them the privilege to practice in all compact member states. For social work, those privileges are auto-generated and there is no purchasing of privileges like with JCC. The multi-state license acquisition process happens outside of CompactConnect. That all true?

backend/social-work-app/docs/devops/STAFF_USER_MFA_RECOVERY.md
```diff
@@ -36,23 +36,20 @@ Before beginning this procedure, ensure you have:
    - Click on "Explore table items"
    - Search for an item where:
      - `pk` = `USER#{user_sub}` (where `user_sub` is the sub value from Cognito)
-     - `sk` = `COMPACT#{compact}` (where `compact` is the user's compact, e.g., "socw")
+     - `sk` = `COMPACT#{compact}` (where `compact` is the user's compact, e.g., "aslp", "coun", "octp")
```
jlkravitz: Why this change if we're in the social-work-app directory?

backend/cosmetology-app/docs/devops/STAFF_USER_MFA_RECOVERY.md
```diff
@@ -36,23 +36,20 @@ Before beginning this procedure, ensure you have:
    - Click on "Explore table items"
    - Search for an item where:
      - `pk` = `USER#{user_sub}` (where `user_sub` is the sub value from Cognito)
-     - `sk` = `COMPACT#{compact}` (where `compact` is the user's compact, e.g., "cosm")
+     - `sk` = `COMPACT#{compact}` (where `compact` is the user's compact, e.g., "aslp", "coun", "octp")
```
jlkravitz: Why this change if we're in the cosmetology app?

backend/social-work-app/docs/design/README.md
```diff
@@ -336,18 +336,14 @@ Because the list of privilege records for practitioners is dynamically determine
...
...practitioner has licenses uploaded by multiple states, the system must choose a **home state license** per license type. Unlike the JCC model, where practitioners register under a specific home state, this Social Work system does not currently allow the practitioner to specify which state is their current home state. It was determined that the home state license would be automatically selected based on which **multi-state** license was issued or renewed most recently, but only when that multi-state license has an active and eligible single-state license in the same jurisdiction and license type. Privileges are then generated from that multi-state home license only when that pairing exists: one privilege per compact member jurisdiction (other than the home jurisdiction) for that license type. If the most recent multi-state license lacks an active and eligible single-state license in the same jurisdiction and license type, no privileges are generated for that license type (there is no fallback to an older jurisdiction). This means that privileges are only returned from the API for a practitioner when the most recently issued or renewed multi-state license is the one with the pairing.
```
jlkravitz: Remind me: why does the system require a designated home jurisdiction in the first place? What purpose does it serve in this case when privileges are for "all compact members"?

backend/social-work-app/lambdas/python/common/cc_common/data_model/provider_record_util.py
```diff
@@ -567,40 +689,46 @@ def get_update_records_for_license(
...
         provider = self.get_provider_record().to_dict()
         licenses = []
         privileges = []
 
         if is_public_response:
-            # only include the most recent license for each license type in the public response
-            license_records = self.find_most_recent_licenses_for_each_license_type()
+            # only include the most recent multi-state license for each license type in the public response
+            license_records = self.find_most_recent_licenses_for_each_license_type(LicenseScopeEnum.MULTI_STATE)
         else:
```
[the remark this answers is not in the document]
landonshumway-ia: No, our requirements are to only show multi-state related license records for public search, not single-state.

backend/social-work-app/lambdas/python/common/tests/unit/test_provider_record_util.py
```diff
@@ -551,9 +986,151 @@ def test_find_best_license_complex_scenario(self):
...
+
+    def test_find_best_license_in_current_known_licenses_respects_type_filter_with_scope_precedence(self):
+        from cc_common.data_model.provider_record_util import ProviderUserRecords
+
+        provider_user_records = ProviderUserRecords(
+            self._make_provider_records(self._license_fixture_with_mixed_scopes())
+        )
+
+        best_license = provider_user_records.find_best_license_in_current_known_licenses(
+            license_type_abbreviation='lcsw'
```
jlkravitz: It seems this filter has no effect on the outcome (`best_license`) given the test data generated by `_license_fixture_with_mixed_scopes`, since there is only one multi-state license. I _think_ this might be a better test if the fixture included another (older) multi-state license of a different type; then search by that type to make sure it respects the filter?

backend/social-work-app/lambdas/python/provider-data-v1/tests/function/test_handlers/test_ingest.py
```diff
@@ -58,6 +64,57 @@ def _get_provider_via_api(self, provider_id: str) -> dict:
...
+            update
+            for update in provider_records.get_all_provider_update_records()
+            if update.updateType == UpdateCategory.HOME_JURISDICTION_CHANGE.value
+        ]
+        self.assertEqual(1, len(home_jurisdiction_updates), home_jurisdiction_updates)
+        provider_update = home_jurisdiction_updates[0]
+        self.assertEqual(former_home_jurisdiction, provider_update.previous['licenseJurisdiction'])
+        self.assertEqual(new_home_jurisdiction, provider_update.updatedValues['licenseJurisdiction'])
+
+    def _assert_provider_update_license_upload_other(
```
jlkravitz: nit: would this be a better name?
 
```suggestion
    def _assert_provider_update_license_upload_name_change(
```

backend/social-work-app/lambdas/python/provider-data-v1/tests/function/test_handlers/test_ingest.py
```diff
@@ -610,15 +686,16 @@ def test_multiple_license_types_different_jurisdictions(self):
         with open('../common/tests/resources/ingest/event-bridge-message.json') as f:
             message = json.load(f)
 
-        # Update the message to be for an esthetician license in 'ky' with a newer issuance date
+        # Update the message to be for an licensed master social worker license in 'ky' with a newer issuance date
```
jlkravitz: If it has a newer issuance date, why would step 3 be true (pasted below)? Wouldn't the provider data be updated, just like it was in the test above this one?

> 3. Verifies that both licenses are present and the provider data is NOT updated to the most recently issued license.

backend/social-work-app/lambdas/python/provider-data-v1/tests/function/test_handlers/test_ingest.py
```diff
@@ -723,85 +1044,740 @@ def test_same_license_types_different_jurisdictions_triggers_home_jurisdiction_c
...
-        # Verify provider record remains the same
-        provider_data = self._get_provider_via_api(provider_id)
-        self.assertEqual('oh', provider_data['licenseJurisdiction'])
-        self.assertEqual('Björk', provider_data['givenName'])
+    def test_home_jurisdiction_change_when_multi_state_ingested_before_paired_single_state(
+        self,
+    ):
+        """
+        OH multi-state may be processed before OH single-state when uploads are close together.
+        Home jurisdiction change must still fire when the single-state license completes the pair.
```
jlkravitz: Wait, I thought an earlier test ensured that a home jurisdiction changes only when the multi-state license comes after the single-state license? This seems to contradict that. What is defined as 'close together'? What's the logic for this ordering business rule?

backend/social-work-app/lambdas/python/provider-data-v1/tests/function/test_handlers/test_ingest.py
```diff
@@ -723,85 +1044,740 @@ def test_same_license_types_different_jurisdictions_triggers_home_jurisdiction_c
...
+        provider_records = self._provider_table.query(
+            Select='ALL_ATTRIBUTES',
+            KeyConditionExpression=Key('pk').eq(f'socw#PROVIDER#{message["detail"]["providerId"]}'),
+        )['Items']
+        multi_state_license = next(
+            record
+            for record in provider_records
+            if record['type'] == 'license' and record['licenseScope'] == 'multi-state'
+        )
+        self.assertEqual('B0608337260', multi_state_license['licenseNumber'])
```
jlkravitz, line 1729: Wait, so there's a validation error but the license is still uploaded?

backend/social-work-app/lambdas/python/disaster-recovery/tests/function/test_rollback_license_upload.py
```diff
@@ -541,6 +784,46 @@ def test_provider_skipped_if_license_updates_detected_after_end_of_time_window_w
         license_updates = provider_records.get_all_license_update_records()
         self.assertEqual(2, len(license_updates), 'License updates should still exist')
 
+    def test_provider_skipped_when_other_jurisdiction_home_changed_during_window(self):
+        """Providers with an in-window home jurisdiction change outside the rollback jurisdiction are skipped."""
```
jlkravitz: I'm a bit confused by what this is testing. What do you mean by 'skipped' in this context?

backend/social-work-app/lambdas/python/disaster-recovery/tests/function/test_rollback_license_upload.py
```diff
@@ -1011,6 +1305,40 @@ def test_transaction_failure_is_logged_and_provider_marked_as_failed(self):
             self.assertEqual(0, len(results_data['revertedProviderSummaries']))
             self.assertEqual(0, len(results_data['skippedProviderDetails']))
 
+    def test_orphaned_license_update_skipped_when_only_other_scope_license_exists(self):
+        """License updates must match jurisdiction, type, and scope; another scope is not sufficient."""
```
jlkravitz: Same here- a bit confused by what this means.

jlkravitz (changes requested):
Good stuff! Few comments/questions.

backend/social-work-app/docs/devops/STAFF_USER_MFA_RECOVERY.md
landonshumway-ia: Ah sorry, that was copy paste from when I corrected the original doc in the jcc folder. I'll fix that.

backend/social-work-app/docs/design/README.md
landonshumway-ia: The home state determines whether or not a practitioner is eligible for the privilege to practice in other states. For example, if you have two multi-state licenses, but one is marked ineligible, if your home state is the jurisdiction for the ineligible multi-state license, then you will not show up as eligible to practice in other states.

backend/cosmetology-app/docs/devops/STAFF_USER_MFA_RECOVERY.md
landonshumway-ia: Copy/paste error. I've corrected it.

backend/social-work-app/lambdas/python/provider-data-v1/tests/function/test_handlers/test_ingest.py
landonshumway-ia: In the test above this one, the licenses are from the same state. In this case, the license is from a new home state, but only one of the two license scopes has been uploaded, so it doesn't trigger a home state change or a provider record update. For social work, we only update the provider record data under the following conditions (referenced [from here](https://github.com/csg-org/CompactConnect/blob/e2f5255a8812326c809c7f7d522f7bf96a2162ea/backend/social-work-app/lambdas/python/provider-data-v1/handlers/ingest.py#L244-L249))
```
# Determine which license, if any, should populate the top-level provider record:
# - On a home jurisdiction change, use the new home multi-state license.
# - On the provider's first license upload, use the posted license.
# - Otherwise (no home change) only refresh provider data when the posted license is the best license
#   for the provider's current home jurisdiction.
# If none of the above conditions are met, do not update the provider record.
```

backend/social-work-app/lambdas/python/provider-data-v1/tests/function/test_handlers/test_ingest.py
landonshumway-ia: Here is the conditions that trigger a home state change (documented in the [ingest.py file](https://github.com/csg-org/CompactConnect/blob/e2f5255a8812326c809c7f7d522f7bf96a2162ea/backend/social-work-app/lambdas/python/provider-data-v1/handlers/ingest.py#L351-L358))

```
    A home jurisdiction change is triggered when all of the following are true:
    - There is an existing provider record (i.e. this is not the provider's first license upload).
    - The most recently issued/renewed multi-state license across all known licenses is from a different
      jurisdiction than the provider's current home jurisdiction.
    - That multi-state license has a paired single-state license of the same type in the same jurisdiction.

    Because the check operates on ``known_licenses`` (which already includes the license being ingested),
    it correctly handles either upload ordering: multi-state first then single-state, or vice versa.

```

The previous test was checking that the home state change does not trigger until the multi-state license is uploaded to complete the pairing with the single-state license. This is checking the opposite case where the multi-state license is uploaded first, and verifies that the home state change is not triggered until after the associated single-state license is uploaded.

backend/social-work-app/lambdas/python/provider-data-v1/tests/function/test_handlers/test_ingest.py
landonshumway-ia: Yes, the validation error notifies the state that they have uploaded a multi-state license as eligible but the single-state license is ineligible, which is a logical error. We can still process the record, because we ultimately can determine the final eligibility based on the state of the data on all of the license records, but we at least want to let the state know that there is a mismatch of eligibility between the single and multi-state licenses that they need to investigate and correct.

backend/social-work-app/lambdas/python/disaster-recovery/tests/function/test_rollback_license_upload.py
landonshumway-ia: The provider data is not rolled back by the license rollback process if their home state was changed during the rollback window. I will update the test name and description.

backend/social-work-app/docs/design/README.md
jlkravitz: Ah got it- It's sort of the 'governing state', so to speak, of the provider. That state has the authority to deny or approve the provider's privilege to practice in other member compact states.

backend/social-work-app/lambdas/python/disaster-recovery/tests/function/test_rollback_license_upload.py
landonshumway-ia: I have renamed this and updated the description to hopefully make the edge case easier to understand.

backend/social-work-app/docs/design/README.md
landonshumway-ia: Correct. I have added some details to this document to hopefully make it more clear.

backend/social-work-app/docs/README.md
landonshumway-ia: Yes, and I've added a `License Scopes` section to the document to hopefully clarify the difference between the single state and multi state licenses.

backend/social-work-app/docs/design/README.md
```diff
@@ -330,9 +330,22 @@ operational needs and audit requirements for healthcare provider licensing acros
 ## Multi-State License Model / Privilege Generation
 [Back to top](#backend-design)
 
-Privileges are authorizations that allow licensed providers to practice their profession in jurisdictions other than their home state. TheSocial WorkCompact follows a multi-state licensure model, where privileges are automatically granted to a licensee as a result of having a multi-state license in their home state. The list of jurisdictions where privileges are granted is determined by which states have onboarded into the CompactConnect system for theSocial WorkCompact.
+Privileges are authorizations that allow licensed providers to practice their profession in jurisdictions other than their home state. The Social Work Compact follows a multi-state licensure model, where privileges are automatically granted to a licensee as a result of having a multi-state license in their home state. The list of jurisdictions where privileges are granted is determined by which states have onboarded into the CompactConnect system for theSocial WorkCompact.
```
jlkravitz: missing some spaces: "theSocial WorkCompact"

backend/social-work-app/lambdas/python/provider-data-v1/tests/function/test_handlers/test_ingest.py
jlkravitz: Got it. I guess I still wonder: What does 'close together' mean in the context of this comment? It seems like the upload of the multi-state license could happen arbitrarily in the future, no?

backend/social-work-app/lambdas/python/provider-data-v1/tests/function/test_handlers/test_ingest.py
jlkravitz: Ah, got it. Thank you!

backend/social-work-app/lambdas/python/provider-data-v1/tests/function/test_handlers/test_ingest.py
landonshumway-ia: Yeah, I removed that 'close together' comment. The gap in the upload times between the two licenses is irrelevant.

backend/social-work-app/lambdas/python/common/tests/unit/test_provider_record_util.py
landonshumway-ia: As I was working through this, I found that the current implementation for 'find_best_license_in_current_known_licenses' was not taking into account multi-state licenses that had a matching single-state license. I have updated the implementation and added some more test cases to cover these conditional branches.

landonshumway-ia:
@jlkravitz This is now ready for another review. Thanks

backend/social-work-app/lambdas/python/common/tests/unit/test_provider_record_util.py
jlkravitz: Can you clarify what the prior behavior was?

landonshumway-ia: The prior behavior was copied from the Cosmetology multi-state implementation, which only considers what the most recent license is. Their implementation does not have any concept of a 'licenseScope', so it does not check for a single and multi state license pairing.

jlkravitz (approved):

PR #1618, opened 2026-06-01
export public lambda arns to avoid CDK deadly embrace lock in beta deployment

landonshumway-ia:
We have removed the public API search endpoints from beta (see #1591). The deployment for that change failed due to the 'deadly embrace' issue in CDK where a stack attempts to remove an unneeded export, but CloudFormation can't allow that deletion until the consuming stack of that export has been updated. This adds dummy exports for those lambdas to fix the deployment and allow all the stacks to be updated.

## Summary by CodeRabbit

* **Chores**
  * Updated backend stack configurations to improve deployment stability and reliability across services.


ChiefStief (approved):
Looks good!

jlkravitz (approved):

PR #1624, opened 2026-06-03
[Cosmetology] Add 'surrender of privilege' encumbrance type

landonshumway-ia:
During a demo of the Cosmetology features, it was determined that there needs to be a specific encumbrance type to be used when encumbering privileges. This adds the desired discipline action.

### Testing List
- Code review

Closes #1617 

## Summary by CodeRabbit

* **New Features**
  * Added "surrender of privilege" as an allowed encumbrance type across APIs, request validation, UI discipline options, and English/Spanish translations.
* **Tests**
  * Updated JSON schema snapshots to reflect the expanded encumbranceType enum so requests including the new value validate in tests.

### DIFF
--- a/webroot/src/components/LicenseCard/LicenseCard.ts
+++ b/webroot/src/components/LicenseCard/LicenseCard.ts
@@ -261,6 +261,10 @@ class LicenseCard extends mixins(MixinForm) {
             const includeList = ['suspension', 'revocation', 'surrender of license'];
 
             options = options.filter((option) => includeList.includes(option.value));
+        } else {
+            const excludeList = ['surrender of privilege'];
+
+            options = options.filter((option) => !excludeList.includes(option.value));
         }
 
         options.unshift({
--- a/backend/cosmetology-app/docs/internal/api-specification/latest-oas30.json
+++ b/backend/cosmetology-app/docs/internal/api-specification/latest-oas30.json
@@ -2334,7 +2334,8 @@
                 "enum": [
                   "suspension",
                   "revocation",
-                  "surrender of license"
+                  "surrender of license",
+                  "surrender of privilege"
                 ]
               }
             },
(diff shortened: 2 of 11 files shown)


ChiefStief (approved):
:+1:

jsandoval81 (approved):

landonshumway-ia:
@jlkravitz This PR is now ready for your review. Thanks

webroot/src/components/LicenseCard/LicenseCard.ts
```diff
@@ -261,6 +261,10 @@ class LicenseCard extends mixins(MixinForm) {
             const includeList = ['suspension', 'revocation', 'surrender of license'];
 
             options = options.filter((option) => includeList.includes(option.value));
+        } else {
+            const excludeList = ['surrender of privilege'];
```
jlkravitz, line 265: small preference for this being an `includeList` here and in `PrivilegeCard.ts`. thoughts?

jlkravitz (commented):
one nit!

webroot/src/components/LicenseCard/LicenseCard.ts
jsandoval81: @jlkravitz Since JCC is using the superset of values and only has this 1 new need for an exclusion, it felt simpler to go with an exclusion in that mode to start. Also without introducing the need to update the list here every time a new item is added to the master list. However, I don't feel strongly. If after considering the initial reasoning you want us to convert this to an `includeList` with 15 or so values for now, just let us know.

jlkravitz: @jsandoval81 Seems reasonable. I think my concern stems from the fact that we are about to add more compacts, and it may be difficult to track new encumbrance types that should be added to this exclusion list. That said, I'm good to keep as-is and monitor.

jlkravitz (approved):

PR #1637, opened 2026-06-10
Set validation error times to current time to avoid race condition

landonshumway-ia:
There is a race condition that sometimes occurs when a low volume of invalid licenses are uploaded in a CSV file. We have a error email reporter that sends out license upload failure emails every 15 minutes, and this reporting process looks at the last 15 minutes of events in the DB to see if any upload errors occurred. While beta testing, a state uploaded a single invalid license record. The upload failure event was placed on the data event bus which placed a message on a SQS queue for processing. That queue has a batch size of 10 and it will wait for up to 5 minutes to process messages if the batch size is not met. The one upload failure notification was sitting in the queue long enough that by the time it was stored in the DB, the notification error reporter had already completed its scan for the 15 minutes and never detected the error. This race condition exists in all three compact codebases (JCC, Cosm, and SW). This change reduces the batch window for the data event processor queue from 5 minutes to 5 seconds to address this.

This can also occur if there is a large CSV upload and all of the errors are toward the end of the file. Previously, we were setting the same time stamp for all row failures, so if a large upload took 15 minutes to process, and the errors were toward the end of the file, they wouldn't be stored in the database until after the ingest error reporter had scanned the 15 minute window, so the errors wouldn't be reported. This updates the timestamp to use the actual time that the row was processed and the error was detected. 

### Testing List
- Code review

Closes #1633

### DIFF
--- a/backend/social-work-app/stacks/persistent_stack/data_event_table.py
+++ b/backend/social-work-app/stacks/persistent_stack/data_event_table.py
@@ -104,7 +104,7 @@ class DataEventTable(Table):
             process_function=self.event_handler,
             visibility_timeout=Duration.minutes(1),
             retention_period=Duration.hours(1),
-            max_batching_window=Duration.minutes(5),
+            max_batching_window=Duration.seconds(5),
             max_receive_count=3,
             batch_size=10,
             encryption_key=encryption_key,
--- a/backend/compact-connect/lambdas/python/common/tests/unit/test_utils.py
+++ b/backend/compact-connect/lambdas/python/common/tests/unit/test_utils.py
@@ -5,7 +5,7 @@ from tests import TstLambdas
 
 
 class TestUtils(TstLambdas):
-    @patch('cc_common.config._Config.license_preprocessing_queue')
+    @patch('cc_common.utils.config.license_preprocessing_queue')
     def test_send_licenses_to_preprocessing_queue_handles_failures(self, mock_preprocessing_queue):
         from cc_common.utils import send_licenses_to_preprocessing_queue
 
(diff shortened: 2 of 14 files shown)


ChiefStief (approved):
Looks good!

backend/social-work-app/stacks/persistent_stack/data_event_table.py
```diff
@@ -104,7 +104,7 @@ def __init__(
             process_function=self.event_handler,
             visibility_timeout=Duration.minutes(1),
             retention_period=Duration.hours(1),
-            max_batching_window=Duration.minutes(5),
+            max_batching_window=Duration.seconds(5),
```
jlkravitz, line 107: Any considerations here with respect to price, performance, etc? (Or, why did we choose 5 minutes before?)

jlkravitz (commented):
One question, but otherwise- great catch!

backend/social-work-app/stacks/persistent_stack/data_event_table.py
landonshumway-ia: There shouldn't be a major impact in price or performance because the batch size is so small (10 records) which with normal usage will be filled with a batch license upload within 5 seconds. 5 minutes has been our go to standard for most of the SQS queues up to this point as a baseline, but there wasn't any hard rule for that. We will likely see an increase of lambda invocations during low tide since other events such as encumbrances will send one message which will trigger the lambda after 5 seconds of sitting in the queue, but the cost should be negligible and won't impact performance.

jlkravitz: Got it-thanks!

jlkravitz (approved):

PR #1638, opened 2026-06-10
Add backend check to prevent state admin from granting compact permission

landonshumway-ia:
When the readSSN permission was added during development, the check for how it is granted at the compact level was not included in the utility function which verifies the caller's permission scopes to ensure a state admin cannot add permissions at the compact level. The UI prevents this, but in theory a state admin could call the API directly outside of the app UI, so this adds that auth check to prevent that.


jlkravitz (approved):
LGTM! @isabeleliassen This is good to merge. Once merged, I will do a prod deploy since this is a hotfix.

PR #1642, opened 2026-06-12
Prevent cross jurisdiction permission patching

landonshumway-ia:
The patch_user endpoint is the only staff-user mutation handler that does NOT verify the target user is within the calling jurisdiction admin's purview. All other staff endpoints that modify staff user permissions fetch the target and require the caller to have admin permissions for the target user's jurisdiction. patch_user relies only on `collect_and_authorize_changes`, which validates the requested changes against the caller's scopes but never loads the target to confirm the caller may administer that specific user.

This adds that check so that if the caller does not have the needed jurisdiction admin scopes for the target user, the endpoint returns a 404 not found response.

### Testing List
- Code review

Closes #1640 

## Summary by CodeRabbit

* **Bug Fixes**
  * Tightened jurisdiction-based access for admin user edits: admins can only modify users who share at least one allowed jurisdiction; attempts outside permitted jurisdictions now return 404. Caller activity is verified earlier in the request flow.

* **Tests**
  * Added functional tests to cover jurisdiction-scoped access control and the 404 behavior when editing users outside an admin’s allowed jurisdictions.

### DIFF
--- a/backend/compact-connect/lambdas/python/staff-users/handlers/users.py
+++ b/backend/compact-connect/lambdas/python/staff-users/handlers/users.py
@@ -98,9 +98,16 @@ def patch_user(event: dict, context: LambdaContext):  # noqa: ARG001 unused-argu
     user_id = event['pathParameters']['userId']
     scopes = get_event_scopes(event)
     path_compact = event['pathParameters']['compact']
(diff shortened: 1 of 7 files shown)


landonshumway-ia:

jlkravitz (approved):

PR #1705, opened 2026-07-02
[Social Work] - Enhance technical documentation

landonshumway-ia:
We received a request from the chair of the Social Work commission for further technical documentation. It was determined with @KbisonCSG to deliver technical documentation for the following elements of the request:

- The Commission unique-identifier structure - RID and CUID identity management;
- All-licensee data ingestion, including single-state licensees
- MSL issuance and home-state confirmation, and remote-state Multistate Authorization to Practice
- deactivation effects from adverse action processing
- current significant investigative information
- reissuance when a licensee changes home state, and one-home-state / one-active-MSL enforcement
- automated notifications to the Commission and Member States
- authenticated business-record auditability

This enhances our existing technical design documentation and ERD to cover these aspects of the system.

Closes #1702

## Summary by CodeRabbit

* **Documentation**
  * Expanded and realigned backend design documentation across ingest, data modeling, privilege generation (home-state rules), notifications, investigations, search indexing, and audit logging, including a reordered/extended table of contents and updated jurisdiction mapping references.
  * Added clarification for single-state licensee ingestion behavior, eligibility mismatches, and re-upload history handling.
  * Updated provider data model visuals and terminology, replacing the internal provider identifier definition and introducing an optional public-facing identifier concept (planned for CUID/UUID4).


backend/social-work-app/docs/design/provider-data-model-erd.mmd
```diff
@@ -8,7 +8,8 @@ erDiagram
         string sk "{compact}#PROVIDER"
         string type "provider"
         datetime dateOfUpdate
-        string providerId
+        uuid providerId "RID, UUID4, internal"
+        uuid publicCompactIdentifier "optional, CUID, UUID4, public"
```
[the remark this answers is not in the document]
landonshumway-ia: I have updated the type to string since that is the actual type used when storing the id in DynamoDB.

backend/social-work-app/docs/design/README.md
```diff
@@ -297,6 +324,39 @@ License-related sort keys use a shared suffix of the form
...
+
+Once implemented, the CUID will be stored as the optional `publicCompactIdentifier` field on the top-level
+[Provider Record](#record-types-in-detail). Generation is planned as part of the license ingest flow in
+[ingest.py](../../lambdas/python/provider-data-v1/handlers/ingest.py): whenever a license a multi-state license is uploaded for a practitioner, the ingest handler will check whether `publicCompactIdentifier` is already set on the practitioner's provider record. If it is not yet set, a new CUID will be generated at that point and written to the provider record. Because the check is against the existing value on the provider record, the CUID is generated exactly once and is never regenerated on subsequent multi-state license uploads, including after a [home state change](#home-state-changes). Because it will live on the top-level provider record rather than on an individual license, the CUID will also be added as a top-level field in the [OpenSearch index mapping](#index-mapping), making practitioners searchable by CUID (see [Advanced Data Search](#advanced-data-search)).
```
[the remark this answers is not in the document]
landonshumway-ia: I just clarified with the client that the CUID will be generated whenever any multi-state license is uploaded in to the system for a practitioner.

backend/social-work-app/docs/design/README.md
```diff
@@ -96,16 +101,36 @@ same KMS key as the SSN table to invoke the license preprocessor Lambda function
...
+
+If a jurisdiction uploads a multi-state license as compact-eligible while its paired single-state license in the same jurisdiction is already ineligible (inactive, expired, or encumbered), the multi-state license is **still persisted** rather than rejected, and a
+`license.validation-error` event is published so the uploading jurisdiction can be notified of the mismatch (see
+`_check_for_multi_state_single_state_eligibility_validation_error` in
+[ingest.py](../../lambdas/python/provider-data-v1/handlers/ingest.py)). This ensures that licensee data is always
+accepted and ingested even in cases where eligibility for compact privileges cannot yet be fully confirmed.
+
+Re-uploading that same license (determined by providerId, jurisdiction, licenseType, and licenseScope) updates the existing record in place (rather than creating a duplicate), and, when relevant fields changed, creates a `licenseUpdate` history record categorized as a `renewal` (when the expiration date moved forward — i.e., the jurisdiction reissued the license for a new term), a `deactivation` (when the jurisdiction reports the license as inactive), or `other` (for any other changed field).
```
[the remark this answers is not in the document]
landonshumway-ia: I have added details about these update categories to the respective encumbrance and investigation sections where the information is more relevant.

landonshumway-ia:
@jlkravitz This is now ready for your review. Thanks

backend/social-work-app/docs/design/users-arch-diagram.pdf
jlkravitz, line 1: I can't tell if this was deleted or modified? If the latter, should we update the date or add an "Updated Date" so it's clear that it's been more recently updated than February 2025?

backend/social-work-app/docs/design/README.md
```diff
@@ -96,16 +102,36 @@ same KMS key as the SSN table to invoke the license preprocessor Lambda function
 This architecture ensures that SSN data is protected throughout the ingest process while still allowing the system to
 associate licenses with the correct providers across jurisdictions.
 
-### Asynchronous validation feedback
-Asynchronous validation feedback for boards to review is not yet implemented.
+### All-Licensee Ingestion, Including Single-State Licensees
+
+The ingest pipeline described above processes **every uploaded licensee record identically**, regardless of the
+license's `licenseScope` (`single-state` or `multi-state`, see [License Scopes](#license-scopes)). A jurisdiction is
+not required to upload only its multi-state licensees; it is expected to upload data for **all** of its eligible
```
jlkravitz: This sort of implies jurisdictions do not _have_ to upload its single-state licenses, but I think you are saying they must?  This might be more clearly stated in 'must', 'should', and 'may' terms.

> A jurisdiction must upload data for all of its eligible providers, including all single-state and multi-state licenses associated with each provider.

Maybe? I changed to use the term 'providers' because reading it quickly led me to read it as licenses, rather than licensees.

backend/social-work-app/docs/design/README.md
```diff
@@ -297,6 +326,38 @@ License-related sort keys use a shared suffix of the form
...
+  portal.
+- **Visible to staff users**: although hidden from the public portal, the RID is visible to staff users through the staff user search.
+
+#### Compact Unique Identifier (CUID) — Planned
+
+The CUID is the Compact's persistent, public-facing identifier for individuals who hold multistate
+authorization (i.e. at least one generated privilege — see [Multi-State License Model / Privilege
+Generation](#multi-state-license-model--privilege-generation)). Per Commission rule, the CUID will be:
+- **Issued only when a state uploads a Multistate License** — i.e. generated the first time a practitioner has a multi-state license uploaded into the system, rather than at first license ingestion in general (contrast with the RID, which is assigned on first license ingestion regardless of scope).
+- **Public-facing**, unlike the RID, and required by Commission rule to be shown in public lookup.
```
jlkravitz:
```suggestion
- **Public-facing**, unlike the RID, required by Commission rule to be shown in public lookup.
```

backend/social-work-app/docs/design/README.md
```diff
@@ -297,6 +326,38 @@ License-related sort keys use a shared suffix of the form
...
+- **Person-based, not license-based**, and persistent for mobility: the CUID will remain the same even if the practitioner's home state later changes under Compact rules (see [Home State Changes](#home-state-changes)).
```
jlkravitz: Isn't this true of the RID, too?

backend/social-work-app/docs/design/README.md
```diff
@@ -297,6 +326,38 @@ License-related sort keys use a shared suffix of the form
    - `licenseScope` identifies the target license when `investigationAgainst` is `license`
    - Privilege-targeted investigations always use `licenseScope` `single-state`, since privileges are scoped to a particular state.
 
+### Commission Unique Identifiers: RID and CUID
```
jlkravitz, line 326: Despite being part of some of these discussions, I'm still a bit confused by what the CUID provides that the RID does not, except for it only being generated when the provider is eligible to practice in compact states. But why could the RID just be 'revealed' publicly at the same point when the CUID would otherwise be generated?

backend/social-work-app/docs/design/README.md
```diff
@@ -373,10 +434,39 @@ type (for example, an Ohio LCSW single-state license and an Ohio LCSW multi-stat
 How scopes are represented in DynamoDB sort keys and related record types is documented under
 [License Composite Identifier](#license-composite-identifier) in the Data Model section.
 
+### MSL Issuance and Home-State Confirmation
```
jlkravitz: Can we spell this out for the header: "Multi-state License"

backend/social-work-app/docs/design/README.md
```diff
@@ -373,10 +434,39 @@ type (for example, an Ohio LCSW single-state license and an Ohio LCSW multi-stat
...
+Privileges are generated from a multi-state home license: one privilege per compact member jurisdiction (other than the home jurisdiction) for that license type. If the most recent multi-state license is ineligible or its associated single-state license is ineligible, no privileges are generated for that license type (there is no fallback to an older jurisdiction). This means that privileges are only returned from the API for a practitioner when the most recently issued or renewed multi-state license is the one with the pairing.
```
jlkravitz: Couple small suggested edits for clarity.

```suggestion
Privileges are generated from a multi-state home license: one privilege per compact member jurisdiction (other than the home jurisdiction) for that license type. If the most recent multi-state license is ineligible or its associated single-state license is ineligible, no privileges are generated for that license type (there is no fallback to an older multi-state license from another jurisdiction). This means that privileges are only returned from the API for a practitioner when the most recently issued or renewed multi-state license is eligible and paired with an eligible single-state license.
```

backend/social-work-app/docs/design/README.md
```diff
@@ -386,6 +476,32 @@ Each privilege’s status (active/inactive, under investigation) is derived from
 that privilege jurisdiction and license type. License-targeted adverse actions and investigations additionally specify
 which `licenseScope` they apply to.
 
+### Home State Changes
+
+A home jurisdiction change occurs when, after a license is ingested, the most recent multi-state license with a paired single-state license is from a **different** jurisdiction than the practitioner's current home state on the provider record. When this is detected:
```
jlkravitz:
```suggestion
A provider's home jurisdiction change occurs when, after a license is ingested, the most recent multi-state license (across all license types) with a paired single-state license is from a **different** jurisdiction than the practitioner's current home state on the provider record. When this is detected:
```

backend/social-work-app/docs/design/README.md
```diff
@@ -386,6 +476,32 @@ Each privilege’s status (active/inactive, under investigation) is derived from
...
+
+1. A `provider.homeStateChange` event is published to the data event bus (see [Notifications](#notifications)), which results in an email notification to the **former** home jurisdiction only.
+2. The practitioner's top-level provider record is rebuilt using the data from the new home license (name, contact information, `licenseJurisdiction`, etc.).
+3. A `providerUpdate` history record is written with `updateType: homeJurisdictionChange`, preserving the previous provider-record values.
+4. Because privileges are not stored and are instead generated fresh on every read (see
+   [Privilege Runtime Generation](#privilege-runtime-generation-for-multi-state-licenses)), privileges tied to the
+   **old** home jurisdiction stop being generated as soon as the change takes effect — there is no persisted privilege
+   record to separately deactivate, and no grace period.
```
jlkravitz: Hmm, is this 100% true? Your earlier note mentions the (rare) case where a provider can have multiple licenses with multiple home states, such that the privilege-generation home-state for a particular license type is different from the provider's saved home state.

Maybe we mean to say here, "...privileges tied to the **old** home jurisdiction for the license type that triggered the provider home jurisdiction change stop being generated as soon..."

(That is wordy, there's probably a better phrasing...)

backend/social-work-app/docs/design/README.md
```diff
@@ -386,6 +476,32 @@ Each privilege’s status (active/inactive, under investigation) is derived from
...
+
+**One-home-state-per-license-type enforcement:** For a given license type, the home MSL is always the single most
+recently issued/renewed **paired** multi-state license (sorted by `dateOfRenewal`, falling back to `dateOfIssuance`,
+with `dateOfIssuance` as a final tiebreaker). Notably, this selection does **not** consider whether the candidate
```
jlkravitz: > (sorted by `dateOfRenewal`, falling back to `dateOfIssuance`, with `dateOfIssuance` as a final tiebreaker)

Just confirming this is right– `dateOfIssuance` listed twice?

backend/social-work-app/docs/design/README.md
```diff
@@ -394,10 +510,222 @@ The privilege system is built around several core concepts:
...
+
+The system supports encumbering (taking adverse action against) both licenses and privileges. Encumbrance processing is split across two layers:
+
+1. **Synchronous data writes**: The staff admin API (`encumbrance.py`) validates the request and, through the data
+   access layer, writes an `adverseAction` record and updates the relevant `encumberedStatus` field(s) in DynamoDB.
+2. **Asynchronous notifications**: Once the write succeeds, an event is published to the data event bus. The
+   corresponding data-events handler (`encumbrance_events.py`) only sends email notifications to affected states
+   (see [Notifications](#notifications)) — it does not perform any further data mutation.
+
+`encumberedStatus` is the only value actually persisted by an encumbrance action. As described in
```
jlkravitz: Do we mean to say... 
```suggestion
`encumberedStatus` is the only value actually persisted by an encumbrance action in the target license record. As described in
```

Or is it the provider record? Since privileges don't exist in the DB, where does `encumberedStatus` live for those?

backend/social-work-app/docs/design/README.md
```diff
@@ -394,10 +510,222 @@ The privilege system is built around several core concepts:
...
+**derived** from `encumberedStatus` (along with expiration and jurisdiction-reported status) at load time, rather
+than being written directly. Encumbrance and lift actions each write a categorized `licenseUpdate` history record
+with `updateType: encumbrance` or `lifting_encumbrance`, respectively.
+
+The following four scenarios describe how encumbering a license or privilege affects deactivation, depending on what
+is targeted:
+
+| Scenario | Effect on generated privileges |
+|---|---|
+| Home-state **single-state** license encumbered | The paired home multi-state license is displayed as ineligible (its eligibility is inherited from the single-state license, see [Privilege Runtime Generation](#privilege-runtime-generation-for-multi-state-licenses)), so **all** privileges generated for that license type are suppressed |
```
jlkravitz, line 570: To clarify, if the single-state license is encumbered, _both_ the multi-state and single-state licenses are marked as 'ineligible'. Why not just mark the single-state license as ineligible and rely on the privilege generation logic that requires both to be eligible?

backend/social-work-app/docs/design/README.md
```diff
@@ -394,10 +510,222 @@ The privilege system is built around several core concepts:
...
+| Home-state **multi-state** license encumbered | The home multi-state license itself becomes ineligible. Because privilege generation requires **both** the home multi-state license and its paired single-state license to be eligible, the still-eligible single-state license cannot rescue privilege generation — **all** privileges for that license type are suppressed, the same as above |
+| **Remote** (non-home) state license encumbered | **No effect** on generated privileges. This is true regardless of whether the remote license's `licenseScope` is `single-state` or `multi-state` — both scopes are encumbered through the identical code path, so behavior is the same. The encumbrance is scoped to that one license record (and it flags the provider as encumbered compact-wide); it only starts to matter for privilege generation if that license later becomes the practitioner's home license |
```
jlkravitz: To clarify, if I have a MSL-SSL pairing in two states (say, OH and KY), OH is my home state (more recent MSL), and my KY license is encumbered, I can still get privileges to practice in the compact states?

Would I still get a privilege to practice in KY? Would that not just override the encumbrance that was placed on my KY license?

jlkravitz (commented):
few questions/clarifying questions!

backend/social-work-app/docs/design/users-arch-diagram.pdf
landonshumway-ia: It was deleted. This file was copied over from JCC, which included an illustration of the provider user pool which is not applicable to Social Work or Cosmetology.

backend/social-work-app/docs/design/README.md
landonshumway-ia: That was my argument as well, but apparently the commission is insistent that they are legally required to have a public id distinct from an internal id. @KbisonCSG has informed us that the commission will almost certainly be implementing this feature despite alternative options to only make the practitioner visible in the public search if they are able to practice in multiple states, so this is our proposed solution for supporting a CUID.

backend/social-work-app/docs/design/README.md
landonshumway-ia: That is how we do it. We do not mark the multi-state license as ineligible in the DB, only display it as ineligible: 

> The paired home multi-state license is **displayed** as ineligible

backend/social-work-app/docs/design/README.md
landonshumway-ia: So with the way the system is coded today, yes the system would display an active privilege for KY as a result of the eligible OH license, even if the KY license of the same type is encumbered.

This is a great question which we will need to clarify with @KbisonCSG. @KbisonCSG - could you verify what the behavior should be in this case?

KbisonCSG: @landonshumway-ia & @jlkravitz Technically, yes since their eligibility is only based on encumberance of the home state license (OH) in this example. However, I would say this would be a very niche case as i dont see a reason for someone to keep an SSL in a state they are elgible to work in through the compact anyways, unless it is just leftover for now while they are moving between home states. In that case, I also think it would be very rare for another state to issue a license to someone with an encumbered licese in another state.

If for some reason a SW additionally had an SSL in KY and it was encumbered, my thinking is that licensing board would then decide to revoke their privlege, which they could then go in and do on a case-by-case basis, but it is possible that they might want the data system to do so automatically? If possible, I can ask the admins tommorrow if that is something they would like to have happen.

backend/social-work-app/docs/design/README.md
jlkravitz: Ah, got it. Thanks!

backend/social-work-app/docs/design/README.md
jlkravitz: You and Landon are the SMEs here, so I'll defer to your read on complexity/feasibility/frequency/etc. In my experience, though, rare cases like this tend to surface more often than expected as systems scale, so I'd lean toward scoping rules that handle them explicitly rather than leaving them to admins to handle on a case-by-case basis.

landonshumway-ia: @jlkravitz - to @KbisonCSG's point, in this particular edge case the state admin for KY could encumber the individual privilege record in KY if needed, so we do have a path forward if it does occur.  

That being said @KbisonCSG, it would be good to clarify this behavior with the admins if possible so we can automate that deactivation now if needed before the system is live.

backend/social-work-app/docs/design/README.md
landonshumway-ia: It is awkwardly worded. The intent was to call out that state that if the `dateOfRenewal` is the same for both licenses, the `dateOfIssuance` is the fallback and tiebreaker. I'll find a better way to reword this.

backend/social-work-app/docs/design/README.md
landonshumway-ia: encumberedStatus is specific to license records in this case, since there is a persistent record in the DB. The encumberedStatus of privileges are determined by the adverse action records associated for that particular state/license type, which are persisted in the DB. I will add this detail to the document.

backend/social-work-app/docs/design/README.md
landonshumway-ia: I have reworded this to hopefully clarify what the actual behavior is.

backend/social-work-app/docs/design/README.md
landonshumway-ia: I have reworded this to make it more clear that we expect states to upload both license types for all of the providers they intend to upload information for.

backend/social-work-app/docs/design/README.md
```diff
@@ -105,12 +105,9 @@ associate licenses with the correct providers across jurisdictions.
...
 
 The ingest pipeline described above processes **every uploaded licensee record identically**, regardless of the
-license's `licenseScope` (`single-state` or `multi-state`, see [License Scopes](#license-scopes)). A jurisdiction is
-not required to upload only its multi-state licensees; it is expected to upload data for **all** of its eligible
-licensees, including practitioners who hold only a **single-state** license and no multi-state license.
+license's `licenseScope` (`single-state` or `multi-state`, see [License Scopes](#license-scopes)). It is expected that jurisdictions will upload all single-state and multi-state licenses associated with each provider that they intend to upload data for.
 
-Single-state-only licensees are fully ingested and stored as `license` records, and appear in the practitioner's
-provider record like any other license. However, because privilege generation requires a **paired** multi-state and
+Single-state-only licensees are fully ingested and stored as `license` records. However, because privilege generation requires a **paired** multi-state and
```
jlkravitz: licenses here, not licensees, right?

jlkravitz (commented):
one nit!

backend/social-work-app/docs/design/README.md
jlkravitz: Oh wait, did you mean something like "Single-state licenses are fully ingested and stored as `license` records, with or without a paired multi-state license."?

backend/social-work-app/docs/design/README.md
landonshumway-ia: @jlkravitz I have received feedback on this from @KbisonCSG and added the details in this commit https://github.com/csg-org/CompactConnect/pull/1705/changes/6b3873e0334ee379ded3c461802ecc4ac9aedbd7

backend/social-work-app/docs/design/README.md
landonshumway-ia: I've reworded this to hopefully make it a little more clear.

backend/social-work-app/docs/design/README.md
```diff
@@ -787,6 +819,29 @@ POST /v1/compacts/{compact}/providers/search
...
 Returns one result row per indexed document (one per license). Each hit is a full provider-shaped document for that license row (including the single license in `licenses` and generated privileges as applicable).
 
+#### Public Provider Search
+```
+POST /v1/public/compacts/{compact}/providers/query
+```
+
+An unauthenticated endpoint (implemented in
+[public_search.py](../../lambdas/python/search/handlers/public_search.py)) that allows members of the public to
+search for practitioners by license number, jurisdiction, family name, and/or given name. Like Provider Search, it
```
jlkravitz: Do you mean "Like the non-public Provider Search endpoint..."?

backend/social-work-app/docs/design/README.md
```diff
@@ -787,6 +819,29 @@ POST /v1/compacts/{compact}/providers/search
...
+The public query filters documents down to a practitioner's home multi-state license per license type
+(`licenses.licenseScope: multi-state` and `licenses.mostRecentLicenseForType: true`), consistent with the fact that
+only a home multi-state license can currently generate privileges.
+
+**Planned:** once the CUID is implemented (see [Commission Unique Identifiers: RID and
+CUID](#commission-unique-identifiers-rid-and-cuid)), the public query will add an additional filter requiring the
+top-level `publicCompactIdentifier` field to be populated (i.e. non-null/non-empty) on the document. Since the CUID
+is only generated once a practitioner has a paired single-state/multi-state license on file (see [Compact Unique
+Identifier (CUID)](#compact-unique-identifier-cuid--planned)), this filter has the effect of excluding practitioners
+who have not yet met that pairing requirement from public search results entirely. This ensures the public can only find and reference practitioners using a
```
jlkravitz: To confirm my understanding: All practitioners will live in OpenSearch, but the public API will _require_ a CUID provided, such that the public search is really only usable for searching for a specific provider's multi-state licenses?

jlkravitz (commented):
Changes look good- two small comments/questions.

backend/social-work-app/docs/design/README.md
landonshumway-ia: Yes, I've reworded that section for clarity.

backend/social-work-app/docs/design/README.md
landonshumway-ia: Correct, the public search will only match against practitioners that have both a single and multi-state license pairing. Until both of those records are uploaded, we will not generate a CUID and the practitioner will not be visible in the public search. I have reworded this section slightly to hopefully make this clearer.

jlkravitz (approved):

PR #1726, opened 2026-07-09
[JCC] - Add 'previousSSN' field to license upload process for correcting incorrect SSN uploads

landonshumway-ia:
There have been issues where states have entered the wrong SSN for a license record, but the last 4 of the SSN were still accurate so practitioners have been able to register under a license record and purchase privileges. We need to provide a way for states to correct the SSNs for their particular license records. This ticket involves adding a new optional field to the license upload API, 'previousSSN'. When this value is included with a properly formatted SSN, then the system will look up any existing records for the provider id associated with that SSN. If a matching license record is found, it will move that license record along with any associated privileges and other records over to the new provider id, and delete the old records under the previous ssn to remove the duplicate records.

### Requirements List
- This feature is gated with a feature flag so we can disable it if needed to prevent states from using it.

### Testing List
- Code review

Closes #1725 

## Summary by CodeRabbit

* **New Features**
  * Added optional `previousSSN` to SSN correction flows, enabling migration of an existing license (and related records) to the corrected provider.
  * Introduced a dedicated SSN-correction migration feature flag and added an SSN-correction re-registration email notification for full migrations.

* **Bug Fixes**
  * Improved rollback behavior when rollback start/end timestamps share the same month but differ by time of day.

* **Documentation**
  * Updated API, ingest, and CSV field documentation to describe `previousSSN`, plus refreshed smoke-test setup notes.

* **Tests**
  * Added unit, regression, and smoke-test coverage for `previousSSN` validation, routing, migration outcomes, and email notification behavior.


backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
```diff
@@ -2835,6 +2854,325 @@ def update_provider_home_state_jurisdiction(
...
+          the old Cognito user, and notifying the practitioner.
+        - Partial (other licenses remain): the old provider keeps its person-level records and its top-level
+          record is repopulated from its remaining licenses. The old provider id remains a valid (partial)
+          practitioner until its remaining licenses are also corrected.
+
+        Concurrency: the write against the old top-level provider record is conditioned on the dateOfUpdate
+        read at the start of the migration and executed in the first transaction batch. A concurrent migration
+        for the same old provider will fail that condition before writing anything, and its SQS retry re-reads
+        current state. The targeted license's delete is executed in the last batch so a crash mid-migration
+        leaves the license in place for the replay's idempotency guard to find; all other writes are idempotent.
```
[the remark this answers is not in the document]
landonshumway-ia: It would perhaps be more robust to split these writes into a minimum of two transactions, first all of the new records under the new provider id should be created and ensure those creative transactions are successful. Then we attempt to delete the old records in one or more follow up transactions. We would have update records at the front of the list of deletions, since they will have already been moved over and aren't critical for replays. Then the provider and target license record will be appended to the list of items to be deleted last, since those two records are critical for replay.

With this strategy, in theory if the write fails during the creation of new records, we just replay and they will get repopulated and the pk/sk pattern will stay the same for idempotent writes to the same record. If the write fails during the deletion process, the license and provider record will be in the same batch which will allow us to replay, get at least the target license and provider record, and then trigger the cognito deletion/email process if the provider was registered. 

Please analyze this strategy in terms of its viability for replays.

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
```diff
@@ -2835,6 +2854,325 @@ def update_provider_home_state_jurisdiction(
...
+        try:
+            old_provider_data = old_provider_records.get_provider_record()
+        except CCInternalException:
+            # The top-level provider record was already deleted by a partially-completed teardown that is now
+            # being replayed; continue the migration without the concurrency fence
+            logger.warning('Old provider record not found; continuing replay of a partially-completed migration')
+            old_provider_data = None
+
+        # The corrected license was the old provider's only license: tear the old provider down entirely
+        full_teardown = len(old_provider_records.get_license_records()) == 1
```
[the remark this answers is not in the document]
landonshumway-ia: If that were to actually occur (and the odds on that seem astronomically small) the state uploading the new license record for that exact same SSN could just re-upload the license again, and the provider record would be repopulated. It is true that the Cognito account would have been cleaned up if the practitioner had by chance registered with the old invalid SSN, but again, the odds of all of those conditions lining up within the subsecond window of a transaction write migrating an invalid SSN are so small that the added complexity smells of YAGNI. Let me know your thoughts on this if you disagree.

backend/compact-connect/lambdas/python/common/cc_common/data_model/data_client.py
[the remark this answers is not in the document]
landonshumway-ia: I have pushed up the initial implementation of this. Please review.

backend/compact-connect/lambdas/python/provider-data-v1/handlers/ingest.py
```diff
@@ -336,3 +370,109 @@ def _populate_update_record(*, existing_license: dict, updated_values: dict, rem
...
+        )
+    except config.cognito_client.exceptions.UserNotFoundException:
+        logger.info('Old Cognito user not found (already deleted); skipping re-registration email')
+        return
+
+    logger.info('Deleted old Cognito user after SSN correction; sending re-registration email')
+    config.email_service_client.send_provider_ssn_correction_reregistration_email(
+        compact=compact,
+        provider_email=old_registered_email,
+    )
```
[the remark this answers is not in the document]
landonshumway-ia: We have broken the transactions out into three phases in the case of a large number of items that need to be migrated. Please review and confirm if this address the issue.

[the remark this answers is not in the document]
landonshumway-ia: Please verify our fix for breaking out transactions to support replayability.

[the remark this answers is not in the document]
landonshumway-ia: @coderabbitai please verify that the current state of the code resolves this finding.

backend/compact-connect/lambdas/python/provider-data-v1/handlers/bulk_upload.py
```diff
@@ -162,25 +163,30 @@ def process_bulk_upload_file(
...
                 except TypeError as e:
                     # This will be raised, if `raw_license` includes compact and/or jurisdiction fields
                     logger.error('License contains unsupported fields', fields=list(raw_license.keys()), exc_info=e)
                     raise ValidationError('License contains unsupported fields') from e
+                # TODO - remove this flag once the feature is proven stable  # noqa: FIX002
+                if not ssn_correction_migration_flag_enabled:
+                    logger.info(
+                        'SSN-correction migration feature is disabled. Ignoring the previousSSN field if present'
+                    )
+                    validated_license.pop('previousSSN', None)
```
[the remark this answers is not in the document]
landonshumway-ia: This is acceptable as the main purpose of the feature gate is to prevent the migration. A failed validation attempt with an invalid SSN format will let the state know their field is in an invalid format which they can correct for when the feature us re-enabled.

backend/compact-connect/lambdas/python/provider-data-v1/handlers/ingest.py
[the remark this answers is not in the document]
landonshumway-ia: Resolving this as the other comment showed that this flow is now replayable based on the ordering of the transactions.

landonshumway-ia:
@coderabbitai review

landonshumway-ia:
@coderabbitai review

landonshumway-ia:
@jlkravitz This is now ready for your review. Given how critical this logic is in terms of handling practitioner records, we will want to consider some QA testing in CSG's test environment for this before releasing to prod. Thanks

backend/compact-connect/docs/README.md
```diff
@@ -107,7 +108,9 @@ If data is not available for a required field, that particular license record ca
...
+Yes. CompactConnect is designed to automatically detect and track changes to license records over time. The Social Security Number (SSN) is the unique identifier CompactConnect uses to create and match individual practitioner accounts. When you upload a license record, CompactConnect uses the provided SSN to determine whether that practitioner already has an account in the system, creating one if not found. If the practitioner's account already exists, CompactConnect will check the differences between the existing license record and the changes uploaded by the state, and apply the changes accordingly.
+
+Because accounts are matched on SSN, simply changing the SSN in your state's system and then uploading the corrected license will **not** update the practitioner's existing CompactConnect account. It will create a brand new, separate account under the new SSN and leave the original account (and any privileges tied to it) unchanged. If a license was previously uploaded with an incorrect SSN, use the `previousSSN` field (see the field table above) when uploading the corrected SSN so CompactConnect can migrate the practitioner's existing account instead of creating a duplicate.
```
jlkravitz, line 113: What are the odds that - at any stage of this correction process - multiple SSNs overlap? And how do we handle those cases? e.g.

- Incorrect SSN in system is a real SSN for someone else
- Two SSNs A and B were swapped - if they are corrected in a single upload, is that handled appropriately? What if they are corrected as two separate uploads? Then providers will be incorrectly and irrecoverably merged, no? 

(Not implying we should handle every edge case here, but wondering if it's worth emphasizing the risk behind (a) uploading incorrect SSNs and (b) trying to correct them, even though the feature exists.)

backend/compact-connect/lambdas/python/common/tests/function/test_data_client_ssn_correction.py
```diff
@@ -0,0 +1,762 @@
...
+        self.assertEqual(2, len(provider_updates))
+        self.assertEqual(1, len([update for update in provider_updates if update['updateType'] == 'ssnCorrection']))
+
+        # the migrated license must carry the corrected ssnLastFour
+        migrated_license = self._get_records_of_type(NEW_PROVIDER_ID, 'license')[0]
+        self.assertEqual(NEW_SSN_LAST_FOUR, migrated_license['ssnLastFour'])
+
+        # the migrated military affiliation record must reference document keys under the new provider id.
+        # (The caller moves the underlying S3 objects by listing the old provider id's keyspace directly, not
+        # from this DynamoDB record, so it is not reflected in the migration result.)
```
jlkravitz: What exactly is "not reflected in the migration result"? 

This comment otherwise implies to me that the S3 object itself is moved- true?

backend/compact-connect/lambdas/python/common/tests/function/test_data_client_ssn_correction.py
```diff
@@ -0,0 +1,762 @@
...
+                with self.assertRaises(CCInternalException):
+                    self._migrate()
+            sks_per_attempt.append(_update_record_sks_under_new_provider())
+
+        # every replay reproduced the exact same set of update-record sks, so no duplicates accumulated
+        self.assertEqual(sks_per_attempt[0], sks_per_attempt[1])
+        self.assertEqual(sks_per_attempt[0], sks_per_attempt[2])
+        # one migrated record of each update type was exercised (no ssnCorrection yet, since it is written in
+        # the final transaction that never commits here)
+        self.assertEqual(3, len(sks_per_attempt[0]))
```
jlkravitz, line 494: Why does the final transaction never commit here? Didn't we run `_migrate`?

backend/compact-connect/lambdas/python/common/tests/function/test_data_client_ssn_correction.py
```diff
@@ -0,0 +1,762 @@
...
+        elif isinstance(value, str) and target in value:
+            paths.append(path)
+        return paths
+
+    def test_migration_rekeys_every_provider_id_reference_except_ssn_correction_previous_snapshot(self):
+        """Every provider id reference in the migrated records — including the 'previous' snapshots embedded
+        in update records and military affiliation document keys — must be re-keyed to the new provider id.
+        The single intentional exception is the ssnCorrection provider update record, whose 'previous' object
+        snapshots the old provider record verbatim, old provider id included.
+        """
```
jlkravitz, line 533: I love this test- very clearly written and self-documenting!

backend/compact-connect/lambdas/python/common/tests/function/test_data_client_ssn_correction.py
```diff
@@ -0,0 +1,762 @@
...
+            for index, transaction in enumerate(executed_transactions)
+            if all('Put' in item and self._key(item)['pk']['S'] == new_provider_pk for item in transaction)
+        ]
+        delete_transaction_indexes = [
+            index
+            for index, transaction in enumerate(executed_transactions)
+            if all('Delete' in item and self._key(item)['pk']['S'] == old_provider_pk for item in transaction)
+        ]
+        self.assertTrue(create_transaction_indexes)
+        self.assertTrue(delete_transaction_indexes)
```
jlkravitz: Should this just be a length assertion?

backend/compact-connect/lambdas/python/common/tests/function/test_data_client_ssn_correction.py
```diff
@@ -0,0 +1,762 @@
...
+        ):
+            with self.assertRaises(CCInternalException):
+                self._migrate()
+
+        # the old top-level provider record and target license must still be present for the replay
+        self.assertEqual(1, len(self._get_records_of_type(DEFAULT_PROVIDER_ID, 'provider')))
+        self.assertEqual(1, len(self._get_records_of_type(DEFAULT_PROVIDER_ID, 'license')))
+
+        # replay: succeeds, tears the old provider down, and still reports the registered email for the
+        # Cognito/email path (the bug this ordering fixes was losing that email on replay)
```
jlkravitz: What do you mean by the "Cognito/email path" here?

backend/compact-connect/lambdas/python/common/tests/function/test_data_client_ssn_correction.py
```diff
@@ -0,0 +1,762 @@
...
+            item
+            for item in final_transaction
+            if 'Delete' in item
+            and self._key(item)['pk']['S'] == old_provider_pk
+            and 'license/' in self._key(item)['sk']['S']
+        )
+        self.assertIn('Delete', target_license_delete)
+
+    def test_large_migration_replays_cleanly_after_final_transaction_failure(self):
+        """The core replay guarantee for large migrations: if the atomic final transaction fails after the
```
jlkravitz, line 717: Is the migration replayable even if other changes happen in the meantime? 

e.g.

- Old provider and New provider exists
- Provider buys new privileges
- Admin adds an encumbrance

Can I safely replay now?

backend/compact-connect/lambdas/python/common/tests/unit/test_provider_record_util.py
```diff
@@ -1335,3 +1335,244 @@ def test_calculation_returns_oldest_renewal_date_if_privilege_expired_and_then_w
         )
 
         self.assertEqual(datetime.fromisoformat('2098-04-04T12:59:59+00:00'), active_since)
+
+
+class TestGetRecordsAssociatedWithLicense(TstLambdas):
+    """Coverage for ProviderUserRecords.get_records_associated_with_license.
+
+    The SSN-correction migration (migrate_provider_for_ssn_correction) relies on this method to return the
+    target license together with EVERY record that depends on it across MULTIPLE privileges and multiples
```
jlkravitz: What does "multiples" mean in this context?

backend/compact-connect/lambdas/python/disaster-recovery/handlers/rollback_license_upload.py
```diff
@@ -429,9 +429,14 @@ def _query_gsi_for_affected_providers(
...
-    # Generate list of year-month strings to query
-    current_date = start_datetime.replace(day=1)
-    end_month = end_datetime.replace(day=1)
+    # Generate list of year-month strings to query.
+    # NOTE: We must zero out the time-of-day components here, not just the day. Otherwise, if
+    # start_datetime's time-of-day is later than end_datetime's time-of-day (e.g. start=21:09:55,
+    # end=12:00:00), the initial current_date <= end_month comparison below can incorrectly evaluate
+    # to False even though both timestamps fall within the same month, causing this loop to produce
+    # zero year-months and silently skip the GSI query entirely.
+    current_date = start_datetime.replace(day=1, hour=0, minute=0, second=0, microsecond=0)
```
jlkravitz: Is `current_date` even the right name here? Given we're replacing the day, this seems like it should be called `current_month`?

backend/compact-connect/lambdas/python/disaster-recovery/tests/function/test_rollback_license_upload.py
```diff
@@ -558,6 +558,39 @@ def test_provider_license_record_reset_to_prior_values_when_upload_reverted(self
         license_updates = provider_records.get_all_license_update_records()
         self.assertEqual(len(license_updates), 0, 'License update records should be deleted')
 
+    def test_provider_found_when_start_time_of_day_is_after_end_time_of_day(self):
```
jlkravitz, line 561: Nice find!

backend/compact-connect/lambdas/python/provider-data-v1/handlers/ingest.py
```diff
@@ -44,8 +47,24 @@ def preprocess_license_ingest(message: dict):
 
             # Add the last 4 digits of SSN to the detail
             message['ssnLastFour'] = ssn[-4:]
-            # delete the ssn value from memory so it can be cleaned up as soon as we are done with it
+
+            if previous_ssn is not None and previous_ssn != ssn:
+                # The state is correcting a previously-uploaded SSN. Resolve the previous SSN to its provider id
+                # so the ingest handler (which has no SSN access) can migrate that provider's records. If the
```
jlkravitz, line 53: Isn't this the ingest handler?

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_licenses.py
```diff
@@ -94,6 +94,63 @@ def test_post_licenses_puts_expected_messages_on_the_queue(self):
...
+        # patch the module-level cached flag value directly, so this test is independent of module import order
+        with patch('handlers.licenses.ssn_correction_migration_flag_enabled', True):
+            queue_messages = self._post_license_with_previous_ssn()
+
+        self.assertEqual(1, len(queue_messages))
+        message = json.loads(queue_messages[0].body)
+        self.assertEqual('123-12-9876', message['previousSSN'])
+
+    # TODO - remove this test once the LICENSE_SSN_CORRECTION_MIGRATION_FLAG scaffolding is removed # noqa: FIX002
+    def test_post_licenses_rejects_previous_ssn_when_flag_disabled(self):
```
jlkravitz, line 130: Can't this test juse use _post_license_with_previous_ssn?

backend/compact-connect/stacks/ingest_stack.py
```diff
@@ -77,13 +98,43 @@ def _add_v1_ingest_chain(self, persistent_stack: ps.PersistentStack, data_event_
             treat_missing_data=TreatMissingData.NOT_BREACHING,
         ).add_alarm_action(SnsAction(persistent_stack.alarm_topic))
 
+        # The invocation-error alarm above only catches failures that escape the handler. The sqs_handler
+        # reports per-message failures as batch item failures (the invocation still succeeds), and some paths
+        # (e.g. the best-effort S3 document move during an SSN-correction migration) log an ERROR without
```
jlkravitz, line 103: What are the consequences of the S3 document move failing and us not fixing the issue?

backend/compact-connect/tests/smoke/config.py
```diff
@@ -84,6 +84,25 @@ def test_provider_user_username(self):
...
+        """The provider id the test provider's records live under when their SSN is CC_TEST_PROVIDER_MOCK_SSN.
+
+        Used by the full migration smoke test to detect and recover from a prior run that was interrupted
+        after migrating the test provider off of this provider id but before migrating it back.
+        """
+        return os.environ['CC_TEST_PROVIDER_ORIGINAL_PROVIDER_ID']
+
+    @property
+    def provider_user_bucket_name(self):
+        """The provider users S3 bucket, which holds practitioner-uploaded documents."""
```
jlkravitz:
```suggestion
        """The provider user's S3 bucket, which holds practitioner-uploaded documents."""
```

backend/compact-connect/lambdas/python/provider-data-v1/tests/unit/test_handlers/test_ingest.py
```diff
@@ -45,3 +45,62 @@ def test_preprocess_license_ingest_removes_ssn_from_record(self, mock_config):
...
+        self.assertEqual(new_provider_id, detail['providerId'])
+        self.assertEqual(previous_provider_id, detail['previousProviderId'])
+        self.assertEqual('1234', detail['ssnLastFour'])
+        # neither SSN may ever reach the event bus
+        self.assertNotIn('ssn', detail)
+        self.assertNotIn('previousSSN', detail)
+
+    @patch('handlers.ingest.config', autospec=False)
+    def test_preprocess_license_ingest_omits_previous_provider_id_when_it_resolves_to_same_provider(self, mock_config):
+        # both SSNs resolve to the same provider id, so there is nothing to migrate
```
jlkravitz, line 86: Is this a real possibility (two SSNs mapping to the same provider ID), or just a stress test of the event bus message ingestion pipeline?

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_ingest.py
```diff
@@ -897,3 +897,325 @@ def test_multiple_license_types_different_jurisdictions(self):
...
+@mock_aws
+@patch('cc_common.config._Config.current_standard_datetime', datetime.fromisoformat('2024-11-08T23:59:59+00:00'))
+class TestIngestSsnCorrection(TstFunction):
+    """Function tests for the SSN-correction migration orchestration in ingest_license_message.
+
+    The old (incorrect-SSN) provider uses the test-data-generator default provider id; the corrected SSN
+    resolves to NEW_PROVIDER_ID.
+    """
+
+    OLD_PROVIDER_ID = '89a6377e-c3a5-40e5-bca5-317ec854c570'
```
jlkravitz: I think there's a constant defining the default provider ID in another test file? Could we just use that to more tightly couple these?

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_ingest.py
```diff
@@ -897,3 +897,325 @@ def test_multiple_license_types_different_jurisdictions(self):
...
+        )
+
+    def test_full_migration_deletes_cognito_user(self):
+        self._put_old_provider_records()
+        self._create_old_cognito_user()
+
+        resp = self._run_ingest_with_previous_provider_id()
+        self.assertEqual({'batchItemFailures': []}, resp)
+
+        self.assertFalse(self._when_old_cognito_user_exists())
```
jlkravitz: This helper might be better named `_does_old_cognito_user_exist`, given how it's being used?

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_ingest.py
```diff
@@ -897,3 +897,325 @@ def test_multiple_license_types_different_jurisdictions(self):
...
+        old_record_types = {record['type'] for record in old_records}
+        self.assertIn('provider', old_record_types)
+        self.assertIn('license', old_record_types)
+        self.assertIn('militaryAffiliation', old_record_types)
+
+        # person-level records are not copied to the new provider
+        new_record_types = {record['type'] for record in self._get_provider_records(self.NEW_PROVIDER_ID)}
+        self.assertNotIn('militaryAffiliation', new_record_types)
+
+        # the old Cognito user remains and no re-registration email was sent
```
jlkravitz, line 1110: Do we have any stats on the following set of providers?

- The provider will require a _partial_ (not full) migration.
- There exists a cognito user associated with the old provider ID.

Assuming the odds of two providers having identical registration data – last 4 SSN, name, DOB, etc – are near zero, a full migration constitutes a simple provider ID reassignment of a known identity. That's why I'm focused on partial migrations only, which imply that a registered provider will have seen data that isn't theirs. 

e.g.
- KY license uploaded for Alice with SSN XXX, feeds provider info
- OH license uploaded for Bob with SSN XXX (incorrect, but matches Alice's), KY is still the primary license
- Alice registers under KY info (Alice, SSN XXX)
- Alice sees their license info for KY and OH.
- OH realizes they uploaded the wrong SSN XXX, which should have been YYY. They do so, forcing a partial migration.
- Alice now sees just her KY license. 

At best, that's confusing ("Why do I, Alice, have a license in OH?") and at worst, revealing of sensitive info.  I'd be interested in understanding:

1. The scale of this case for this particular scenario we're solving for, which will help us understand  if this is even a problem worth thinking about;
2. Mitigations for future cases (e.g.,  what data would Alice have erroneously seen about Bob's OH license in the above scenario, if anything? Sensitive-data aside, would it improve UX to send Alice an email, too, in the case of a partial migration?

jlkravitz (approved):
Lots of good stuff! Sorry for the delay on the review here. Going to do some testing tomorrow.

jlkravitz (changes requested):
Whoops- accidentally approved. Unapproving only to avoid any accidental merges :)

backend/compact-connect/lambdas/python/provider-data-v1/handlers/ingest.py
landonshumway-ia: No, this is the preprocessor, which runs before the ingest handler and is the only lambda in the license upload pipeline that has IAM access to the SSN table. We kept all both handlers in the same file (I can't remember why) which makes it somewhat confusing, but they are different lambdas.

backend/compact-connect/stacks/ingest_stack.py
landonshumway-ia: If the automated process does not fix the issue, a developer would need to manually copy the documents over from the old keyspace to the new (likely through a CLI command or through the console). If this is not performed the user's military documents would not be visible in the app.

backend/compact-connect/lambdas/python/provider-data-v1/tests/unit/test_handlers/test_ingest.py
landonshumway-ia: It is a stress test (likely overkill to be honest). Every SSNs generates a new UUID which while the probability is not 0, in practical terms the provider UUIDs will always be unique for every SSN.

backend/compact-connect/lambdas/python/common/tests/function/test_data_client_ssn_correction.py
landonshumway-ia: The comment is outdated from when the S3 objects were moved _after_ the DynamoDB records, which is no longer the case. I updated the algorithm to move the S3 objects before the DynamoDB records for improved replayability. I'll update the comment.

backend/compact-connect/lambdas/python/common/tests/function/test_data_client_ssn_correction.py
landonshumway-ia: Yes, but during that run we purposefully fail any attempts to clean up the old records so we can replay the same migration multiple times and ensure we produce the same pk/sk patterns for update records. So in this particular test the final transaction is never committed.

backend/compact-connect/lambdas/python/common/tests/function/test_data_client_ssn_correction.py
landonshumway-ia: I'll update the comment to be more explicit

backend/compact-connect/lambdas/python/common/tests/function/test_data_client_ssn_correction.py
landonshumway-ia: It depends on which phase it fails. This does not protect against every theoretical case, but it is fairly robust because we are running the create transactions before the delete transactions. So in theory, if only the create transactions are successful, and then the provider buys new privileges under their old provider id, or an encumbrance is added between replays, on the replay it would detect the new records associated with the old license record and run a new set of create transactions, which would now include the privilege/encumbrance records. If the privileges were purchased in the sub-seconds when the old records were being deleted... honestly the margin is so slim I don't think it's practically possible and the purchase handler would error out trying to write records to the old pk because we have a condition expression where the purchase transaction only succeeds if the pk still exists in the table.

So in almost every practical case, the replay logic here is safe, and for the vast majority of migrations there will be less than 100 operations to perform so they will all be managed within a single atomic transaction.

backend/compact-connect/lambdas/python/common/tests/unit/test_provider_record_util.py
landonshumway-ia: It is just referring to the fact that the logic ensures that every matching record of a specific type even if there are more than one. I will reword this.

backend/compact-connect/docs/README.md
landonshumway-ia: I've added an additional section here capturing the risks/impacts of incorrect SSNs.

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_licenses.py
landonshumway-ia: I've updated the helper so it could be used in this test as well.

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_ingest.py
landonshumway-ia: We don't have stats on how many will be a partial migration (though there is at least one practitioner that we are aware of that has licenses in two states impacted by this, but the other license was uploaded under the correct SSN, so that migration for the incorrect license record will still be a full migration). As you stated, the odds for a license uploaded under an incorrect SSN matching another licensed practitioner that happens to have a license in another state is low (though the probability increases over time as more states come onboard and start uploading records into production). 

For the particular issue this feature was added to address, on the day of the incident we counted 615 licensees in Counseling that are already registered which will need to re-register, with over 455 privileges purchased under account under an account with an incorrect SSN, so all of those individuals will need to re-register. 

These are great questions that I think are worth discussing with @isabeleliassen, though I don't think they would hold up this current feature work, since the damage in this case has already been done.

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_ingest.py
jlkravitz: @landonshumway-ia _very_ much a nit, but the update leaves the 's' in exists- can we update it to read `_does_old_cognito_user_exist`?

landonshumway-ia: @jlkravitz done

backend/compact-connect/lambdas/python/common/tests/function/test_data_client_ssn_correction.py
```diff
@@ -0,0 +1,760 @@
...
+        ):
+            with self.assertRaises(CCInternalException):
+                self._migrate()
+
+        # the old top-level provider record and target license must still be present for the replay
+        self.assertEqual(1, len(self._get_records_of_type(DEFAULT_PROVIDER_ID, 'provider')))
+        self.assertEqual(1, len(self._get_records_of_type(DEFAULT_PROVIDER_ID, 'license')))
+
+        # replay: succeeds, tears the old provider down, and still reports the registered email for the
+        # cqller to delete the Cognito account and send the email notification
```
jlkravitz: nit :)
```suggestion
        # caller to delete the Cognito account and send the email notification
```

backend/compact-connect/lambdas/python/common/tests/function/test_data_client_ssn_correction.py
jlkravitz: Got it-helpful. Thank you!

backend/compact-connect/lambdas/python/provider-data-v1/handlers/ingest.py
jlkravitz: Ahhh, OK. I see.

backend/compact-connect/lambdas/python/provider-data-v1/tests/function/test_handlers/test_ingest.py
jlkravitz: Got it. I agree the damage is done in this case, though I'm thinking about it from two angles.

1. We should probably have a full understanding of the damage that was done around data leakage (rather than just UX around re-registering). I think it would make sense to get that number if we can – whether by adding a log in this PR, or doing a one-off query before running the migration.
2. Since this feature won't necessarily _require_ CSG's intervention for a state to use in the future (is that really what we want?), it seems wise to have at least some form of audit trail around its use and impact.

I think (1) is worth considering before we merge (or at least before we run the first prod migration), while (2) could come soon after.

landonshumway-ia: @jlkravitz We could add CloudWatch custom metrics for SSN corrections that result in full migrations and partial migrations so we can measure in CloudWatch, and potentially set an alarm for partial migrations (though I'm considering what the action item would be if we do set an alarm. Perhaps it would just make CSG aware that a practitioner may have viewed the license data for another practitioner)? This would at least give us some visibility into the impact of states uploading incorrect SSNs. Thoughts?

jlkravitz (approved):
@isabeleliassen Landon will be adding CloudWatch alarms and metrics, along with some updates to the documentation on `previousSSN`, in a follow-up PR. Approving this so that we can merge into the `test` environment and test thoroughly. We will not be merging this feature into prod until (likely) early next week.

All to say- this is good to merge, and we'll follow-up with some small enhancements!

PR #1761, opened 2026-07-15
[Social Work] - Add detail of random 4 digit segment for CUID

landonshumway-ia:
Based on feedback from the commission, we are refining the design for the CUID format to include a random 4 digit segment to be included. This helps reduce the risk that a user transposes a number while searching and matches a different practitioner than who they intended to search for. This updates the technical design documentation explaining the new format.

Closes #1702 

## Summary by CodeRabbit

* **Documentation**
  * Updated the planned compact CUID format to `SWC-<random>-<counter>` by adding a four-digit random segment to reduce the chance of mis-keyed lookups resolving to a different valid practitioner.
  * Clarified that identifier uniqueness is still guaranteed exclusively by an incrementing DynamoDB counter (via an atomic claim mechanism).
  * Documented that claimed numbers are never reused, even when identifiers are skipped, and that unused claims are discarded on failure.

### DIFF
--- a/backend/social-work-app/docs/design/README.md
+++ b/backend/social-work-app/docs/design/README.md
@@ -370,7 +370,9 @@ guarantees to be atomic even under concurrent invocations, ensuring no two pract
 counter value. Because the counter is only ever incremented and never read-then-written non-atomically, uniqueness
 is guaranteed regardless of concurrent ingest activity across jurisdictions.
 
-**Random segment:** The four-digit `<random>` segment is generated at CUID-assignment time and prepended to the
+##### Random segment
+
+The four-digit `<random>` segment is generated at CUID-assignment time and prepended to the
 claimed counter value. Its purpose is to make it harder for a member of the public to accidentally land on a
 *different* valid, issued CUID when a value is mis-keyed — for example when digits of the counter are transposed or
 mistyped. Because the counter alone is a short, dense, monotonically-increasing sequence, most nearby integers
--- a/backend/social-work-app/docs/design/README.md
+++ b/backend/social-work-app/docs/design/README.md
@@ -357,24 +357,41 @@ short and easy for a member of the public to read, share, and search by. Per Com
 - **Person-based, not license-based**, similar to the RID, persistent for mobility: the CUID will remain the same even if the practitioner's home state later changes under Compact rules (see [Home State Changes](#home-state-changes)).
 - **Permanently linked to the practitioner's RID** (`providerId`), so historical and future records remain consistent regardless of any home-state changes.
 
-**Format:** The CUID will use the format `SWC-<counter>`, where `<counter>` is a monotonically-increasing integer
-(e.g. `SWC-1`, `SWC-2`, `SWC-3`, ...). The counter will be tracked in a dedicated DynamoDB item keyed by
-`pk`/`sk` of `{compact}#CUID_COUNT`, and claimed atomically via a new `claim_cuid_number` method on the data
-client, following the same atomic-increment pattern used by `claim_privilege_number` for JCC privilege ID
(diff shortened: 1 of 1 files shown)


landonshumway-ia:

backend/social-work-app/docs/design/README.md
```diff
@@ -357,24 +357,44 @@ short and easy for a member of the public to read, share, and search by. Per Com
...
+digits means a transposition or single-character error must *also* match the random segment to collide with a real
+record, which is far less likely. This protection depends on CUID lookups always matching the **full** CUID string.
+partial or counter-only matches will not be supported.
+
+The random segment is generated using Python's standard-library
+[`secrets`](https://docs.python.org/3/library/secrets.html) module — specifically `secrets.randbelow(10000)`,
+zero-padded to four digits (`f"{secrets.randbelow(10000):04d}"`). No third-party library is required.
+The random segment is **not** relied on for uniqueness (it is not checked for collisions and two
+practitioners may legitimately share the same four digits). Uniqueness is provided solely by the atomically-claimed
+`<counter>`. It is purely a defense against accidental mis-entry resolving to an unrelated practitioner.
```
jlkravitz, line 389: I'm not sure this paragraph adds anything meaningful beyond what's already described above? Thoughts?

backend/social-work-app/docs/design/README.md
```diff
@@ -357,24 +357,44 @@ short and easy for a member of the public to read, share, and search by. Per Com
...
 
 **Numbers may be skipped, but are never reused:** A counter value is atomically claimed *before* the rest of the
 CUID-assignment work (writing `publicCompactIdentifier` to the provider record) completes. If a failure occurs
 after a number is claimed but before it is successfully persisted to the provider record (for example, a
 transient network error or a Lambda retry), that claimed number is simply never used and no attempt is made to
 reclaim or reuse it. Consequently, CUIDs are guaranteed to be **unique** but not guaranteed to be **sequential** or
-**gap-free**. Consumers of the CUID (including the public) should not assume that `SWC-<n>` implies exactly `n`
-practitioners have been issued a CUID, only that each issued CUID is distinct.
+**gap-free**. Consumers of the CUID (including the public) should not assume that a CUID whose counter segment is
+`<n>` implies exactly `n` practitioners have been issued a CUID, only that each issued CUID is distinct.
```
jlkravitz, line 397: I have to assume this has been considered but I have to ask for completeness – are we OK with folks knowing how many practitioners are in the system (even approximately)?

jlkravitz (approved):
Two nits!

backend/social-work-app/docs/design/README.md
landonshumway-ia: Asking @KbisonCSG to confirm, but my understanding is that is not a concern.

backend/social-work-app/docs/design/README.md
landonshumway-ia: It explains the how the design will be implemented, which is important in case another developer other than myself implements it since we don't want them to pull in an unnecessary 3rd party dependency. So while it is a little verbose, given that this is documentation for a feature not yet implemented I think it is acceptable.

backend/social-work-app/docs/design/README.md
KbisonCSG: @landonshumway-ia Unless it would be a security issue for any reason, I don't see a problem with that.

landonshumway-ia: I'm walking through it and I can't think of any reason how knowing the count of practitioners in the system would be exploitable. If anything comes to mind @jlkravitz we can discuss.

